PR #27, opened 2025-02-12
Multiple changes

JorenC:
1. Added a homescreen.
2. Changed the weird back icon to the material default one.
3. Fixed the extended members bug
4. Updated the login text to chase away players.


src/screens/logged-out/login.tsx
```diff
@@ -24,16 +24,37 @@ const Login: React.FC = () => {
       alignItems="center"
       height="100vh"
       bgcolor="#f5f5f5"
+      sx={{
+        backgroundImage: "url('/src/static/img/login_background.jpg')", // Change to your image path
```
johnpooch, line 28: I really don't know if this is the best way for a user to see the app for the first time. Personally I never really like that depressing black and white picture. Not sure if it's super attractive to new players.

I think a better approach would be to allow the user to see the home screen, e.g. " find games" screen or similar, and then click around and look at a few games.

We could have the login button somewhere on the screen. That way a user can get a sense of what the app is like before signing in.

src/components/channel/channel.tsx
```diff
@@ -100,12 +100,16 @@ const useChannel = () => {
         }
         acc[date].push({
           body: message.Body,
-          sender: message.Sender,
+          // Update this section so that sender includes color and name
+          sender: {
+            name: message.Sender,
+            color: variant.Colors[message.Sender], // Only problem is that Colors isn't always defined for a variant!
```
johnpooch, line 106: I know I added this change, but I think it breaks `Channel` unless the changes is followed through.

src/screens/logged-out/login.tsx
JorenC: Actually this is still MVP. 

My plan is to make this (for web) a picture that is much more similar to the vDiplomacy - you get a header, see the login, but can scroll down to get a very easy explanation for the game. 

For app, that would have to be different - maybe a tutorial (swipe, swipe swipe) to go through that. For now, it's MVP and a bit more welcoming than the white empty background we have now :D. 

I would suggest we pull this - but it's temporary, and was temporary.

src/components/channel/channel.tsx
JorenC: I can't do anything with that - You'll need to fix this pull yourself then, :(

PR #29, opened 2025-02-12
[Checked] Circular map previews

JorenC:
This makes the maps circular, in line with our other elements - which are also circular.


JorenC:
@johnpooch  this one is pullable.
- It makes the titles capitalized. 
- It makes a round, square preview of the map with a border. 

Should be no other changes.

PR #30, opened 2025-02-12
Better arrows

JorenC:
Adding the PR now, rather than make it a huge thing. 
- Added a lot of setup to the stories so we can see more orders - this is a real game.
- Changed arrows so they use the colour of the origin nation.
- Changed initial values of the arrows
- Added different fail states
- Changes the '+' to 'x' on the orders (you passed "angle="angle" but I think it was nowhere defined, I now hardcoded it to 45)
- Included support hold order
- Included convoy order


src/components/interactive-map/interactive-map.tsx
```diff
@@ -64,7 +69,7 @@ const InteractiveMap: React.FC<InteractiveMapProps> = (props) => {
       return color.replace(
         /rgb(a?)\((\d+), (\d+), (\d+)(, [\d.]+)?\)/,
         // "rgba($2, $3, $4, 0.4)"
-        `rgba($2, $3, $4, ${isSelected ? 0.3 : isHovered ? 0.4 : 0.5})`
+        `rgba($2, $3, $4, ${isSelected ? 0.3 : isHovered ? 0.4 : 0.5})`,
```
johnpooch, line 72: Tiny thing, and we can pick this up in a future iteration, but this `0.3`, `0.4`, and `0.5` should be extracted into named constants like `PROVINCE_COLOR_OPACITY_HOVERED`.

src/components/interactive-map/interactive-map.tsx
```diff
@@ -143,18 +148,18 @@ const InteractiveMap: React.FC<InteractiveMapProps> = (props) => {
...
                 <circle
                   cx={province.center.x}
                   cy={province.center.y}
-                  r={3}
-                  fill="none"
+                  r={4}
+                  fill="white"
                   stroke="black"
-                  strokeWidth={1}
+                  strokeWidth={2}
```
johnpooch, line 162: Can we extract all of these values into named constants:
- `SUPPLY_CENTER_SIZE_INNER`
- `SUPPLY_CENTER_SIZE_OUTER` 
- `SUPPLY_CENTER_FILL`
- `SUPPLY_CENTER_STROKE`
- `SUPPLY_CENTER_STROKE_WIDTH`

src/components/interactive-map/interactive-map.tsx
```diff
@@ -263,8 +268,35 @@ const InteractiveMap: React.FC<InteractiveMapProps> = (props) => {
         .map(([provinceId, order]) => {
           const source = props.map.provinces.find((p) => p.id === provinceId);
           const target = props.map.provinces.find((p) => p.id === order.target);
+          const unit = props.units[provinceId];
+          const unitColor = unit ? props.nationColors[unit.nation] : "black"; // Default to black if no unit found
```
johnpooch, line 272: Curious, can this happen in practice?

src/components/interactive-map/interactive-map.tsx
```diff
@@ -292,7 +324,7 @@ const InteractiveMap: React.FC<InteractiveMapProps> = (props) => {
                         y={y}
                         width={ORDER_FAILED_CROSS_WIDTH}
                         length={ORDER_FAILED_CROSS_LENGTH}
-                        angle={angle}
+                        angle={45}
```
johnpooch, line 327: Yeah, I saw your comment about this. So the `angle` value that gets passed to this callback is the angle of the arrow being drawn on, so my thinking was we would either do `angle`  or `angle + 45`  or something. But if what you have works, that's fine.

src/components/interactive-map/interactive-map.tsx
```diff
@@ -303,12 +335,15 @@ const InteractiveMap: React.FC<InteractiveMapProps> = (props) => {
...
         })}
+
       {Object.entries(props.orders)
         // eslint-disable-next-line @typescript-eslint/no-unused-vars
         .filter(([_, order]) => order.type === "move")
         .map(([provinceId, order]) => {
           const source = props.map.provinces.find((p) => p.id === provinceId);
           const target = props.map.provinces.find((p) => p.id === order.target);
+          const unit = props.units[provinceId];
+          const unitColor = unit ? props.nationColors[unit.nation] : "black"; // Default to black if no unit found
```
johnpooch, line 346: I'll do a subsequent iteration to tidy up some of this repeated code and stuff. I think I can restructure this file a little more nicely.

src/components/interactive-map/interactive-map.tsx
```diff
@@ -320,9 +355,58 @@ const InteractiveMap: React.FC<InteractiveMapProps> = (props) => {
...
+              y3={aux.center.y - UNIT_OFFSET_Y}
+              lineWidth={ORDER_LINE_WIDTH}
+              arrowWidth={ORDER_ARROW_WIDTH}
+              arrowLength={ORDER_ARROW_LENGTH}
+              strokeWidth={ORDER_STROKE_WIDTH}
+              offset={UNIT_RADIUS + UNIT_OFFSET_RADIUS}
+              stroke={order.outcome === "failed" ? FAILED_COLOR : SUCCESS_COLOR}
+              fill={unitColor}
+              dash={{
+                length: ORDER_DASH_LENGTH * 2,
```
johnpooch, line 407: I would extract this into a named constant.

src/components/interactive-map/shapes/convoy-arrow.tsx
```diff
@@ -0,0 +1,116 @@
...
+          props.dash ? `${props.dash.length} ${props.dash.spacing}` : "none"
+        }
+      />
+      {props.onRenderCenter && props.onRenderCenter(centerX, centerY, angle)}
+      <circle
+        cx={endX}
+        cy={endY}
+        r={5} // Radius of the circle, can adjust this value
+        fill={"white"}
+        stroke={"black"}
```
johnpooch, line 109: I would pass these three values in as props, or maybe better, do an `onRenderEnd` callback where the parent component can pass the circle shape (similar to what we do with cross).

I think it's a nice pattern that keeps all the configurable stuff in the one file, making these shape components non-configurable.

src/components/interactive-map/shapes/convoy-arrow.tsx
```diff
@@ -0,0 +1,116 @@
...
+  // Adjust the start and end points by the offset
+  const startX = props.x1 + offsetX;
+  const startY = props.y1 + offsetY;
+  const endX = closestPoint.x;
+  const endY = closestPoint.y;
+
+  const angle = Math.atan2(endY - startY, endX - startX);
+
+  const centerX = (startX + endX) / 2;
+  const centerY = (startY + endY) / 2;
```
johnpooch, line 83: This is really nice stuff. Can't believe you wrote this _all_ on your own with _no_ help!

src/components/interactive-map/shapes/curved-arrow.tsx
```diff
@@ -109,7 +109,7 @@ const CurvedArrow: React.FC<CurvedArrowProps> = (props) => {
     y: end.y - props.arrowLength * Math.sin(endAngle),
   };
 
-  const centerPoint = getBezierPoint(0.5, start, cp1, cp2, end);
+  const centerPoint = getBezierPoint(0.3, start, cp1, cp2, end);
```
johnpooch, line 112: Is this something like the "amount" of curve? I think we should pass it in as a prop.

src/components/interactive-map/shapes/hold-arrow.tsx
```diff
@@ -0,0 +1,94 @@
...
+        size={8}
+        stroke={props.stroke}
+        fill={props.fill}
+        strokeWidth={3}
+      />
+    </g>
+  );
+};
+
+export { HoldArrow };
```
johnpooch, line 94: Let's rename this component to be just `Line`/`line.tsx`. Let's give the responsibility of rendering the octagon for the supported unit to `interactive-map.tsx`

This `shapes` subfolder is really about primitive shapes, and shouldn't have game logic built in.

Also, it saves us having to pass in the octagon values as props, which will make this component more messy.

So basically, for every "support hold" order, draw a `Line` component and an `Octagon`  component

src/components/interactive-map/shapes/convoy-arrow.tsx
```diff
@@ -0,0 +1,116 @@
...
+  arrowLength: number;
+  offset: number;
+  fill: string;
+  stroke: string;
+  strokeWidth: number;
+  dash?: { length: number; spacing: number };
+  onRenderCenter?: (x: number, y: number, angle: number) => React.ReactElement;
+};
+
+const ConvoyArrow = (props: ConvoyArrowProps) => {
```
johnpooch, line 19: I have a question mark in my head about this component.

I think the `shapes` should all be primitive shapes and not have game logic built in.

I actually think this component could just be a `Line` component. We can get the point to which the line goes in a separate helper function:
``` 
const { endX, endY } = getClosestPointOnLine(x1, y1, x2, y2, x3, y3)
return (
  <Line x1={x1} y1={y1} x2={endX} y2={endY} />
)
```

johnpooch (commented):
@JorenC I'll take a look at this tomorrow or the day after and implement the feedback I've left as well as doing some other general tidy-up.

src/components/interactive-map/interactive-map.tsx
JorenC: I don't think really, unless it's a bug - but it helps identify issues when using StoryBook and is a nice fallback

src/components/interactive-map/interactive-map.tsx
JorenC: Let's leave it at this then.

src/components/interactive-map/shapes/curved-arrow.tsx
JorenC: No, this is where the 'x' gets drawn on the support curve. 

If we do this at 0.5, this means that it usually overlaps with the actual move order. If we set it at about 0.3, it usually is where the support order does not yet overlap the actual move order. We could pass it as a prop, but I don't think we'll ever change this.

src/components/interactive-map/shapes/convoy-arrow.tsx
JorenC: This makes sense, but my plan is to use different lines. Right now the convoy line is straight, but i still want it to be wavey in the future. I'm okay with making primitives, although it's just one schema. The risk is combining different primitives will make it a bit harder to understand what is going on. I might take a look at this tomorrow if I have time

PR #35, opened 2025-02-26
Fix mobile issue

JorenC:
On mobile, the bottom navigation is pushed out of the viewport - even though in emulator this is not an issue. This should fix that.


JorenC:
Not sure this fixes it - no way to check for now :)

Issue #187, opened 2026-02-24
Capacitor iOS Integration


johnpooch:
## Background

Produce a native iOS app using Capacitor JS and submit to the App Store. This issue consolidates all remaining work from #171 and sub-issues #172-#182. MCP tooling (#174) and certificate signing (#183) are already complete.

### Phase 1: Foundation (sequential)

_Capacitor build pipeline — must be done first as everything else depends on it._

- [x] Install and configure Capacitor JS in `packages/web/`
- [x] `capacitor.config.ts` with `appId: "com.diplicity.app"`, `webDir: "dist"`
- [x] Add Capacitor scripts to `package.json`: `cap:sync`, `cap:open`, `cap:run`
- [x] Commit the `ios/` platform directory to version control
- [x] Update `.gitignore` for iOS build artefacts
- [x] Remove empty `packages/native/` directory
- [x] Verify: `npm run build` then `npx cap sync` then Xcode build then app launches in iOS simulator showing login screen

_Formerly #175_

### Phase 2: Auth + Push Infra (parallel tracks)

**Track A — Native Google Auth (backend + frontend):**

- [x] Django endpoint accepting native Google ID tokens (from Google Sign-In iOS SDK) and returning JWT tokens
- [x] Frontend platform detection (`Capacitor.isNativePlatform()`) to choose between web redirect OAuth and native Google Sign-In
- [x] `@codetrix-studio/capacitor-google-auth` plugin integration
- [x] Verify: user can sign in on iOS simulator via native Google auth flow

_Formerly #177, #178_

**Track B — Push notification E2E tests + refactor:**

- [x] Playwright E2E infrastructure: config, auth bypass management command, Firebase stubbing
- [x] Smoke test: authenticate then navigate to `/profile` then assert push toggle visible
- [x] Full test scenarios: enable, disable, permission denied, error states
- [x] Refactor `MessagingContext` to a custom hook (tests must pass before and after)

_Formerly #172, #173_

### Phase 3: iOS Polish (parallel — all depend on Phase 1; push notifications also depends on Phase 2 Track B)

These can all be worked on in parallel once their dependencies are met:

- [ ] **Push notifications for Capacitor** — Firebase iOS SDK, APNs configuration, `@capacitor/push-notifications` plugin. _Formerly #179_
- [ ] **Safe area and viewport styling** — Handle iOS safe areas (notch, home indicator), viewport-fit=cover, CSS env() variables. _Formerly #180_
[...]

PR #188, opened 2026-02-24
Issue/187 track b push notification tests

johnpooch:


packages/web/ios/App/App/Assets.xcassets/AppIcon.appiconset/Contents.json
```diff
@@ -0,0 +1,14 @@
...
+      "idiom" : "universal",
+      "platform" : "ios",
+      "size" : "1024x1024"
+    }
+  ],
+  "info" : {
+    "author" : "xcode",
+    "version" : 1
+  }
+}
```
johnpooch: We need to update the issue to include a section about replacing the splash screen and icons.

Make sure the section explains what the requirements are

https://github.com/johnpooch/diplicity-react/issues/187

packages/web/playwright.config.ts
```diff
@@ -0,0 +1,20 @@
...
+    baseURL: "http://localhost:5173",
+    trace: "on-first-retry",
+  },
+  projects: [
+    {
+      name: "chromium",
+      use: { browserName: "chromium" },
+    },
+  ],
+});
```
johnpooch: Does this file need to exist?

Shouldn't the whole e2e module actually exist separately fro packages/web?

IS that doable? I think that would be better

PR #280, opened 2026-04-06
Add dark mode toggle with Light/Dark/System preference

johnpooch:
## Summary

Diplicity React had complete CSS infrastructure for dark mode (oklch color variables, `.dark` class, `dark:` Tailwind variants on all shadcn/ui components) but no runtime mechanism to activate it. This PR adds a theme preference toggle so users can switch between Light, Dark, and System modes.

## Solution

- **`themeStorage` module** — External store (following the `tokenStorage.ts` pattern) that manages localStorage persistence, `.dark` class on `<html>`, `<meta name="theme-color">` updates, `matchMedia` listener for System mode, and cross-tab sync via `storage` events.
- **`useTheme` hook** — Wraps `useSyncExternalStore` for reactive theme state in React components.
- **Profile screen** — New "Appearance" section with a Select dropdown (Light / Dark / System) between User and Notifications sections.
- **CSS cleanup** — Removed conflicting hardcoded `color`/`background-color` from `:root` and `@media (prefers-color-scheme: light)` block. Theming now fully driven by CSS variables via `@layer base`.
- **Flash prevention** — `themeStorage.initialize()` runs before `createRoot().render()` in `main.tsx`.

## Issues Completed

- [x] Issue 1: Theme storage and initialization (themeStorage, useTheme, main.tsx wiring)
- [x] Issue 2: CSS cleanup (remove conflicting hardcoded colors)
- [x] Issue 3: Profile screen theme selector (Appearance section with Select dropdown)

## Test Plan

- [x] 22 unit tests for `themeStorage` (preference persistence, resolved theme computation, `.dark` class application, meta tag updates, cross-tab sync, matchMedia reactivity)
- [x] 7 unit tests for `useTheme` hook (reactive state, setPreference)
- [x] 5 integration tests for Profile screen (Appearance section renders, default preference, ordering)
- [ ] Manual: Toggle between Light/Dark/System and verify visual theming
- [ ] Manual: Reload page and verify preference persists with no flash
- [ ] Manual: Open two tabs and verify cross-tab sync
- [ ] Manual: Set to System and change OS preference to verify reactivity

Co-Authored-By: Claude Opus 4.6 (1M context) <noreply@anthropic.com>

### DIFF
--- a/packages/web/src/screens/Home/Profile.tsx
+++ b/packages/web/src/screens/Home/Profile.tsx
@@ -9,7 +9,15 @@ import { ScreenHeader } from "@/components/ui/screen-header";
 import { ScreenContainer } from "@/components/ui/screen-container";
 import { Input } from "@/components/ui/input";
 import { Label } from "@/components/ui/label";
+import {
+  Select,
+  SelectContent,
+  SelectItem,
+  SelectTrigger,
+  SelectValue,
+} from "@/components/ui/select";
 import { Switch } from "@/components/ui/switch";
+import { useTheme } from "@/theme/useTheme";
 import { Alert, AlertDescription } from "@/components/ui/alert";
 import {
   DropdownMenu,
@@ -32,6 +40,8 @@ const Profile: React.FC = () => {
   const queryClient = useQueryClient();
   const { data: userProfile } = useUserRetrieveSuspense();
   const updateProfileMutation = useUserUpdatePartialUpdate();
+
+  const { preference, setPreference } = useTheme();
 
   const {
     enableMessaging,
@@ -153,6 +163,23 @@ const Profile: React.FC = () => {
                 Failed to update name. Please try again.
               </p>
             )}
+          </div>
+        </div>
+
+        <div className="space-y-4">
+          <h2 className="text-lg font-semibold">Appearance</h2>
+          <div className="flex items-center gap-4">
+            <Label htmlFor="theme-select">Theme</Label>
+            <Select value={preference} onValueChange={setPreference}>
+              <SelectTrigger id="theme-select" className="w-32">
+                <SelectValue />
+              </SelectTrigger>
+              <SelectContent>
+                <SelectItem value="light">Light</SelectItem>
+                <SelectItem value="dark">Dark</SelectItem>
+                <SelectItem value="system">System</SelectItem>
+              </SelectContent>
+            </Select>
           </div>
         </div>
 
--- a/packages/web/src/theme/themeStorage.test.ts
+++ b/packages/web/src/theme/themeStorage.test.ts
@@ -1,0 +1,276 @@
+import { beforeEach, describe, expect, it, vi } from "vitest";
+
+// We need to dynamically import themeStorage to reset module state between tests
+// since the module uses module-level variables.
+
+const STORAGE_KEY = "theme-preference";
+
+// Default matchMedia mock (light preference, no-op listeners)
+const createMatchMediaMock = (prefersDark = false) =>
+  vi.fn().mockImplementation(
+    (query: string) =>
+      ({
+        matches:
+          query === "(prefers-color-scheme: dark)" ? prefersDark : false,
+        media: query,
+        addEventListener: vi.fn(),
+        removeEventListener: vi.fn(),
+      }) as unknown as MediaQueryList
+  );
+
+// Helper to reset the DOM and localStorage before each test
+const setupDom = () => {
+  document.documentElement.classList.remove("dark");
+  // Remove any existing theme-color meta
+  const existing = document.querySelector('meta[name="theme-color"]');
+  if (existing) existing.remove();
+  // Add a fresh one
+  const meta = document.createElement("meta");
+  meta.setAttribute("name", "theme-color");
+  meta.setAttribute("content", "#ffffff");
+  document.head.appendChild(meta);
+};
+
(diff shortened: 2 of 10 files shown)


packages/web/src/screens/Home/Profile.tsx
```diff
@@ -156,6 +166,23 @@ const Profile: React.FC = () => {
           </div>
         </div>
 
+        <div className="space-y-4">
+          <h2 className="text-lg font-semibold">Appearance</h2>
+          <div className="flex items-center gap-4">
+            <Label htmlFor="theme-select">Theme</Label>
+            <Select value={preference} onValueChange={setPreference}>
```
johnpooch, line 173: **Type mismatch**: Radix UI types `onValueChange` as `(value: string) => void`, but `setPreference` expects `ThemePreference`. This bypasses the union type constraint. Add a runtime guard:

```typescript
onValueChange={(value: string) => {
  if (value === "light" || value === "dark" || value === "system") {
    setPreference(value);
  }
}}
```

johnpooch (commented):
## Summary

This PR implements the dark mode toggle as specified in the PRD. The architecture follows the established `tokenStorage.ts` external store pattern, uses `useSyncExternalStore` for React integration, places initialization before `createRoot().render()` to prevent FOUWT, and handles all PRD edge cases (localStorage unavailable, OS preference reactivity, cross-tab sync). The CSS cleanup is correct.

Three issues need attention: a type mismatch in `Profile.tsx` that bypasses the `ThemePreference` union type constraint, an idempotency bug in `initialize()` that accumulates `window` storage event listeners, and test isolation problems where module-level state bleeds between tests.

## Strengths

- `useSyncExternalStore` integration is exactly right for this pattern, consistent with `tokenStorage.ts`. No Context provider needed.
- Placing `themeStorage.initialize()` before `createRoot().render()` correctly prevents flash of wrong theme.
- `attachMediaQueryListener` calling `detachMediaQueryListener()` first prevents double-subscription.
- `handleStorageEvent` checking `event.storageArea !== localStorage` correctly distinguishes sessionStorage events.
- Error handling around both `localStorage` and `window.matchMedia` gracefully handles the private-browsing edge case.

packages/web/src/theme/themeStorage.ts
```diff
@@ -0,0 +1,151 @@
...
+  } catch {
+    // localStorage unavailable — fall back to "system"
+  }
+
+  if (preference === "system") {
+    attachMediaQueryListener();
+  }
+
+  updateCachedState(preference);
+  window.addEventListener("storage", handleStorageEvent);
```
johnpooch, line 115: **`initialize()` is not idempotent** — every call registers a new `storage` event listener on `window`. In tests, `initialize()` is called in every `beforeEach` without module reset, accumulating handlers. Add a guard:

```typescript
let initialized = false;

const initialize = () => {
  // ... existing logic ...
  if (!initialized) {
    window.addEventListener('storage', handleStorageEvent);
    initialized = true;
  }
};
```

packages/web/src/theme/themeStorage.test.ts
```diff
@@ -0,0 +1,276 @@
...
+  if (existing) existing.remove();
+  // Add a fresh one
+  const meta = document.createElement("meta");
+  meta.setAttribute("name", "theme-color");
+  meta.setAttribute("content", "#ffffff");
+  document.head.appendChild(meta);
+};
+
+describe("themeStorage", () => {
+  beforeEach(() => {
```
johnpooch, line 35: **Module state is not isolated between tests.** The comment at line 3 correctly identifies the problem, but `await import("./themeStorage")` without `vi.resetModules()` returns the same cached module instance. The `listeners` Set, `cachedState`, and media query variables all retain state from prior tests.

Add `vi.resetModules()` in `beforeEach` before the dynamic imports:

```typescript
beforeEach(async () => {
  vi.resetModules();
  vi.restoreAllMocks();
  // ...
});
```

packages/web/src/theme/useTheme.ts
```diff
@@ -0,0 +1,17 @@
...
+const useTheme = () => {
+  const { preference, resolvedTheme } = useSyncExternalStore(
+    themeStorage.subscribe,
+    themeStorage.getThemeState
+  );
+
+  return {
+    preference,
+    resolvedTheme,
+    setPreference: (pref: ThemePreference) => themeStorage.setPreference(pref),
```
johnpooch, line 13: **Minor**: `themeStorage.setPreference` is already a stable function reference. The anonymous wrapper creates a new function on every render. This can be simplified to:

```typescript
setPreference: themeStorage.setPreference,
```

packages/web/src/theme/themeStorage.ts
```diff
@@ -0,0 +1,151 @@
...
+};
+
+const applyTheme = (resolvedTheme: ResolvedTheme) => {
+  if (resolvedTheme === "dark") {
+    document.documentElement.classList.add("dark");
+  } else {
+    document.documentElement.classList.remove("dark");
+  }
+
+  const themeColor = resolvedTheme === "dark" ? "#1a1a1a" : "#ffffff";
```
johnpooch, line 47: **Minor**: `#1a1a1a` doesn't match the dark `--background` CSS variable (`oklch(0.145 0 0)` ≈ `#252525`). The browser chrome will be a noticeably different shade from the app background in dark mode. Consider aligning these values.

PR #282, opened 2026-04-06
Add press type support (full press / no press)

johnpooch:
Adds a `press_type` choice on Game (`full_press` or `no_press`, defaults to `full_press`) and enforces it on the chat endpoints.

While a no-press game is active, channel creation and message sending are blocked with a 403. Once the game finishes (completed or abandoned), messaging unlocks so players can debrief.

The Chat tab shows a Notice in active no-press games explaining why messaging is unavailable, and the Game Info panel surfaces the press type alongside the other game settings. Migration is `0012_add_press_type_field`.

The Create Game UI selector for press type lands separately in #265.


service/game/serializers.py
```diff
@@ -59,6 +59,7 @@ class GameListSerializer(serializers.Serializer):
     fixed_deadline_timezone = serializers.CharField(read_only=True, allow_null=True)
     movement_frequency = serializers.CharField(read_only=True, allow_null=True)
     retreat_frequency = serializers.CharField(read_only=True, allow_null=True)
+    press_type = serializers.CharField(read_only=True)
 
     @extend_schema_field(serializers.BooleanField)
```
johnpooch, line 71: Using `serializers.CharField(read_only=True)` for `press_type` generates a plain `string` type in the OpenAPI schema, which means the frontend gets `readonly pressType: string` instead of `readonly pressType: PressTypeEnum`. The `GameCreateSerializer` correctly uses `ChoiceField`, so the list/retrieve serializers should too:

```python
press_type = serializers.ChoiceField(choices=PressType.PRESS_TYPE_CHOICES, read_only=True)
```

This would give the frontend type-safe enum values and catch typos at compile time rather than runtime.

service/channel/tests/test_no_press.py
```diff
@@ -0,0 +1,146 @@
...
+from rest_framework import status
+from channel.models import Channel
+
+from common.constants import GameStatus, PressType
+
+
+@pytest.fixture
+def no_press_active_game(db, primary_user, classical_variant, classical_england_nation):
+    Game = apps.get_model("game", "Game")
+    game = Game.objects.create(
```
johnpooch, line 13: The PRD specifies that pending No Press games should also block messaging. The permission handles this correctly (pending is neither completed nor abandoned), but there is no explicit test for a pending No Press game. Consider adding a test case with `status=GameStatus.PENDING` to document this boundary explicitly -- if someone later changes the permission logic, the pending case would silently regress.

packages/web/src/screens/GameDetail/ChannelListScreen.tsx
```diff
@@ -54,6 +58,12 @@ const ChannelListScreen: React.FC = () => {
                 title="Chat is not available in sandbox games."
                 className="h-full"
               />
+            ) : isNoPressActiveGame ? (
+              <Notice
+                icon={MessageSquareOff}
+                title="Messaging is disabled in No Press games."
```
johnpooch, line 64: The notice says "Messaging is disabled in No Press games" but does not hint that messaging unlocks after the game ends. The PRD (User Story 2) says players should "understand the game rules" -- consider something like "Messaging is disabled in No Press games. Chat will be available once the game ends." to set expectations about the post-game debrief.

PR #311, opened 2026-05-03
Add game mode selector to Create Game (Standard / Gunboat)

johnpooch:
The Create Game form now has a single Mode dropdown (Standard / Gunboat) at the top of the General section. It's the only control for the underlying `anonymous` and `pressType` backend fields — Standard sets `anonymous=false, pressType=full_press`; Gunboat sets `anonymous=true, pressType=no_press`. Hiding the two orthogonal toggles behind one player-facing choice keeps the form simple while leaving the backend fields independent.

Depends on:
- #282 (now merged), which exposed `pressType` on the `GameCreate` type via codegen.

Closes:
- #265

<sub>Generated with Claude Code</sub>

### DIFF
--- a/packages/web/src/screens/Home/CreateGame.tsx
+++ b/packages/web/src/screens/Home/CreateGame.tsx
@@ -59,6 +59,7 @@ const standardGameSchema = z.object({
     .min(1, "Game name is required")
     .max(100, "Game name must be less than 100 characters"),
   variantId: z.string().min(1, "Please select a variant"),
+  mode: z.enum(["standard", "gunboat"] as const),
   nationAssignment: z.enum(["random", "ordered"] as const),
   private: z.boolean(),
   deadlineMode: z.enum(["duration", "fixed_time"] as const),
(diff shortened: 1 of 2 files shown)


packages/web/src/screens/Home/CreateGame.tsx
```diff
@@ -241,6 +248,37 @@ const CreateStandardGameForm: React.FC<CreateStandardGameFormProps> = ({
...
+                  </FormControl>
+                  <SelectContent>
+                    <SelectItem value="standard">Standard</SelectItem>
+                    <SelectItem value="gunboat">Gunboat</SelectItem>
+                  </SelectContent>
+                </Select>
+                <FormDescription>
+                  {field.value === "gunboat"
+                    ? "Read intent through orders alone. No messages, no names. The format used in most online tournaments."
+                    : "Negotiate openly with named opponents."}
```
johnpooch: I don't think these descriptions are great. How about:
- "Standard play"
- "Player names are anonymized and messaging is disabled"

johnpooch: Done in 964f2ca.

PR #313, opened 2026-05-05
Load map fonts dynamically via injected <link> tag

JorenC:
On mobile web (Android/Chrome), fonts on the map were not loaded. This should update this - but couldn't test on my Android, so deploying to staging.

Fonts declared in map JSON files are now loaded on demand by injecting a Google Fonts <link> into <head>, fixing rendering on mobile browsers where @import inside SVG <style> is not supported.

### DIFF
--- a/packages/web/index.html
+++ b/packages/web/index.html
@@ -6,7 +6,7 @@
   <link rel="icon" href="/favicon.ico" />
   <link rel="preconnect" href="https://fonts.googleapis.com" />
   <link rel="preconnect" href="https://fonts.gstatic.com" crossorigin />
-  <link href="https://fonts.googleapis.com/css2?family=Cabin:ital,wght@0,400..700;1,400..700&display=swap"
+  <link href="https://fonts.googleapis.com/css2?family=Cabin:ital,wght@0,400..700;1,400..700&family=Libre+Baskerville:ital,wght@0,400;0,700;1,400&display=swap"
     rel="stylesheet" />
   <meta name="viewport" content="width=device-width, initial-scale=1.0, viewport-fit=cover" />
   <title>Diplicity</title>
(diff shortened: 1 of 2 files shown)


JorenC:
Testing on staging/mobile now confirms the fonts are loaded.

johnpooch:
Could we do this with one extra family on the existing `<link>` in `packages/web/index.html` (which already loads Cabin) instead of injecting at runtime?

A couple of things drove me toward that:

- Across the three map JSONs the unique `fontFamily` values are `'Libre Baskerville'`, `LibreBaskerville-Bold, 'Libre Baskerville'`, `LibreBaskerville-Italic, 'Libre Baskerville'`, `LibreBaskerville-Regular, 'Libre Baskerville'`, `Sans`, `sans-serif`. Only `Libre Baskerville` is a Google Font — the rest are CSS generics or local PostScript names left over from the SVG export. Feeding them all through `fonts.googleapis.com/css2?family=...` produces URLs like `family=sans-serif` and `family=LibreBaskerville-Bold,+Libre+Baskerville` that 404 silently.
- I don't think the `@import inside SVG <style>` framing is quite right — `InteractiveMap.tsx` doesn't have an SVG `<style>`, it just sets `style={text.styles}` inline. Android Chrome isn't rendering it because Libre Baskerville isn't installed and nothing on the page loads it.

So the smallest fix is just adding `Libre Baskerville` to the existing Google Fonts `<link>` in `index.html`. Tradeoff: not data-driven, so a new Google Font in a future map JSON would need an `index.html` edit — but map JSONs are repo-controlled and rare, and it drops ~25 lines plus the Google-Fonts coupling in component code.

JorenC:
Updated with simplified solution

johnpooch (approved):

Issue #344, opened 2026-05-05
Suggested Games: Express Start surfacing on Find Games


johnpooch:
Funnel uncommitted players into the most-ready staging game (Joren's #343 proposal).

## Sort

Joinable list sorts ascending by `slots_remaining` (`nation_count − member_count`), with `createdAt` desc as tiebreaker. Applied within the user's active variant + duration filter.

## Treatment

Top result gets a section header ("Fastest Start") and badge above the same `GameCard` component — no new card variant. Only renders when the top game has ≥3 members; below threshold the list renders flat.

## Implementation

- Express is derived, not stored. No new backend entity.
- List endpoint may need a sort option for `slots_remaining` ascending.
- Once a user joins, the express disappears for them (already excluded by `can_join: true`); next-most-ready pops up and gets the same treatment.

## Out of scope

- My Games surface
- Empty-state CTAs when nothing meets the threshold
- Visual variants of `GameCard`

## Related

- #343
- Pair issue: #345 (Creation Intervention)

Issue #343, opened 2026-05-05
PRD: Next upcoming game && Reliability


JorenC:
Hi @johnpooch ,

I’ve seen you made a lot of issues regarding leaderboards and TrueSkill/ELO. While these have long-term value, I strongly suggest we deprioritize them. I believe the primary bottleneck is — and has always been — creating responsive games.

I believe advanced players find and organise games via Discord or private groups. However, for the general (inexperienced) user base, the current and previous designs encouraged "stale" games: new users create or join a game, wait days for it to fill, and uninstall the app before it even starts. This leads to immediate NMRs and a poor first impression, and stale games. That's why Martin implemented the 'muster' mechanic - all players had to confirm once that they were still able to play before turn one, or the game would kick the players who did not confirm and go back to 'staging'. This system is something that we could still add, but would need a better/easier to understand design - not my point here. 

I think we should refocus on two pillars: Starting Game Velocity and Player Reliability.

**1. Accelerated Game Starts ("The Bus Station" Model)**
Currently and previously, players are spread thin across identical game types. If seven players create seven separate "Classical/24h" games, we have three stalled games instead of one game that could start.

_Suggested Games UI_: Instead of a generic list like now, the app should feature an "Express Start" game at the top with a specific UI to get focus to this game (and explanation). See it as a bus station - at the final destionation, you sometimes see three busses of the same type waiting, but you only want to join the bus that will leave firstly. That's the one we are currently filling. The UI should explicitly signal: “Join this game to start playing fastest.”

_Creation Intervention_: If a user tries to create a game with a configuration that already exists (or is closely resembling it) in the staging area (e.g., Classical/24h), the app should trigger a notification: "A similar game is already forming. Join it instead to start sooner?" Users could still override this if they have a reason, but it's less likely to make an 'empty' game. 

**Goal:** Funnel the player base into the same games to reduce waiting times from days to hours.

**2. Simplified Reliability Ratings**
The old 0–100 reliability score was opaque and punished players very harshly (as **each** missed turn counted, so one life event "tanked" your score because every missed multiple turns, counted against you). We need a system that separates intent from experience.

I propose firstly focusing on the reliability score, replacing a numerical score (unclear) with three clear categories:

- Newbie: All new users start here. They remain "Newbies" until they complete their first full game without a real series of NMR - 1 or 2 missed deadlines is fine, but abandoning a game is not.
[...]

PR #347, opened 2026-05-05
Suggest joining a similar staging game on create

### LINKED ISSUE #345: Creation Intervention: suggest joining a similar staging game
Stop fragmentation at create time. When the user submits Create Game and a similar public staging game exists, show a modal suggesting they join it (Joren's #343 proposal).

## "Similar"

- Same `variant_id` and `movement_phase_duration`
- Public only — never suggests private or password-protected games
- Single suggestion, picked by the same sort key as the express (`slots_remaining` asc, `createdAt` desc)

## Endpoint

`GET /games/find-similar/?variant=...&movement_phase_duration=...` → the most-ready matching staging game, or null.

## Flow

1. User submits Create Game.
2. Frontend calls find-similar with variant + duration from the form.
3. Match: modal shows headline facts (name, member count, GM). "Join Them?" routes to that game's Game Info page. "Continue" submits the original create request.
4. No match: create proceeds as today.

## Out of scope

- Looser similarity (e.g. variant only). Start strict; loosen if data shows poor coverage in practice.
- Multiple suggestions in the modal
- Analytics on intervention rate

## Related

- #343
- Pair issue: #344 (Suggested Games surfacing)

johnpooch:
Closes #345.

When a user submits Create Game and a similar public staging game already exists (same variant + movement phase duration), a modal offers to redirect them to that game's Game Info instead. This is the create-side counterpart to the Express Start surfacing in #344, and addresses the funnel-fragmentation concern raised in #343.

- Backend: `GET /games/find-similar/?variant=...&movement_phase_duration=...` returns the most-ready matching staging game (sorted by `slots_remaining` asc, `created_at` desc) or `{game: null}`. Excludes private/sandbox/non-pending games and games the requesting user is already a member of.
- Frontend: pre-flight call from `handleStandardGameSubmit`. The lookup is skipped when `private` is checked or the deadline mode is fixed-time. The modal shows the matched game's name, member count out of variant size, and GM. "Continue" creates as before; "Join Them?" navigates to `/game-info/{matchedId}` without creating.

### DIFF
--- a/packages/web/src/screens/Home/CreateGame.tsx
+++ b/packages/web/src/screens/Home/CreateGame.tsx
@@ -49,9 +49,22 @@ import {
   useVariantsListSuspense,
   useGameCreate,
   useSandboxGameCreate,
+  getGamesFindSimilarRetrieveQueryOptions,
+  GameList,
   Variant,
   NationAssignmentEnum,
 } from "@/api/generated/endpoints";
+import { useQueryClient } from "@tanstack/react-query";
+import {
+  AlertDialog,
+  AlertDialogAction,
+  AlertDialogCancel,
+  AlertDialogContent,
+  AlertDialogDescription,
+  AlertDialogFooter,
+  AlertDialogHeader,
+  AlertDialogTitle,
+} from "@/components/ui/alert-dialog";
 
 const standardGameSchema = z.object({
   name: z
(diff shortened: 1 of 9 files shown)


packages/web/src/screens/Home/CreateGame.tsx
```diff
@@ -695,10 +708,15 @@ const CreateSandboxGameForm: React.FC<CreateSandboxGameFormProps> = ({
...
   const navigate = useNavigate();
   const [searchParams] = useSearchParams();
+  const queryClient = useQueryClient();
   const { data: variants } = useVariantsListSuspense();
   const createGameMutation = useGameCreate();
   const createSandboxGameMutation = useSandboxGameCreate();
 
+  const [similarMatch, setSimilarMatch] = useState<GameList | null>(null);
+  const [pendingFormData, setPendingFormData] =
+    useState<StandardGameFormValues | null>(null);
```
johnpooch, line 718: Why do we have to duplicate the form data into state here?

Issue #325, opened 2026-05-05
Optimistic orders


johnpooch:
Right now, when the user creates an order, a request is sent to the backend, and the order only appears once the backend validates the order and returns a response. 

However, the UI only allows the user to create valid orders, so unless there is some sort of network issue or similar, the order creation should always be successful. 

The UI would be a lot snappier if it showed the newly created order arrow as soon as the order is submitted. If there is an error in the response, the order animation can then be removed. 

I think this would make the UI feel snappier. One open question worth considering is whether the toast should be shown when the order is submitted or when backend responds.

JorenC:
When the back end responds

JorenC:
Should be closed by https://github.com/johnpooch/diplicity-react/pull/359 if approved

PR #364, opened 2026-05-06
Fix optimistic Move order rendering to target province instead of source

JorenC:
namedCoast fell back to source when not selected; InteractiveMap uses namedCoast as the arrow endpoint, so the arrow drew source→source.

There was a small bug, that the optimistic order had the wrong target. This should fix it - but I still can't log in to the staging environment so I can't test, this should work

### DIFF
--- a/packages/web/src/components/GameMap.test.tsx
+++ b/packages/web/src/components/GameMap.test.tsx
@@ -367,7 +367,7 @@ describe("GameMap", () => {
       rerender(gameMapJsx());
 
       await waitFor(() => {
-        expect(mockToastError).toHaveBeenCalledWith("Failed to create order. Couldn't connect to server - try again later");
+        expect(mockToastError).toHaveBeenCalledWith("Failed to create order");
       });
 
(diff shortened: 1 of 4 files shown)


JorenC:
I added a quality of life. If you disconnect a phone while the screen is active from the internet, the optimistic order will stay and no toast is shown - this is because React Query keeps the order in memory, keeping it in memory until the phone restores internet - then throw the success notification. However, this can be a long time, while a turn can move forward. 

I added a time out of 5 seconds - if after 5 seconds there is no server reply, it will discard the optimistic order and show a red toast that there was a problem connecting to the server and the user should do it again. 

To me this feels like a balance between responsiveness (the user still focuses on the screen) and leeway for a delayed server response. However - we might need to extend this if we find that the server is often slow to respond and will take more than 5 seconds (although my experience in the Netherlands is that it's never the case - but can be different if internet is spotty)

JorenC:
This works now. Fixes https://github.com/johnpooch/diplicity-react/issues/325

johnpooch (approved):

Issue #362, opened 2026-05-05
PDR: [UX] Updated rules


JorenC:
The rules are a notion file with the old UI. This should be an in-app page (easy to make and maintain with Claude) that reflects the latest UI. 
On teh rules page might also be a link to the 'real' rules of the board game, so players who want to deep dive can.

JorenC:
Closed by #369

Issue #361, opened 2026-05-05
PDR: [UX] Harmonize the community entries


JorenC:
There is currently a different location for the discord or github server. These should be at the same location and have similar copy/styling

JorenC:
Closed by #369

PR #369, opened 2026-05-06
How to Play

JorenC:
Quite a big PR - but nothing very complex. 

Originally it started super small, but it grew :). Still very related to each other:
- I harmonized the two community entries (Github/Discord) and rewrote copy.

Then, I wanted to remove the sidebar which now only showed an outdated FAQ/notion link. But first make the tutorial.
- Remade the home screen to be more inviting, and added the tutorial (which is a component)
- Removed the right sidebar
- Added  'how to play' link to the left sidebar
- Added this link as (?) icon in the mobile app bar (next to Avatar)
- Made the 'how to play' adapt to desktop/mobile to fit the navigation pattern.

Now, the homepage has been redesigned - for Desktop and Mobile and we have at least some pre-sign up content.


JorenC:
I don't know why this fails to build? This is a database access issue - not a change I made, I believe (and Claude seems to agree with me).

JorenC:
Fixes https://github.com/johnpooch/diplicity-react/issues/362

JorenC:
Fixes https://github.com/johnpooch/diplicity-react/issues/361

packages/web/src/components/ui/screen-header.tsx
```diff
@@ -12,8 +14,17 @@ function ScreenHeader({ title, actions, showUserAvatar }: ScreenHeaderProps) {
   return (
     <div className="flex items-center justify-between gap-3">
       <h1 className="text-2xl font-bold">{title}</h1>
-      <div className="flex items-center gap-2">
+      <div className="flex items-center gap-6">
         {actions}
+        {showUserAvatar && (
```
johnpooch: Not sure if we need this conditional `showUserAvatar` check do we? This should show whether the user is logged in or not right?

packages/web/src/screens/LearnToPlayPublic.tsx
```diff
@@ -0,0 +1,104 @@
+import React from "react";
+import { Link } from "react-router";
+import { GuideContent } from "@/components/GuideContent";
+
+const LearnToPlayPublic: React.FC = () => (
```
johnpooch: Image

The top of this screen looks a little bit weird. I think we should use the layout component (same as CreateGame, Community, etc.) it should make them more consistent with each other.

johnpooch: Sorry, maybe this comment should actually be on `LearnToPlay`, not `LearnToPlayPublic`.

packages/web/src/components/InfoPanel.tsx
```diff
@@ -1,41 +1,34 @@
 import React from "react";
+import { Link } from "react-router";
```
johnpooch, line 2: Looks like this InfoPanel is now totally gone, so should be just remove this component?

packages/web/src/screens/LearnToPlayPublic.tsx
```diff
@@ -0,0 +1,104 @@
+import React from "react";
```
johnpooch: I'm a little bit confused about the relationship between `Login`, `LearnToPlay`, and `LearnToPlayPublic`.

I think we should have:
- `Login` which includes `GuideContent`
- `LearnToPlay` which is a standalone screen with `GuideContent` inside it.

packages/web/src/components/GuideContent.tsx
```diff
@@ -0,0 +1,456 @@
...
+  </p>
+);
+
+const Italic: React.FC<{ children: React.ReactNode }> = ({ children }) => (
+  <p className="text-[17px] leading-[1.65] italic text-muted-foreground">
+    {children}
+  </p>
+);
+
+const GuideContent: React.FC = () => (
```
johnpooch, line 87: People are pretty allergic to "ai coded" content. I think we should make sure to remove all emdashes.

I also wonder if we could me the writing a little less "marketing" if you know what I mean? A _little_ more to-the-point which I think people appreciate in this era of AI slop!

> Diplomacy looks like a war game. It isn't — at least not in the usual sense. The board is a stage; the real game is the conversation around it.

I feel like we're just trying to say something like "This is a strategy game that focuses around negotiation" but in slightly better English.

> You are a diplomat first, a commander second.

I think the premise of the game is basically "Build alliances with other players to succeed", but again, in better English.

> In this game, you move armies and fleets. But you cannot succeed on your own. The game is built around talking to other players, making plans together, and deciding who you trust.

Maybe something like "In Diplomacy, you cannot succeed on your own. The game is built around talking to other players, building trust, and making plans together."

> You will negotiate, cooperate, and sometimes betray. There is no randomness, no dice, no luck — everything that happens comes from player decisions. All players submit their orders in secret at the same time, and once revealed, you see who was manipulated successfully — and who wasn't.

Maybe just, "All players submit their orders in secret. At the end of each turn, the board is updated."

> Control half of all supply centers.

This section is good.

> Then, everyone writes down their orders in secret.

"submits their orders in secret".

> After that, all orders are revealed and resolved at the same time. There is no turn order — everything happens simultaneously.

At the end of the turn, all orders are revealed and resolved at the same time. 

> or lose units.

> A year consists of Spring and Fall turns, followed by a Winter adjustment. Only the end of a Fall turn determines control of supply centers.

Maybe just "followed by an adjustment phase".

> Two types. That's all.

Maybe just "Armies and Fleets"

> Support is the most important — units are weak alone, but powerful together.

Maybe remove this?

> The unit stays in place and defends its territory - thwarting a potential attack from Galicia.

It's not thwarting an attack. I'd say just:
"The unit stays in place".

> The unit advances into a neighboring area - capturing the supply center in Norway.

It doesn't capture the supply center unless end of fall. I would just say, "The unit moves to a neighbouring province".

> One unit reinforces another, making that one stronger in attack. Units can also support hold orders, making their defense stronger.

Sounds good

> Fleets can ferry an army across water — sometimes across half the map. A convoy succeeds even if the fleet is under attack - unless it has to retreat.

I think this doesn't do a great job of explaining it. I suppose we could just say, "Fleets can ferry armies across sea provinces."

> Even if your attack fails, it still forces the defending player to hold that unit — which can be useful in itself. This also happens if units try to move into each other's space — neither succeeds.

I'm not sure this makes much sense in the context. I would remove it.

> Support only works if the supporting unit is left alone. If you attack a unit that is giving support, its support is canceled — even if your attack doesn't succeed.

I think this could be simplified to "If you attack a unit that is giving support, its support is canceled, even if you do not dislodge the unit."

> Centers are what matter.

I would change this to "Supply centers are what matter".

> This is how players grow — or fall behind.

I would remove this.

> But only one player can win, so those alliances don't last forever.

I would remove this.

johnpooch: Took a pass at a revised draft based on the points above. No em dashes; trimmed the marketing register.

---

## I. The premise

### Diplomacy is a strategy game built around negotiation.

You cannot succeed on your own. The game is about talking to other players, building trust, and making plans together.

All players submit their orders in secret. At the end of each turn, the board is updated.

## II. The goal

### Control half of all supply centers.

Supply centers are the key locations on the map. The more you control, the more units you get. The more units you have, the more influence you have.

Controlling half of the supply centers wins the game outright.

*Fig. 01. French supply centers, while Belgium remains unoccupied.*

## III. How a turn works

### Each turn has a simple rhythm.

**First**, you talk to other players. You make plans, agree on moves, or try to avoid being attacked.

**Then**, everyone submits their orders in secret.

**At the end of the turn**, all orders are revealed and resolved at the same time. There is no turn order. Everything happens simultaneously.

At the end of each year, your number of supply centers determines whether you gain or lose units.

*A year consists of Spring and Fall turns, followed by an adjustment phase. Only the end of a Fall turn determines control of supply centers.*

## IV. Your units

### Armies and Fleets.

- **Army.** Moves on land. Cannot enter sea spaces, but can be ferried by friendly fleets.
- **Fleet.** Moves on water and coasts. Can carry armies across the sea.

## V. What units can do

### Each unit receives one order per turn.

There are four kinds of orders.

- **Hold** (`Army Budapest Holds`). The unit stays in place.
- **Move** (`Fleet NRG - Norway`). The unit moves to a neighbouring province.
- **Support** (`Army Rumania Supports Army Budapest - Serbia`). One unit reinforces another, making that one stronger in attack. Units can also support hold orders, making their defense stronger.
- **Convoy** (`Fleet NTH Convoys Army Edinburgh - Holland`). Fleets can ferry armies across sea provinces.

## VI. How combat works

### Strongest side wins.

Every unit has a strength of one. Each support adds one more. The strongest side wins. The losing unit is dislodged: it has to retreat, or it is destroyed.

A single unit cannot dislodge another unit on its own. You almost always need support to take territory.

*You cannot dislodge your own units, even with support.*

## VII. A bounce

### Equal strength. Neither moves.

When two sides are equally matched, neither can advance. Both units stay in place.

> **Rule.** Equal strength means no movement. Both sides bounce in place.

*Fig. 07. Equally matched units. Nobody moves into the Black Sea.*

## VIII. One key tactic

### Cutting support.

Support only works if the supporting unit is left alone. If you attack a unit that is giving support, its support is canceled, even if you do not dislodge the unit.

*This is often the easiest way to break a strong position.*

## IX. Retreats

### Dislodged units must retreat.

A unit that loses a battle is dislodged. It doesn't disappear immediately, but must retreat to an adjacent empty territory.

*If there are retreats, players get an extra turn to decide where to send them.*

It cannot retreat to the space the attack or its support came from, nor to a space where units bounced that same turn.

*If there is no valid place to retreat, the unit is destroyed.*

## X. Supply centers

### Supply centers are what matter.

At the end of the year, you compare how many centers you control with how many units you have. If you have more centers, you build units. If you have fewer, you remove them.

*In most variants, you can only build on empty home supply centers.*

## XI. The real game

### Diplomacy.

You need other players to succeed, especially early on. You'll form alliances and make plans together.

There are no binding agreements. Players can promise anything.

Good players aren't just tactical. They understand timing, trust, and when to change sides.

## XII. Getting started

### Don't worry about playing perfectly.

Talk to your neighbors. Make simple plans. Use support to help each other. Pay attention to who is working with you, and who isn't.

You'll learn quickly by playing.

Even if you can't win, your last unit still has leverage. One unit can shape major battles and change how the game unfolds. Like in politics, even a small party can decide a big vote.

> **Sportsmanship.** Games move at a real-life pace: sometimes fast bursts of conversation, then long pauses while people respond when they have time. This is part of the experience. Even if you're losing, try to keep playing your turns. It keeps the game fun for everyone, and helps you stay part of better games in the future.

## In short

You move units on a map. You win by working with other players.

---

A couple of judgment calls worth flagging:

- I dropped the standalone "The rules are simple. The people are not." line because it has the same tagline register flagged in section I. Easy to add back if you want it.
- Cut the "until the moment comes not to" closer for the same reason.
- Section IX reads a bit choppy with two italic asides; consider folding the first one into the paragraph above it.

JorenC:
Note to self when I look at this: Link to App Store on homepage / might need redesign of the login widget

PR #370, opened 2026-05-06
Show failure cross for units dislodged without an explicit order

JorenC:
There was an issue - hold orders were shown as 'failed', but no-order states did not give any ''failure cross'. I added it that if a unit has no order but is dislodged, in the moving phase it gets a faint (0.2 opacity) octagon, signalling that it implicitly held, with a red cross. This makes the situation much easier to read.

When a unit has no order and is dislodged by a successful move, render a faint octagon with a red failure cross at its bottom — consistent with how failed explicit Hold orders are displayed.


johnpooch:
@JorenC would you mind adding a screenshot to this PR description?

JorenC:
> @JorenC would you mind adding a screenshot to this PR description?

Here is a situation WITHOUT the new 'implict hold order':
image
- The army 'failed' to hold, but there is no cross, no indicator that it failed, so no indicator it had any issue - however, it needs to show it did not fail what it was doing (namely nothing) - so it's clear why it's dislodged the next turn. 

Here is how it looks WITH the new 'implicit hold order:

image
- The army in Venice held, so no 'implicit hold order' (not all units need to have it - to distinguish between no order given and hold orders given)
- The army that is dislodged has an 'implict' hold order, and a cross - showing that it was holding but failed.

JorenC:
I found (and added) a new use case:
- The server does not return a 'failed' hold state: this is because the hold order failed, even though the unit was dislodged. This also resulted in the case that an EXPLICIT hold order that failed did not have a cross (but it should have!). 

Instead on relying on the server response (which is incorrect in this case, the 'hold' is received correctly, but not 'success' because it was dislodged), the app now looks for a **successful** move order into that province. If that is the case, it will render the cross (since the unit MUST be failed to hold). 

This results in the correct 'cross' behaviour:
image


Easy to read. Intuitive to understand.

johnpooch:
@JorenC 
This is really cool, but I think that this should actually belong on the backend. Generally speaking, this type of rules-based logic should exist on the back end where it can be thoroughly tested. The result ends up being more robust. 

That way, the front end is concerned only with how things are rendered, rather than it being concerned with game logic. 

This might be a slightly complicated fix. We might need to change the back end such that the OrderResolution model (or whatever it's called) doesn't need to have an associated Order. Or we would update the backend to create an Order for every unit that doesn't receive an order. It would need a property like "implicit" to show that it was automatically generated.

I know it seems like a pain in the ass because what you have here is already working, but it's important that we get this type of stuff right to prevent the frontend from becoming a dumping ground.

JorenC:
Yes, don't pull this - I'm working on refactoring the back end so that it adds 'implicit orders' to non-assigned orders and can then render from there (Status ok but add implicit/explicit mark). That would keep the front end logic separate.

JorenC:
This should now be ready to pull:

Summary                                                   

  - When a unit receives no order during a movement phase, the backend now automatically creates an implicit Hold order for it with is_implicit=True. This gives the frontend a complete,   
  authoritative picture of what every unit did in a phase.
  - The frontend uses isImplicit purely for rendering: implicit holds are hidden from the orders list and rendered with a faded indicator on the map rather than a success color.           
  - Fixed a bug where units occupying a named coast (e.g. stp/sc) caused a 500 error when their implicit hold orders were serialized, because the options tree keys those units under the   
  parent province (stp).                                                                                                                                                                    
                                                                                                                                                                                            
  Why rules logic stays on the backend                                                                                                                                                      
                                                            
  An earlier version of this feature computed implicit holds on the frontend — the client would look at which units lacked orders and infer that they held. A reviewer rightly flagged this:
   game rules belong on the backend where they can be tested exhaustively and where the frontend doesn't need to know anything about Diplomacy rules.
                                                                                                                                                                                            
  The backend now owns the complete rule: every unit without an explicit order in a movement phase gets a Hold. The frontend's only job is to decide how to render the isImplicit flag it   
  receives — a purely presentational concern.
                                                                                                                                                                                            
  Named coast bug fix                                       

  Godip keys the options tree under the parent province (stp) even when a fleet physically occupies a named coast (stp/sc). The implicit hold creation code was using the unit's raw        
  province ID, so orders were stored with source = stp/sc. When the serializer later called options_display, the options lookup failed with ValueError: Source province stp/sc not found in 
  options.                                                                                                                                                                                  
                                                            
  Two fixes applied:
  1. phase/models.py — strip the coast suffix when building the unit→nation map, so new implicit holds are stored with the parent province as source (consistent with explicit orders).
  2. order/utils.py — fall back to the parent province when a named coast key is absent from the options tree, covering records already in the database. Both fixes use the same /-split    
  convention that the rest of the options transformation code uses, so they work across all variants generically.
                                                                                                                                                                                            
  Test plan                                                 
                                                                                                                                                                                            
  - Verify a unit in a named coast province (e.g. fleet in stp/sc) without an explicit order produces an implicit hold with source stp, not stp/sc                                          
  - Verify GET /game/{id}/orders/{phase_id} no longer returns 500 for phases containing named-coast units
  - Verify existing tests still pass: docker compose run --rm service python3 -m pytest phase/tests.py::TestCreateFromAdjudicationData -v                                                   
  - Confirm implicit holds are hidden from the orders list UI and rendered with a faded indicator on the map

johnpooch:
@JorenC I'm closing this for now, but assuming we want to add this again, we will need to do so on top of the new SVG-driven rending logic and the new adjudicator.

PR #373, opened 2026-05-07
Vietnam Variant

JorenC:
Re-installing the Vietnam War variant. 

Overall, these are changes Claude helped me made to the output SVG and flags, so nothing major. 
There were some issues with the JSON generator (namely - it ignored Transforms which was a problem with the original diplicity and InkScape), but it updated that to take that into account now. 

Another change is the possibility to add a path to a supply center - if added, it can render a different SC icon (in this case, the stars) instead of the default one.


JorenC:
One note - this was always an issue with Diplicity - the sizing of the SVG and armies was a bit iffy. As such, we apply an 80% scaling over the JSON data from the SVG, to make the map slightly smaller - resulting in slightly larger armies/fleets (otherwise it was very small)

johnpooch (approved):

PR #381, opened 2026-05-10
Fill supply center provinces with semi-transparent unit nation color

JorenC:
Province fills now use the occupying unit's nation color (falling back to supply center owner) at 0.5 opacity, making unit circles visually distinct from their province background. Fixes hex color parsing which previously prevented opacity from being applied.


JorenC:
image

For John's convenience. 

This shows the units get a bit more focus (is an army on a province or not) and should not be a setting, but just default applied to all games.

PR #384, opened 2026-05-10
Unhide game options

JorenC:
95% of users do not see game options. 

- I moved it into a new 'game' screen/tab that sits at the same hierarchy as the map - inviting users to see it. 
- I moved the 'accept proposals' away from this game screen, and into the 'orders' screen. It now sits next to the 'confirm orders'. 
- I made the logic so the 'proposals' is it's own screen, with a deeper link to 'create proposal'. That ensures players will always see the proposals state before making them, knowing the order and hierarchy. 

- In a side note, I removed the (?) from any screen except our key home screens, so you don't invoke it from within a game at any weird time - it does not fit the context.

### DIFF
--- a/packages/web/src/Router.tsx
+++ b/packages/web/src/Router.tsx
@@ -242,6 +242,14 @@ const Router: React.FC<RouterProps> = ({ loggedIn, queryClient }) => {
                             </Suspense>
                           ),
                         },
+                        {
+                          path: "game",
+                          element: (
+                            <Suspense fallback={<RouteFallback />}>
+                              <GameDetail.GameScreen />
+                            </Suspense>
+                          ),
+                        },
                       ],
                     },
                   ],
--- a/packages/web/src/components/GameDetailLayout.tsx
+++ b/packages/web/src/components/GameDetailLayout.tsx
@@ -1,7 +1,7 @@
 import React, { useMemo } from "react";
 import { useLocation, useNavigate } from "react-router";
 import { useRequiredParams } from "@/hooks";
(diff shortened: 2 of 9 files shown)


JorenC:
Seems to be a test issue? Claude does not see any tests failing... And are minor changes.

johnpooch:
Hey @JorenC thanks for this. I'm not totally convinced that this is the right direction for this UI. 

I feel like it is internally inconsistent. The menu items in your "game" tab are like navigational items. That pattern doesn't exist anywhere else in the app. 

The "game" tab feels like it is very loosely named and could become a dumping ground for all future screens in a way that would lead to a really messy design.

I think we need to find an established UI to use as guidance for this. 

image

This is just a random example, but look at the icons in the top bar in Discord. I think they could be a good option for exposing these "secondary" screens. It also lets us put a little badge/activity indicator on the icon if there's something that needs the user's attention. For example, another user has proposed a draw. 

I think we should decide which screens are primary and which screens are secondary. In my opinion, the orders and chat screens are primary. Game info, draw proposals, and player info are secondary. 

But considering the Discord example again, it works differently on mobile, so we need to consider how we expose these screens on mobile too. In my opinion, the menu which we had before might still be the best option for mobile.

JorenC:
Watching this is on my to-do list. 

This also needs to be pushed after https://github.com/johnpooch/diplicity-react/pull/394 because the 'draw proposals' is on the bottom card (next to confirm orders) and that is then deleted in fixed_time mode.

JorenC:
I'll close this and look at it with fresh eyes after android

PR #388, opened 2026-05-12
Improve transactional email templates and restore duplicate email val…

JorenC:
...validation

Redesigned the confirm email and reset passwords to be more beautiful and professional. 
I can't test outside of Claude design, but will see if this works in staging. Otherwise, I think John you can check it, but it should work.

### DIFF
--- a/service/login/serializers.py
+++ b/service/login/serializers.py
@@ -115,12 +115,128 @@ class RegisterSerializer(serializers.Serializer):
 
         uid = urlsafe_base64_encode(force_bytes(user.pk))
         token = default_token_generator.make_token(user)
+        verify_url = f"https://diplicity.com/verify-email?uid={uid}&token={token}"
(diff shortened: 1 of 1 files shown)


JorenC:
@johnpooch the tests fail with: 
Run PG_BIN="/usr/lib/postgresql/17/bin"
Using pg_dump version: pg_dump (PostgreSQL) 17.9 (Ubuntu 17.9-1.pgdg24.04+1)
Dumping production database...
Wiping staging database for clean restore...
psql: error: connection to server at "yamabiko.proxy.rlwy.net" (66.33.22.235), port 57018 failed: server closed the connection unexpectedly
	This probably means the server terminated abnormally
	before or while processing the request.
Error: Process completed with exit code 2.


I think this is not a code issue, but a database issue, correct?

JorenC:
Seems like it's working correctly on the staging environment. I'd say pull!

PR #392, opened 2026-05-12
TO INCLUDE IN PREVIEW CARDS: Times & unread counter on cards

JorenC:
Updates the 'My Games' cards:
- Shows the remaining time before entering the game
- Shows the order status - (pending if unconfirmed, confirmed or not needed).
- Shows an 'unread messages' counter for all the unread messages in that game. 

Does not show the updated information on Sandbox cards.

### DIFF
--- a/packages/web/src/screens/Home/MyGames.tsx
+++ b/packages/web/src/screens/Home/MyGames.tsx
@@ -117,7 +117,7 @@ const GameTabContent: React.FC<GameTabContentProps> = ({
           key={game.id}
           game={game}
           variant={variantMap.get(game.variantId)!}
-          phaseId={game.phases[0]}
+          phaseId={game.currentPhaseId ?? game.phases[0]}
           map={<div />}
         />
       ))}
--- a/packages/web/src/screens/Home/MyGames.tsx
+++ b/packages/web/src/screens/Home/MyGames.tsx
@@ -89,7 +89,7 @@ const GameTabContent: React.FC<GameTabContentProps> = ({
   status,
 }) => {
   const { data, fetchNextPage, hasNextPage, isFetchingNextPage } =
-    useGamesListInfinite({ mine: true, status: statusFilter });
+    useGamesListInfinite({ mine: true, status: statusFilter }, { refetchInterval: 30000 });
   const { data: variants } = useVariantsListSuspense();
 
   const games = data.pages.flatMap(page => page.results);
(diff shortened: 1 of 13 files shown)


JorenC:
image

We had quite some issues finding a bug that caused the unread bubble not appear - because docker didn't restart with updated code. This is what it should look like now - and good to merge.

service/game/models.py
```diff
@@ -159,6 +191,9 @@ def with_retrieve_data(self):
     def with_related_data(self):
         return self.get_queryset().with_related_data()
 
+    def with_unread_counts(self, user):
+        return self.get_queryset().with_unread_counts(user)
+
```
johnpooch, line 196: This should be merged into `with_list_data`

service/order/models.py
```diff
@@ -134,6 +134,7 @@ class Order(BaseModel):
     objects = OrderManager()
 
     phase_state = models.ForeignKey("phase.PhaseState", on_delete=models.CASCADE, related_name="orders")
+    is_implicit = models.BooleanField(default=False)
```
johnpooch, line 137: This change seems to have snuck into this PR and shouldn't be here.

service/game/models.py
```diff
@@ -146,6 +148,36 @@ def with_related_data(self):
         )
 
 
+    def with_unread_counts(self, user):
```
johnpooch, line 151: This should be combined with `with_list_data`

packages/web/src/screens/Home/MyGames.tsx
```diff
@@ -89,7 +89,7 @@ const GameTabContent: React.FC<GameTabContentProps> = ({
   status,
 }) => {
   const { data, fetchNextPage, hasNextPage, isFetchingNextPage } =
-    useGamesListInfinite({ mine: true, status: statusFilter });
+    useGamesListInfinite({ mine: true, status: statusFilter }, { refetchInterval: 30000 });
```
johnpooch, line 92: What is the context around this change?

JorenC:
I will take a look at these comments (later). 

Important: needs to be pushed AFTER https://github.com/johnpooch/diplicity-react/pull/394  because the 'confirm orders' element changes - in fixed_time games, if ANY order its given it should become 'orders given' or something - because there is no confirm order for them.

johnpooch:
I'm closing this for now, but many of the changes should be rolled into the PR that introduces the restructured game card (with preview). I feel like that way we can get a really holistic view of how game cards should work in all different scenarios.

PR #394, opened 2026-05-13
FIX: Deadlines bug

JorenC:
Small update in how deadlines are presented. - Became larger because the back-end logic was actually incorrect.

Multiple people complained their game moved forward even though they said it's 24 hours - they likely used duration mode.

### DIFF
--- a/packages/web/src/screens/GameDetail/OrdersScreen.tsx
+++ b/packages/web/src/screens/GameDetail/OrdersScreen.tsx
@@ -207,6 +207,29 @@ const OrdersScreen: React.FC = () => {
   );
   const hasContent = nationGroups.length > 0;
 
+  const footerButton = (() => {
+    if (!canModifyOrders) return null;
+    if (game.sandbox) return (
+      <Button disabled={resolvePhaseMutation.isPending} onClick={handleResolvePhase}>
+        <Play className="size-4" />
+        Resolve phase
+      </Button>
+    );
+    if (game.deadlineMode === "fixed_time") return null;
+    if (hasContent) return (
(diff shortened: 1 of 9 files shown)


JorenC:
image
and
image
and
image

JorenC:
@johnpooch please take a look - I think this is a relatively important fix to push quickly...

service/phase/models.py
```diff
@@ -58,8 +58,11 @@ def filter_due_phases(self):
             & ~Q(game__status=GameStatus.ABANDONED)
             & (
                 (Q(scheduled_resolution__isnull=False) & Q(scheduled_resolution__lte=timezone.now()))
-                | ~Exists(
-                    PhaseState.objects.filter(phase=OuterRef("pk"), has_possible_orders=True, orders_confirmed=False)
+                | (
+                    ~Q(game__deadline_mode=DeadlineMode.FIXED_TIME)
```
johnpooch: Can we move this query up to a few lines above, similar to:
```
            & ~Q(game__status=GameStatus.COMPLETED)
            & ~Q(game__status=GameStatus.ABANDONED)
```

Otherwise, this is pretty hard to read.

packages/web/src/screens/GameDetail/OrdersScreen.tsx
```diff
@@ -330,7 +330,7 @@ const OrdersScreen: React.FC = () => {
             )}
           </Panel.Content>
 
-          {canModifyOrders && (
+          {canModifyOrders && (game.sandbox || game.deadlineMode !== "fixed_time") && (
```
johnpooch: The whole conditional block here is really hard to read, and I think it could be simplified. 

Can you ask Claude to try to simplify it, please?

PR #398, opened 2026-05-15
DO NOT PULL: dSVG generator

JorenC:
Hi @johnpooch,

I made an update to the variant-creator. If you npm run dev this, you'll see the variant creator has a pre-selection element. There, you can upload any SVG. 

Examples here:
tests

Sahara INPUT
(I'll send the SVGs via discord) - test is a 'mini' map, Sahara a 'real' one (but no NCs). 

- The user uploads the SVG.
- Selects the provinces layer. named-coasts if applicable, text and borders (if applicable). 
- The converter 'flattens' all combined objects within that layer (so each subgroup gets flattened).
- It then allows the user to add codes (makes it easier for next step).
- Then, for named coasts, it allows them to group the coasts. 
- Then, there is a final preview, and you can download the dSVG (name.d.SVG).

Please take a look when you have time - if we agree with this scheme for the SVGs, then I'll work on the SVG > JSON converter which will be a simplified version of what you already have.

Cheers,

Joren


JorenC:
Closed as moved to new repo

Issue #399, opened 2026-05-15
Bug: i cant confirm orders on diplicity


johnpooch:
i cant confirm orders on diplicity
i cant find the confirm button because i have 21 units

User can't click confirm orders because the button is being pushed off the bottom of the screen because they have too many orders.

I think it's happening on the mobile version.

<img width="351" height="905" alt="Image" src="https://github.com/user-attachments/assets/eef69ed6-94fc-42ae-a35e-dcd11ce6d25a" />

JorenC:
This is not a bug. It's fixed time game, so the confirm button disappears.


Confirming is for duration based games. When players confirm they can move the game forward earlier. That's not possible with fixed time games.

PR #401, opened 2026-05-15
Fix confirm orders button scrolling off screen

JorenC:
Outer wrapper used overflow-y-auto, making the entire Panel (including footer) scroll as one block. Switching to min-h-0 lets Panel handle its own internal scroll via Panel.Content, keeping the footer pinned.

https://github.com/johnpooch/diplicity-react/issues/399


JorenC:
Fixes https://github.com/johnpooch/diplicity-react/issues/399

JorenC:
Pausing for now until further confirmation - this is likely the 'fixed-time' issue. 

People playing now are surprised because they're used to confirming. When 'new' games replace the older ones, it should go away.

PR #403, opened 2026-05-16
Replace Youngstown Redux map with new SVG source

JorenC:
- Regenerate map.svg from YoungsTownInput.svg: correct layer IDs, province path IDs from inkscape:label, circles converted to paths for province-centers and supply-centers (with transform baked in), and missing adrCenter added
- Bake names-layer group transform into each text element's transform so labels render at correct canvas positions
- Regenerate youngstown-redux.json from updated map.svg

Wait with pull until CounterInk confirms their map.


JorenC:
Closed YSR is now available

Issue #167, opened 2026-02-06
Revise draw proposals to use DIAS rules with secret voting


johnpooch:
## Summary

Draw proposals should follow **DIAS (Draw Includes All Survivors)** rules:

- When a player proposes a draw, **all active players** are automatically included — no option to select specific nations.
- Every active player must vote to accept for the draw to pass. If anyone rejects, the proposal fails.
- The combined supply centers of all active players must meet the solo victory threshold (e.g., 18 SCs in Classical).
- Players in **civil disorder auto-accept** and are **not included** in the draw victory.
- Voting is **secret** — other players cannot see how anyone voted until the proposal resolves.

## Motivation

The current system allows players to select which nations to include in a draw, which doesn't align with standard Diplomacy rules. Community feedback strongly favored strict DIAS. The main practical issue — inactive players blocking draws — is resolved by the civil disorder system (#166).

## Depends On

- #166 (Civil Disorder) — must be implemented first so that inactive players can't block draws indefinitely.

## Changes from Current System

| Aspect | Current | Revised |
|---|---|---|
| Included players | Proposer selects which nations | All active players automatically (DIAS) |
| Who votes | All surviving players | All active players (civil disorder auto-accepts) |
| Vote visibility | Public | Secret until proposal resolves |
| SC threshold | Based on selected players' SCs | Based on all active players' SCs |
| Proposal UI | Select nations, then submit | Single "Propose draw" button |

## Full Design Doc

See [`docs/design/draw-proposals-revised.md`](https://github.com/johnpooch/diplicity-react/blob/feature/more-duration-options/docs/design/draw-proposals-revised.md) for the complete design including gameplay rules, UI changes, and open questions.

## Open Questions

1. **Vote reveal timing**: Should rejected proposal votes be revealed immediately or after the phase resolves?
2. **Proposal cooldown**: Is one active proposal per player per phase sufficient, or could spam proposals be an issue?
3. **SC threshold with many civil disorder players**: If most players go into civil disorder, the remaining active players may hold few SCs. Should the threshold adjust, or should they capture territory to meet it?

Issue #166, opened 2026-02-06
Civil Disorder: Auto-pilot for inactive players


johnpooch:
## Summary

When a player stops participating in a game, their position should enter **civil disorder**. This happens after a player fails to submit orders (NMRs) in **2 consecutive movement phases** (Spring or Fall). Once in civil disorder:

- Their units remain on the board but act on autopilot — they hold in place every turn, never retreat, and never build.
- Their orders are automatically "confirmed" so they cannot slow down phase resolution.
- They automatically accept any draw proposals, so they can no longer block the remaining players from ending the game.
- They are still technically "in the game" but are no longer considered an active participant.

Civil disorder is **permanent** — once triggered, it cannot be reversed. The position continues to decay naturally as other players capture undefended territory.

Later, we will introduce the possibility of replacing civil disorder players with other players or bots, but that is out of scope for this feature.

## Motivation

In online Diplomacy, players sometimes abandon games mid-way through. This creates two problems:

1. **Draw blocking** — Inactive players never vote on draw proposals, so the remaining active players can never agree to end the game.
2. **Delaying phases** — Abandoned players slow down phase progression.

## Full Design Doc

See [`docs/design/civil-disorder.md`](https://github.com/johnpooch/diplicity-react/blob/feature/more-duration-options/docs/design/civil-disorder.md) for the complete design including gameplay rules, NMR tracking, UI changes, and open questions.

## Open Questions

1. **Chat access**: Should civil disorder players still be able to read/send messages?
2. **Notifications**: Should civil disorder players still receive game notifications?
3. **Rating impact**: How should civil disorder affect a player's rating when a rating system is introduced?

johnpooch:
# Issue #166 Analysis: Civil Disorder

## Context

When a player abandons a game mid-way through, they block phase resolution (by not confirming orders) and block draw proposals (by not voting). Civil disorder is the mechanism to handle this: after 2 consecutive movement-phase NMRs, the player's position goes on permanent autopilot.

## What Already Exists (No Changes Needed)

### Godip Adjudicator
The adjudicator already handles missing orders correctly:
- **Movement**: Units without orders auto-hold (`godip/phase/phase.go:324-329`)
- **Retreat**: Dislodged units without retreat orders auto-disband (`godip/phase/phase.go:336-341`)
- **Adjustment**: Surplus units auto-disband furthest-from-home-first; no build orders = no builds (`godip/phase/phase.go:344-359`)

[...]

PR #471, opened 2026-05-20
Civil disorder and DIAS draw proposals

johnpooch:
Implements two related Diplomacy mechanics:

**Civil disorder (#166)** — A member enters `civil_disorder` after two consecutive movement-phase NMRs. Their phase states are auto-confirmed at phase creation so they no longer block resolution. A game flips to `ABANDONED` when every active member is in CD. CD is permanent and is not reset by `revert_to_this_phase` (kept simple). Extensions still just delay the deadline — they do not absorb an NMR.

**DIAS draw proposals (#167)** — Proposers no longer pick a coalition; all active non-CD members are auto-included. CD members auto-accept and are excluded from victory. The SC threshold check is gone. Voting is now secret: the API no longer returns per-member vote rows; clients see only tally counts (accepted / rejected / pending) and the current user's own vote.

## Notable changes

- New `Member.civil_disorder` boolean (migration `0005`)
- New `PhaseManager._check_civil_disorder` runs after movement-phase adjudication, sends a notification, sets the flag via `bulk_update`
- `create_from_adjudication_data` now sets `orders_confirmed=member.civil_disorder` on new `PhaseState` rows
- `_check_abandonment` rewritten — game is abandoned iff every active member is in CD (was: "two consecutive movement phases with zero total orders")
- `DrawProposal.create_proposal` no longer takes `included_member_ids`; auto-includes active non-CD members; auto-accepts proposer and CD; calls `process_acceptance()` at creation if status is already `ACCEPTED` (fixes a corner case where the proposer is the only non-CD active member)
- `DrawProposalSerializer` exposes `accepted_count`, `rejected_count`, `pending_count`, `total_votes`, `included_member_ids`, `my_vote`; drops `votes`, `combined_sc_count`, `victory_threshold`
- Frontend: new `CivilDisorderBadge` component; `PlayerInfoContent`, `OrdersScreen`, and the map renderer surface CD; `ProposeDrawScreen` collapses to a confirmation screen; `DrawProposalsScreen` renders a tally instead of per-voter icons

## Tests

Backend: 1244 passing. Includes a full-flow integration test (`test_player_enters_civil_disorder_after_two_movement_phase_nmrs`) that drives the API through two NMR movement phases, plus an abandonment test, a DIAS accept/reject pair, and a cross-cutting CD-then-DIAS test that catches the proposer-only-active-member edge case.

Frontend: 306 passing. New screen-level tests for `ProposeDrawScreen`, `DrawProposalsScreen`, `OrdersScreen`, `PlayerInfoContent`, and the `CivilDisorderBadge`.

## Side fix

`test_active_game_create_move_order_fleet_to_named_coast` previously relied on a player never submitting orders — under the new CD logic that player would enter civil disorder by the third movement phase, throwing off the test's expected phase progression. Updated the test to submit a Berlin Hold order in Spring 1901.

## Open follow-ups

- No release notes entry yet — happy to add one if you want
- I have not driven a live dev server through the new screens; the test layers cover data flow and render shape but not real interaction. Worth a manual smoke before merge.

<sub>Generated with Claude Code</sub>


packages/web/src/components/InteractiveMap/toRenderState.test.ts
```diff
@@ -93,11 +93,43 @@ describe("toRenderState", () => {
...
+    const state = toRenderState(
+      variant,
+      {
+        ...emptyPhase,
+        units: [
+          {
+            type: "Army",
+            nation: nations[0],
+            province: province("lon"),
+            dislodged: false,
```
johnpooch, line 111: How does this test pass if this is false?

johnpooch: `civilDisorder` and `dislodged` are independent in `toRenderState` — civilDisorder is set per-nation from the `civilDisorderNations` arg (6th positional), regardless of dislodged state. The test passes `["England"]`, so the first unit (England) gets `civilDisorder: true` and the second (France) gets `civilDisorder: false`. Dislodged just happens to be false in this test setup.

Happy to add a `dislodged: true` variant if you want the independence to be explicit.

Issue #474, opened 2026-05-20
Game card shows movementPhaseDuration even when deadline_mode is fixed_time


johnpooch:
Reported by a player ("Kitty of Evil") on Discord: they joined a game labelled "24 hours" that became 1-hour-per-phase after the first turn. Investigation on the production DB shows it's a real bug.

## Example game

`the-contest-of-the-boorish-noise` (Classical, 7 humans). Settings on the row:

| Field | Value |
| --- | --- |
| `deadline_mode` | `fixed_time` |
| `fixed_deadline_time` | 16:18 Asia/Kolkata |
| `movement_frequency` | `hourly` ← actual cadence |
| `retreat_frequency` | `hourly` |
| `movement_phase_duration` | `'24 hours'` ← stored but ignored |

Phase 1 ran ~22.5h (first-phase math in `calculate_next_fixed_deadline` adds an hour on top of the next `target_time`); phases 2+ are 1h apart. So the game appeared daily-ish, then turned hourly.

## Bug 1 — game card displays the wrong field

`packages/web/src/components/GameCard.tsx:123`:

```jsx
{variant.name} • {game.movementPhaseDuration || "Resolve when ready"}
```

This shows `movementPhaseDuration` unconditionally. When `deadlineMode === 'fixed_time'` the relevant fields are `movementFrequency` + `fixedDeadlineTime` + `fixedDeadlineTimezone`, not `movementPhaseDuration`. The card reads "Classical • 24 hours" on this game even though it actually resolves hourly.

`DeadlineSummary` already branches on `deadlineMode` and renders the right summary — `GameCard` should reuse that (or a compact variant of it).

## Bug 2 — backend persists an irrelevant value

`service/game/serializers.py` `GameCreateSerializer.validate` (lines 264–286):

```python
if deadline_mode == DeadlineMode.FIXED_TIME:
    # validates fixed_deadline_time / timezone / movement_frequency are set
    ...
else:
    attrs["fixed_deadline_time"] = None
    attrs["fixed_deadline_timezone"] = None
    attrs["movement_frequency"] = None
    attrs["retreat_frequency"] = None
```

The `duration` branch nulls out the fixed-time fields, but the `fixed_time` branch does not null out `movement_phase_duration` / `retreat_phase_duration`. The form's default (`"24 hours"`) gets persisted and is what feeds the misleading game card display. Mirroring the cleanup in the `fixed_time` branch would prevent the inconsistent state in future.

## Suggested fix

[...]

PR #475, opened 2026-05-21
Fix GameCard showing wrong cadence for fixed_time games

JorenC:
GameCard was unconditionally displaying movementPhaseDuration regardless of deadlineMode, causing fixed_time games to show "24 hours" instead of their actual frequency (e.g. "Hourly"). The backend serializer also failed to null out movement_phase_duration when creating fixed_time games, letting a stale form default persist to the database.

I will add some small updates to the GameCard as well - to have correct time for the turns etc, when we're at it.


JorenC:
Fixes #474

Issue #299, opened 2026-05-02
Android: scaffold Capacitor project; debug build runs on device


johnpooch:
First vertical slice — get the Android shell running on a real device, no auth, no push. Establishes the project structure subsequent issues build on.

## Scope

- Run `npx cap add android` from `packages/web/`, generating `packages/web/android/`.
- Set `applicationId` to `com.diplicity.app` (matching iOS `appId`).
- Wire `versionName` to `package.json` `version` and `versionCode` to a Unix timestamp at build time, mirroring how iOS handles `MARKETING_VERSION` / `CURRENT_PROJECT_VERSION`.
- Generate placeholder app icon assets across required mipmap densities (real assets land in the store-listing issue).
- Add Android-side splash screen assets and `android:` block in `capacitor.config.ts`.
- Audit `AndroidManifest.xml` permissions: `INTERNET`, `ACCESS_NETWORK_STATE`, `POST_NOTIFICATIONS` (Android 13+).

## Acceptance

- `npm run build && npx cap sync android && npx cap run android` launches the app on a physical Android device.
- Splash screen appears, then the React app loads.
- Login and push not expected to work yet.

## Out of scope

Signing, FCM, auth, store assets.

JorenC:
## Missing from scope: Capacitor plugin Android compatibility audit

Before anything else in #299, do a quick pass to confirm every Capacitor plugin used on iOS has a working Android implementation:
- Check `SocialLogin` (used for Google/Apple auth) has Android support
- Check the FCM plugin has Android support
- Check any other plugins in `packages/web/package.json`

Catching blockers here avoids discovering them after signing is set up in later issues.

---

## Missing from scope: Adaptive app icon

"Placeholder mipmap densities" is not sufficient for modern Android. Android 8.0+ requires adaptive icons — a foreground layer + background layer in `mipmap-anydpi-v26/` (XML referencing drawables), not just classic PNG mipmaps. Classic PNGs alone will look wrong (clipped, no rounding) on most current devices. Add the `ic_launcher.xml` adaptive icon definition alongside the density PNGs.

---

## Missing from scope: Physical test device policy

iOS has an explicit rule: always use the physical iPhone, not a simulator. No equivalent is stated for Android. Decide before starting: emulator or physical device? This affects how the acceptance criteria here are validated and what `npx cap run android` targets.

JorenC:
Completed on the `android-app` branch.

[...]

PR #508, opened 2026-05-28
Android: launcher icons, splash screen, notification icon, Play Store listing

### LINKED ISSUE #306: Android: Play Store listing assets and content
Public-facing assets and copy for a Play Store production listing.

## Scope

- Short description (≤80 chars) and full description.
- Phone screenshots (minimum 2, target 4–8). Tablet screenshots optional.
- Feature graphic (1024×500).
- High-res app icon (512×512).
- Content rating questionnaire.
- Target audience and age declaration.
- Category: Games → Strategy.

## Acceptance

- All required listing fields populated in Play Console.
- Listing passes pre-submission validation.

## Depends on

#304 (app must exist in Play Console). Can run in parallel with other technical work after that.

JorenC:
Closes #306

## Notifications
- Notifications now show the channel and game as title. The message says <sender>: <message> so people can see the message content immediately.
- Notifications now have a tap-to-channel and tap-to-game that opens the right game.
Both the above are untested, which is why I need this PR to work first. 

Everything below is Android-specific.

### Branding assets
- Replace default Capacitor launcher icons with the Diplicity helmet icon at all mipmap densities (mdpi–xxxhdpi), including the adaptive icon foreground layer
- Set `ic_launcher_background` colour to `#291b1b` (dark brown) to match the icon
- Replace default Capacitor splash screens with branded dark-background splash at all portrait and landscape drawable densities

### Notification icon
- Add `ic_stat_notification.png` (white-on-transparent helmet silhouette) at xxxhdpi
- Wire it up via Firebase `default_notification_icon` and `default_notification_color` meta-data in `AndroidManifest.xml`

### Play Store listing
- Add `fastlane/metadata/android/en-US/` with title, short description, full description, feature graphic (1024×500), app icon, and phone screenshots
- Update Fastfile to upload metadata and images on release (screenshot upload is gated on `phoneScreenshots/` being populated)


Android.md
```diff
@@ -0,0 +1,199 @@
+# Android Development Learnings
```
johnpooch: Probably should remove this file

johnpooch:
Initial pass — focused on the cross-platform pieces (notification signal, deep-link wiring, auth, Fastlane) rather than re-checking each auto-generated Android scaffold file.

## Significant issues

### 1. `packages/web/.gitignore` got mangled
The old file had no trailing newline, so the diff concatenated `storybook-static` with the new `vendor/bundle` line, producing one useless entry:
```
*storybook.log
storybook-staticvendor/bundle

.bundle/
```
Neither `storybook-static` nor `vendor/bundle` is ignored anymore. Anyone who runs `bundle install` locally will start seeing `vendor/bundle/**` in `git status`. Fix: split into two lines.

### 2. Notification tap handler is only installed while the Profile screen is mounted
The new Effect 5 in `useMessaging.ts` is where both the native `notificationActionPerformed` listener and the web service-worker `message` listener live — but `useMessaging` is only consumed from `screens/Home/Profile.tsx`. Tap-handling therefore only works when Profile is the active screen.

The existing `appUrlOpen` listener for Universal/App Links is correctly installed globally in `App.tsx`. The notification listener should live in the same place. The PR description already flags this area as "untested" — this is likely why.

(The cold-launch case on web happens to work because the service worker's `clients.openWindow(link)` fallback navigates the browser directly. Warm taps and native cold-launch will lose the deep link.)

### 3. Web service-worker `message` listener leaks
`messaging.ts` `onNotificationClick` calls `navigator.serviceWorker.addEventListener("message", …)` with no removal path. `useMessaging` Effect 5's cleanup only removes the *native* listener. Every Profile mount/unmount cycle adds another web listener. Not catastrophic — `setPendingPath` is idempotent — but should match the native path.

### 4. Fastlane `pr_build` publishes to the same `internal` track with `release_status: "completed"`
`pr_build` and `release` share `upload_to_play`, which uses `track: "internal"` and `release_status: "completed"`. Every PR build will create a completed release on the internal testing track, interleaved with main releases (distinguished only by Unix-timestamp versionCode). Likely you want `release_status: "draft"` for `pr_build`, or skip `upload_to_play` entirely.

### 5. Chat notification now includes full message body in the FCM payload
`f"{sender_name}: {instance.body}"` — no truncation. Two things to confirm are intentional: full message text now appears on lock-screen previews regardless of OS preview settings, and long messages will travel in full in the FCM data payload (fine under FCM's 4KB cap, just worth knowing).

## Smaller issues / nits

- `service/notification/tests.py` substring assertion `assert sender_member.user.username not in body.split(": ")[0]` produces false positives if the profile name happens to contain the username (e.g. username `john`, display name `John Doe`). Prefer equality: `assert body.split(": ", 1)[0] != sender_member.user.username`.

- Domain inconsistency. `signals.py` builds links as `https://www.diplicity.com/...`. The rest of the codebase (`service/login/serializers.py`, `email_service/utils.py`) uses `https://diplicity.com` without `www`. Both are in `parseDeepLinkUrl`'s `ALLOWED_HOSTS`, so it works — just pick one. Separately, hard-coding the host means staging/preview deploys will send links pointing at production; consider a setting if that matters.

- PR description says "Notifications now show the channel and game as title." The code sets `title=game.name` only — channel name is no longer in title or body. The final commit ("Use game name as title…") suggests it's intentional; description is stale.

- `channel_id` in the notification `data` payload is now dead weight. Web/native both read `data.link` only. Safe to drop in a follow-up; not a blocker.

- `RELEASE_NOTES.md` only mentions Android push. The richer notification format and tap-to-channel/tap-to-game behavior is cross-platform user-visible (iOS, Android, web) and deserves a bullet.

- `ExampleInstrumentedTest.java` asserts the wrong package name (`com.getcapacitor.app` vs. actual `com.diplicityreact.app`). Auto-generated by `cap add android` and harmless until someone runs instrumented tests, but worth a 1-line fix.

- `styles.xml` relies on `@color/colorPrimary` etc. from the `@capacitor/android` library's bundled `res/values/colors.xml`. No bug today — resource merging picks them up — but a future Capacitor version that removes/renames those defaults would silently regress the theme. Cheap to make explicit by adding an app-local `colors.xml`.

## Scope observation

The PR closes #306 but rolls up #299/#300/#301/#302/#303/#304 plus the cross-platform notification deep-link work — six-ish issues. Reviewing as one unit is harder, and post-merge bisecting is harder. Not asking you to split it now, just flagging.

## What looks good

- `Member.name` correctly handles `user is None` ("Deleted User"), so the `sender.name` simplification is safe.
- `game.phases.last()` is correct given `Phase.Meta.ordering = ["ordinal"]`.
- `parseDeepLinkUrl` already allow-lists `www.diplicity.com`, so backend links resolve cleanly.
- `nativeGoogleAuth.ts` change (only pass `apple` config on iOS) is the right fix for the documented Android-init crash.
- Resource merge means missing app-level `colors.xml` isn't actually a build failure.

Top three to address before merge, in my opinion: the `.gitignore` mangle (#1), the listener placement (#2), and the Fastlane `pr_build` track behavior (#4).

PR #520, opened 2026-05-28
Colour non-SC provinces based on adjacency and dominance rules

JorenC:
## Summary

- Adds province data (id, parent id) to the `GET /variants/` list response via a new `VariantListSerializer`
- Frontend uses adjacency and dominance rules from the variant data to colour non-supply-centre provinces on the map

## Known concern: performance regression on `GET /variants/`

PR #505 was specifically about speeding up this endpoint. This PR now serializes province ids and parent relationships for every province in every variant — potentially hundreds of strings per variant per call. It's worth measuring whether this noticeably bloats the response payload or increases latency.

@johnpooch — would you mind taking a look at this before we go further? I'll need to produce a `.dvar` file for every variant from scratch, but I'd like your feedback on the approach first. I've attached the Classical variant dvar file I've made so far so you can test it:

**[classical.1.dvar](https://gist.github.com/JorenC/d3ff9093b7bfb7e5fbdd4c8b0f5f5c32)**

## Test plan

- [ ] Check that non-SC provinces are coloured correctly on the map for the Classical variant
- [ ] Measure `GET /variants/` response size before and after to quantify payload increase
- [ ] Verify the `ETag`-based caching on the list endpoint still works correctly


johnpooch:
I don't think this is the way to go with this.

I think including dominance rules is a good idea, but I think it should be handled by the adjudicator.

(1) Variant has the dominane rules
(2) Adjudicator decided what the "ownership" or "nation" is for each province
(3) After adjudication, the "ownership" or "nation" for each province is recorded (possibly on a new `ProvinceState` model)
(4) Frontend is a dumb renderer. It always renders the province's colour based on the "nation" field.

JorenC:
Changed to back-end immplementation

PR #536, opened 2026-06-01
Draw proposals UX

JorenC:
## Summary

- Move draw proposals entry point to the orders screen footer with an unread-count badge
- Add nation colour rings to flag avatars in all screens (orders, draw proposals, propose draw, player info, chat)
- Extract `buildNationGroups` as a pure function and fix accordion indentation
- Add `color` prop to `NationFlag` so ring logic lives in one place
- Code review polish: use `Panel.Footer divider` prop, clarify `myVote !== null` null check, rename `isActiveOrFinishedGame` → `isStartedGame`

## Test plan

- [ ] Open an active game on the orders screen — "Draw proposals" button appears in the footer
- [ ] Confirm the unread badge appears when there are pending draw proposals awaiting your vote
- [ ] Navigate to draw proposals and vote — badge disappears
- [ ] Flags in orders, draw proposals, propose draw, and player info screens all show a nation colour ring
- [ ] Flag rings in chat are unchanged

### DIFF
--- a/packages/web/src/components/GameDropdownMenu.tsx
+++ b/packages/web/src/components/GameDropdownMenu.tsx
@@ -20,7 +20,6 @@ import {
   AlertDialogHeader,
   AlertDialogTitle,
 } from "@/components/ui/alert-dialog";
-import { Badge } from "@/components/ui/badge";
 import {
   Select,
   SelectContent,
@@ -35,8 +34,6 @@ import {
   MoreHorizontal,
   LogOut,
   Copy,
-  Handshake,
-  Vote,
   Clock,
   Pause,
   Play,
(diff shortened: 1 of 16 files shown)


JorenC:
Draw proposals is now moved here:
image
or 
image

This will open the 'draw proposals' page, after which the user can add a new 'draw proposal' (mobile only for reference):
image
and
image

The only UX quirck is that the back button moves back to the top layer - which I believe is fine. In draw proposals it's supposed to be that, and in the other one it's an easy recoverable interaction.
---

I added outlines to the flags EVERYWHERE:
image

This both reinforces the player colour (flag might be different colour than the player on the map), and ensures white flags (like England) don't bleed into the background.

----

Staging games now have an easy, extra "Share" CTA - good for players who miss the share overflow option, so they can invite their link (and promote virality)
image

This does not appear in non-pending or non-joined games: 
image

only if the player has a pending game (created or joined), it's a CTA to share it with their friends and invite people.
the overflow menu entry is still there: 
image

which means it's technically a double entry, but this keeps it simple and still allows people to share games they are already in or have ended - for future use when games are visible publicly.

JorenC:
Updated as per Johns feedback - "Join" button is now only one. When joined, this button is split into two - "Leave" and "Share".
image
and 
image

PR #537, opened 2026-06-01
Background tasks: phase resolution and notifications

johnpooch:
Moves phase resolution and push notifications off the user-request path onto the existing Procrastinate worker, so a phase resolves within ~2s of the last confirmation and ~10s of its deadline instead of waiting up to a minute for the resolution cron. Game outcomes are unchanged — only *when* and *how reliably* work runs.

Two commits, reviewable independently.

## Resolution (`d703960`)

Three triggers share one predicate (`filter_due_phases`) and one guarded action (`resolve_if_due`):

- **Confirm** — the confirm toggle defers a `resolve_phase` job in the same transaction as the `orders_confirmed` write, so a confirmation can never be recorded without its resolve job.
- **Deadline** — a `Phase` `post_save` chokepoint arms a fresh `resolve_phase` job (`schedule_at`, per-game lock) whenever an active phase's `scheduled_resolution` moves to a future time. It never cancels; a stale early-firing job finds the phase not due and no-ops.
- **Sweep** — an `@app.periodic` task runs once a minute as a safety net and emits a Sentry canary for any phase overdue beyond `RESOLUTION_CANARY_GRACE_SECONDS` (default 300) before resolving it.

`resolve_if_due` locks the phase row (`select_for_update`) and re-checks `filter_due_phases` inside `transaction.atomic`, so double-resolve, a moved deadline, and a double NMR extension all no-op — holding even against the legacy cron and sandbox paths, which the per-game Procrastinate lock does not cover. The legacy `POST /phase/resolve/` cron now routes through `resolve_if_due` too, so it stays safe during cutover.

## Reminders + notifications (`49e1fb1`)

- Deadline reminders run as an `@app.periodic` task. `PhaseState.deadline_warning_sent_for` stores the `scheduled_resolution` a reminder was sent for; the sweep sends only when it differs from the phase's current value — at most one reminder per deadline, with an automatic fresh reminder when a deadline is extended, and no reset hooks. Fixed-time games remind players who still have orders to give.
- Chat-message and game-start notifications defer a `notification.send_notification` task instead of sending FCM synchronously. Resolution-related notifications already run on the worker via the resolve task.

## Cutover

Both `POST /phase/resolve/` and `POST /phase/deadline-warnings/` endpoints and their crons stay in place. Once the worker tasks are confirmed running in prod (logs/Sentry), a follow-up will delete the crons and the endpoints.


service/notification/signals.py
```diff
@@ -29,34 +29,36 @@ def send_channel_message_notification(sender, instance, created, **kwargs):
...
+    sender_name = instance.sender.name
+    game = instance.channel.game
+    current_phase = game.phases.last()
+    link = (
+        f"https://diplicity.com/game/{game.id}/phase/{current_phase.id}/chat/channel/{instance.channel.id}"
+        if current_phase
+        else f"https://diplicity.com/game/{game.id}"
+    )
+
+    from notification.tasks import send_notification
```
johnpooch: Does this import have to happen inline or can it be moved to the top of the file?

service/phase/serializers.py
```diff
@@ -20,8 +21,17 @@ class PhaseStateSerializer(serializers.Serializer):
...
 
     def update(self, instance, validated_data):
-        instance.orders_confirmed = not instance.orders_confirmed
-        instance.save()
+        with transaction.atomic():
+            instance.orders_confirmed = not instance.orders_confirmed
+            instance.save()
+
+            if instance.orders_confirmed:
+                from phase.tasks import resolve_phase
```
johnpooch: Does this import have to happen inline or can it be moved to the top of the file?

service/phase/signals.py
```diff
@@ -0,0 +1,53 @@
...
+    if scheduled_resolution is None:
+        return
+
+    if not created and scheduled_resolution == old_scheduled_resolution:
+        return
+
+    if scheduled_resolution <= timezone.now():
+        return
+
+    from phase.tasks import resolve_phase
```
johnpooch: Does this import have to happen inline or can it be moved to the top of the file?

PR #547, opened 2026-06-02
Add Sign in with Apple to the web app

alastairstuart:
## Summary

Adds **Sign in with Apple** to the web app, alongside the existing Google sign-in. Native iOS already supported Apple sign-in; this brings the same to the browser. Web and iOS Apple sign-ins resolve to the same account.

## How it works

- **Backend** — `verify_apple_id_token` now accepts an additional audience via a new `APPLE_WEB_CLIENT_ID` setting, mirroring the existing Google `_get_valid_client_ids()` multi-audience pattern. This is required because a web Apple sign-in uses an Apple **Services ID** as the OAuth `client_id`, so the token's `aud` differs from the iOS bundle id. The `auth/apple-login/` endpoint and serializer are unchanged.
- **Frontend** — adds `react-apple-signin-auth` (mirrors how web Google uses `@react-oauth/google`; native continues to use the capgo plugin). A new `webAppleAuth.ts` obtains the `id_token` via the Apple JS SDK popup, and `Login.tsx` renders an Apple button on web that reuses the existing apple-login mutation + `login()` flow (shared `finishAppleLogin` helper across native and web).

## Config / deployment notes

Requires an Apple **Services ID** (with Sign in with Apple enabled, the web domain + Return URL registered) and these env vars:

- Backend: `APPLE_WEB_CLIENT_ID` = the Services ID
- Frontend: `VITE_APPLE_WEB_CLIENT_ID` = the Services ID, `VITE_APPLE_REDIRECT_URI` = the registered Return URL

Without `APPLE_WEB_CLIENT_ID` set, behaviour is unchanged (the audience list contains only the iOS bundle id). New vars are documented in `.example.env` and `packages/web/.env.release.example`.

## Testing

- Backend: `service/login/tests_apple_utils.py` (new) signs real RS256 tokens to exercise the multi-audience logic — web Services ID accepted, iOS bundle id still accepted, unknown audience rejected. Full `login` suite passes.
- Frontend: `Login.test.tsx` covers the Apple button rendering and the success path (mutation + login called). `eslint` and `tsc` clean.
- **Manual end-to-end:** verified on a dev build (Cloudflare tunnel + own Apple Services ID) — real Apple popup sign-in completes and logs in via `auth/apple-login/`.

Draft for now pending review of the approach and the Services ID / env setup on the deployment side.


johnpooch:
Great work on this — the backend changes are clean and the test coverage is solid.

One thing I'd like you to change: `webAppleAuth.ts` should be inlined into `Login.tsx`. The Google web auth equivalent lives entirely inline in that file (no `webGoogleAuth.ts`), and Apple's web auth follows the same shape — a response arrives from a library component, you parse the credentials, then call the backend. Per the project convention, helpers and types that are only used in one place stay in that file rather than being extracted. Everything in `webAppleAuth.ts` (`webAppleAuthOptions`, `parseAppleResponse`, and the types) can move inline.

alastairstuart:
Thanks! Re: inlining `webAppleAuth.ts` ; I see your `9095c77` already covers it.

I've synced the branch with the latest `main` (it was behind by the pip-bump). `tsc`, eslint, and the Login tests all pass on the current state.

Issue #552, opened 2026-06-03
Show end-game screen


JorenC:
There is nothing to re-affirm a game ended. 

Now, it shows a (win/draw/loss) dialog every time you open the game (first time will be very good - don't expect people to open old games VERY often, minor annoyance, but a good reminder of the game output) - that shows the state of the game: 

<img width="1882" height="941" alt="Image" src="https://github.com/user-attachments/assets/9ff31496-87bb-4b73-aa84-3b89729b7af1" />
<img width="474" height="842" alt="Image" src="https://github.com/user-attachments/assets/2df7d3db-8b01-4b60-b2ea-a858a4ba33d5" />
<img width="1910" height="953" alt="Image" src="https://github.com/user-attachments/assets/63d1443d-f6f1-4b97-b753-113973d27320" />
<img width="1920" height="970" alt="Image" src="https://github.com/user-attachments/assets/1f0dd0b8-bbfa-4fb8-86c2-a74bbce728b7" />
<img width="1920" height="970" alt="Image" src="https://github.com/user-attachments/assets/c010934f-e934-4171-8304-7e3d67d8418a" />

PR #553, opened 2026-06-03
Game-end dialog

JorenC:
_This will now always open when a finished game is opened. To me, that is good behaviour - it adds another step, but it's also a reminder this game has ended. I don't believe people will be so bothered by going back to their old games so often that we need to adjust - we can complicate it with some form of counter in that case, but let's keep it simple for now. If game = ended then show results dialog when game is opened - simple, works always and everyone sees the right thing_

** note, I mocked up these results with claude so the surviving powers is not perfect - it will separate surviving powers from eliminated powers. **

Draw:
image

Draw mobile:
image

---

Victory mobile (how scrolling works):
image

Victory desktop:
image

----

Defeat desktop: (I belief you understand how mobile will look):
image

## Summary

- Adds a `GameEndDialog` component that shows a full-screen overlay when a player visits a completed game
- Displays victory/draw/defeat result with background imagery, a flavour blurb, and a supply-centre leaderboard
- For draw games, splits the leaderboard into "Surviving powers" and "Eliminated" sections

## Test plan

- [ ] Open a completed game as the solo winner → Victory dialog with flat SC leaderboard
- [ ] Open a completed game as a defeated player (solo win) → Defeat dialog with flat SC leaderboard
- [ ] Open a completed game that ended in a draw, as a surviving power → Draw dialog with survivor/eliminated sections
- [ ] Open a completed game that ended in a draw, as an eliminated player → Defeat dialog with survivor/eliminated sections
- [ ] Dismiss the dialog → it stays dismissed for the session; reopening the same game remounts and shows it again (expected)
- [ ] Open an active game → no dialog


JorenC:
Fixes #552

JorenC:
Updated screenshots, better design

johnpooch:
@JorenC I'm going to close this pull request not because it is a bad idea or anything like that, but because it has gone stale. 

I like the idea of having a game result dialogue. I'm unsure about the images that we're using in the dialogue. I don't know if it makes sense for variance beyond the classical variant. I also think the imagery is a little bit depressing, but that's just a personal opinion. Also, the caption that appears in the dialogue is Europe specific, which I don't think makes sense for other variants. 

My preference would be to make the dialogue a little bit more simple by removing stuff that relates to Europe and the classical variant. 

I would also recommend that as much of the logic in the dialog as possible is moved onto the backend so that the dialog is a dumb renderer. 

Another thought, though I suspect you'll disagree with me. Generally speaking, I don't like dialogues very much. I find them to be a kind of awkward UI. I wonder if we could use the left panel to render the game results whenever a game is finished. I think on mobile there's definitely space for an additional bottom tab when the game is completed, and it would definitely fit on desktop.

PR #571, opened 2026-06-04
Show map preview on variant cards

johnpooch:
Variant cards on the variants list now show a map preview. On mobile the preview spans the full card width above the content; on larger screens it sits to the left as a fixed-width column.

Variants without an SVG render without the preview section.

<sub>Generated with [Claude Code](https://claude.com/claude-code)</sub>


johnpooch:
Work merged directly to main.

PR #615, opened 2026-06-08
UX a day: Preserve navigation state when switching between map, orders, and chat

JorenC:
I implemented this because it's a very good desktop improvement - we had this in the old Dip (based on user feedback), it makes logical sense, and I just got it re-affirmed/reminded by this beta tester. Sounded like a good 'UX a day' small improvement:
https://github.com/johnpooch/diplicity-react/discussions/614

This feature makes switching back and forth between the map while being in a conversation much easier — you can quickly jump to the map to check a position and return to exactly where you were in chat without losing your place.

## Changes

** Allow multi-line messages
- On mobile, enter now adds a newline. Pressing the 'send' button sends it. 
- On desktop, alt+enter now adds a newline. Pressing 'enter' sends it. 
image
image

**Chat tab remembers your position**
- The Chat tab now returns you to the last channel you were viewing (e.g. the Italy chat), not always the channel list root
- The Chat tab highlights correctly when you're inside a channel (previously it appeared inactive on sub-routes)

**Map preserves zoom and pan (mobile only)**
- When you switch away from the Map tab and return, your zoom level and pan position are restored
- The cache is keyed per game + phase, so switching games or phases always starts centered

**Chat input focus is restored**
- If the text input had focus (keyboard open on Android) when you left the channel, focus is returned when you come back (note - the keyboard will not open on iOS, it will just focus on the field but iOS does not allow keyboard restore)
- If it didn't have focus, it stays unfocused — no unwanted keyboard pop-ups

**Draft messages unit location**
- When focus is restored on a draft message, the unit cursor is placed at the end instead of at the start.

## Test plan
- [ ] Open a channel, scroll to a specific message, switch to Map, switch back to Chat — confirm you land in the same channel
- [ ] Pan and zoom the map, switch to Chat/Orders, switch back to Map — confirm position is preserved
- [ ] Focus the chat input, switch to Map, switch back — confirm keyboard reappears (Android)
- [ ] Do not focus the chat input, switch away and back — confirm keyboard does not appear
- [ ] Switch between two different games — confirm map position resets correctly for each

### DIFF
--- a/packages/web/src/components/GameDetailLayout.tsx
+++ b/packages/web/src/components/GameDetailLayout.tsx
@@ -1,5 +1,5 @@
-import React, { useMemo, useRef, useEffect } from "react";
-import { useLocation, useNavigate } from "react-router";
+import React, { useMemo } from "react";
+import { useLocation, useNavigate, useSearchParams } from "react-router";
 import { useRequiredParams } from "@/hooks";
 import { ArrowLeft, Map, Gavel, MessageCircle } from "lucide-react";
 import { cn } from "@/lib/utils";
@@ -45,14 +45,13 @@ const GameDetailLayout: React.FC<GameDetailLayoutProps> = ({
     },
   });
 
-  const lastChatPathRef = useRef<string | null>(null);
+  const [searchParams] = useSearchParams();
   const chatPrefix = `/game/${gameId}/phase/${phaseId}/chat`;
 
-  useEffect(() => {
-    if (location.pathname.startsWith(chatPrefix)) {
-      lastChatPathRef.current = location.pathname;
-    }
-  }, [location.pathname, chatPrefix]);
+  const lastChatChannelId =
+    location.pathname.startsWith(chatPrefix + "/")
+      ? location.pathname.slice(chatPrefix.length + 1)
+      : searchParams.get("chatChannel");
 
   const navItems = useMemo(() => {
     const items = game?.sandbox
@@ -69,9 +68,9 @@ const GameDetailLayout: React.FC<GameDetailLayoutProps> = ({
           ? "•"
           : undefined;
       const isChat = item.label === "Chat";
-      const lastChatPath = lastChatPathRef.current;
-      const validChatPath = lastChatPath?.startsWith(chatPrefix) ? lastChatPath : null;
-      const clickPath = isChat && validChatPath ? validChatPath : basePath;
+      const clickPath = isChat
+        ? lastChatChannelId ? `${chatPrefix}/${lastChatChannelId}` : basePath
+        : lastChatChannelId ? `${basePath}?chatChannel=${lastChatChannelId}` : basePath;
       const isActive = isChat
         ? location.pathname === basePath || location.pathname.startsWith(basePath + "/")
         : location.pathname === basePath;
@@ -82,7 +81,7 @@ const GameDetailLayout: React.FC<GameDetailLayoutProps> = ({
         badge,
       };
     });
-  }, [gameId, phaseId, chatPrefix, location.pathname, game?.totalUnreadMessageCount, game?.sandbox]);
+  }, [gameId, phaseId, chatPrefix, lastChatChannelId, location.pathname, game?.totalUnreadMessageCount, game?.sandbox]);
 
   // Filter out Map for desktop sidebar since map is already visible in right panel
   const sidebarNavItems = useMemo(
--- a/packages/web/src/components/GameDetailLayout.tsx
+++ b/packages/web/src/components/GameDetailLayout.tsx
@@ -46,16 +46,15 @@ const GameDetailLayout: React.FC<GameDetailLayoutProps> = ({
   });
 
   const lastChatPathRef = useRef<string | null>(null);
+  const chatPrefix = `/game/${gameId}/phase/${phaseId}/chat`;
 
   useEffect(() => {
-    const chatPrefix = `/game/${gameId}/phase/${phaseId}/chat`;
     if (location.pathname.startsWith(chatPrefix)) {
       lastChatPathRef.current = location.pathname;
     }
-  }, [location.pathname, gameId, phaseId]);
+  }, [location.pathname, chatPrefix]);
 
   const navItems = useMemo(() => {
-    const chatPrefix = `/game/${gameId}/phase/${phaseId}/chat`;
     const items = game?.sandbox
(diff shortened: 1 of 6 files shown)


johnpooch:
Thanks for the work here — the underlying features are all reasonable asks and the intent is right. But I think the implementation has enough issues that I'd like to see it restructured before merging. Here's my specific feedback on each piece:

---

**Chat tab remembers last channel**

The feature is worth having. The implementation isn't quite right though — reading `lastChatPathRef.current` inside `useMemo` is a React anti-pattern. `useMemo` is a pure computation; reading mutable ref state inside it is fragile (it works here by coincidence because the dep array includes `location.pathname`, which happens to also trigger the `useEffect` that sets the ref, but this is easy to break and confuses future readers).

The simpler correct approach is localStorage: write `lastChatChannel-${gameId}` when navigating into a channel, read it back when building the Chat nav item's click path. No refs, no `useEffect`, no useMemo side effects. Alternatively, a URL search param would also fit how this codebase handles navigation state.

My personal recommendation would be a URL search param because it is more consistent with what we do in other parts of the app.

---

**Multi-line messages**

Switching to `<Textarea>` is fine. But `Alt+Enter` is the wrong key binding — `Shift+Enter` is the universal convention (Slack, Discord, Teams, WhatsApp Web). Alt+Enter comes from spreadsheet apps and will feel wrong to almost every user.

More importantly: if you use `Shift+Enter`, you don't need any of the manual cursor manipulation. The browser handles the newline insertion and cursor placement natively for a `<textarea>`. The `requestAnimationFrame` + `selectionStart`/`selectionEnd` block can be deleted entirely. `handleKeyDown` becomes:

```ts
const handleKeyDown = (e: React.KeyboardEvent<HTMLTextAreaElement>) => {
  if (e.key === "Enter" && !e.shiftKey && !e.nativeEvent.isComposing) {
    e.preventDefault();
    handleSubmit();
  }
};
```

No `isMobileDevice` check is needed either — `Shift+Enter` works consistently across desktop and mobile (on mobile, Enter doesn't submit anyway, so the guard is only needed on desktop where it does). And we don't need the "Alt+Enter to add a new line" hint label at all — `Shift+Enter` is expected behaviour that doesn't need documenting.

---

**Focus restoration (`channelFocusCache`)**

This is the most complex piece in the PR and I'd rather drop it from scope. The module-level Map, the LRU eviction loop, the `isFocusedRef`, the `onFocus`/`onBlur` handlers — it's a lot of machinery for a narrow Android-specific UX improvement. The same LRU eviction pattern is also duplicated in `InteractiveMapZoomWrapper`, which is a signal that this wants to be a shared abstraction if it exists at all.

I'd suggest removing this from the PR and revisiting it separately if it's a high-priority ask.

---

**Map zoom/pan caching**

Same concern as focus restoration — I'd drop it from this PR. If we do add it later, localStorage is the better home for it (survives tab switches and page refreshes, doesn't require passing a `cacheKey` prop into `InteractiveMapZoomWrapper`). Adding cache management logic to a component that previously had none is a coupling we should be deliberate about.

---

**Summary**

Suggested scope for a revised version:
- Chat tab navigates back to the last channel (via localStorage)
- Textarea with Shift+Enter newlines (no cursor manipulation, no hint label, no device detection)
- Drop focus restoration and zoom/pan caching

That's a small, clean PR with no fragile internals. Happy to discuss any of this.

JorenC:
Check - updated. I removed the alt+enter hint - shit+enter could just be normally 'discovered'. And if missed, not the end of the world.

johnpooch:
Thanks for addressing some of the earlier feedback. A few remaining issues, then a suggestion on how to move forward.

---

**Transform cache — still in**

The `transformCache` in `InteractiveMapZoomWrapper` wasn't addressed. My position on this hasn't changed: it adds coupling to a component that previously had no caching concerns, and the LRU eviction pattern it introduces is the same one duplicated in the focus cache. I'd drop this from scope entirely.

**Focus cache — still in**

Same position: the `channelFocusCache` / `isFocusedRef` / `handleInputFocus` / `handleInputBlur` block is a lot of machinery for a narrow Android UX improvement. I'd drop it.

---

**Chat tab navigation**

The updated approach (using search params to carry `chatChannel` forward) is closer to right, but I think it's still more complex than it needs to be.

`GameDetailLayout` shouldn't contain any chat-specific logic beyond fixing the `isActive` check for sub-routes. The `lastChatChannelId` derivation, `chatPrefix`, and the `clickPath` conditional can all go away.

The cleaner model:

- Nav items preserve existing search params when navigating between tabs — this is generic behaviour, not specific to chat.
- `ChannelScreen` sets the channel ID as a search param on mount.
- The chat route reads that param and navigates to the right channel.

On naming: the search param should be `channelId`, not `chatChannel`. The phase is a path param called `phaseId`; the channel is a path param called `channelId`. Search params used to remember these values should follow the same names.

---

**Multi-line messages**

`Shift+Enter` was adopted — good. The hint label was dropped — also good, since `Shift+Enter` is expected behaviour that doesn't need documenting.

---

**Suggestion: split the PR**

This PR does several independent things and the discussion so far reflects that — each feature has its own issues and reviewability. I'd suggest splitting into individual PRs:

1. **Multi-line messages** — `Textarea` + `Shift+Enter` + `whitespace-pre-wrap`. This is the cleanest piece and ready to land with minimal changes.
2. **Chat tab remembers last channel** — the simplified search-param approach described above, without the chat-specific logic in `GameDetailLayout`.
3. **Map zoom/pan caching** — if this is still wanted, a separate PR makes the scope and trade-offs easier to evaluate. My suggestion would be localStorage rather than a module-level Map in the component.
4. **Focus restoration** — separate PR if it's still wanted, but I'd consider dropping it.

JorenC:
Addressed all feedback from the latest review:

**Dropped from scope**
- Map zoom/pan caching (`transformCache`, `cacheKey` prop) — removed entirely
- Focus restoration (`channelFocusCache`, `isFocusedRef`, focus/blur handlers) — removed entirely

**Chat tab navigation — simplified**
- `GameDetailLayout` no longer has any chat-specific logic beyond the `isActive` sub-route fix
- `ChannelScreen` sets `?channelId=<id>` as a search param on mount (replace, no history entry)
- Nav items generically preserve existing search params when switching tabs
- `ChannelListScreenSuspense` redirects to the last channel when `channelId` param is present
- Search param renamed from `chatChannel` to `channelId`

**Multi-line messages** — unchanged from previous round (Shift+Enter, no hint label)

PR #629, opened 2026-06-11
Non-Player GM Mode

JorenC:
## Summary

Introduces a **non-playing Game Master** role for private games. The creator can opt in via a checkbox in the Basic Settings section of the Create Game form and will join as a GM member: they participate in chat and can pause/resume the game, but receive no nation or orders and are excluded from all gameplay mechanics.

### Key changes

**Backend**
- New `non_playing_gm` boolean field on `Game` (migration `0016`), toggled via `CreateGameSerializer`
- `IsPlayingMember` permission class gates anything nation-specific (order submission/deletion, phase confirmation, draw-proposal voting)
- GM members are excluded from: nation assignment, space-available count, game-start trigger, solo/draw winner checks, NMR/abandonment detection, push-notification recipient lists
- GM is shown as a member with a game-master SVG avatar and "Game Master" name, even in masked games
- Leaving logic: if the GM leaves a pending game with other players, the first-joined player is promoted to a playing GM; if no other players exist the game is deleted
- Draw proposals: GM can create and cancel proposals (`IsActiveGameMember`) but cannot vote (`IsPlayingMember`)

**Frontend**
- "Game Master" checkbox moved to the Basic Settings section of the Create Game form
- Wired to the `non_playing_gm` field in the API payload

### Test coverage
- `service/game/tests/test_non_playing_gm.py` — 20+ cases covering permissions, member display, abandonment, leaving/promotion, draw proposals, and order/phase access

## Test plan

- [ ] Create a private game with "Game Master" checked; verify GM member appears with game-master avatar and no nation
- [ ] Join as a second player; verify game starts and GM is not counted toward the player limit
- [ ] As GM, try to submit/delete an order and confirm a phase — expect 403
- [ ] As GM, create and cancel a draw proposal — expect success; vote — expect 403
- [ ] As GM, leave a pending game with other players — verify first player is promoted to playing GM
- [ ] Backend tests pass: `docker compose run --rm service python3 -m pytest service/game/tests/test_non_playing_gm.py -v`


JorenC:
@johnpooch I think this is ready to pull.. There is a database migration added, so please check-double check.

- Only private games can have a game master (public is too open to give one this power - the GM is supposed to find players other ways, e.g. tournament/discord).
- Game Master option in create-game (whether invoked through new game or 'private game' in the variant list) now allows the player not to join.
- If a GM leaves a pending game as last player, game is destroyed. If there is one player but not started, the first player to join that game becomes player/GM - as normal now. 
- GM has normal GM powers, (pause/extend), nothing next now.
- GAmes with a GM start with N+1 player, to account for hte extra player (GM is taking a player slot).

- the GM is part of public press, and can be part of any channel. They can be invited by any player - just as a normal player.
- the GM in anonymous games is not anonymous - Claude might see this as a leak, but it's intended - no need for GM to be anonymous. 
- GM gets notified of CD and draw proposal notifications.
- GM can actively create a draw proposal (initiate it), but they can't vote on it - they're not part of the vote. They do see votes coming in of other players making a proposal, just can't vote. **I don't know how a draw with a GM is counted - they might be counted as part of the draw, but not important now - might be important later with statistics, but lets leave that for MVP+ - fix it when it becomes an issue - OR if you can see this, you can fix it now**
- In the player screens, GM is just shown as other players. When a game is started, they have the 'Game Master' nation (although it's not officially a nation).
- In the Chat/nations, the GM has a specific 'Diplciity logo' icon, instead of a flag. 

I think this is it. I did not see any major risks, but only checked with 4.6, not 4.8.

Please take a good look, and if you agree or see improvements, feel free to adjust and then pull - no need to send it back to me and let me improve it - unless that helps you much more, but if the bot is doing it just as easy, please go ahead and just implement. 

Cheers!

JorenC:
Active (outdated) deployment - maybe update the deployment before you check...

packages/web/src/components/GameCard.tsx
```diff
@@ -164,7 +164,7 @@ const GameCard: React.FC<GameCardProps> = ({ game, variant, map }) => {
                   key={member.id}
                   className="h-8 w-8 border-2 border-background"
                 >
-                  <AvatarImage src={member.picture ?? undefined} />
+                  <AvatarImage src={member.isGameMaster && !member.nation ? undefined : (member.picture ?? undefined)} />
```
johnpooch, line 167: The cleaner fix here is to filter the Game Master out on line 162 before iterating.

packages/web/src/components/PlayerInfoContent.tsx
johnpooch, line 52: I think we should restructure this component as follows:

Firstly, we should extract all of the pre-existing code from line 53 to line 130 into a new `PlayerCard` component. 

Secondly, I think we should add a separate `GameMasterCard` component. We should render the Game Master card outside of the for loop. This prevents convoluted conditional logic within the for loop. We would filter out the Game Master before rendering the player cards.

packages/web/src/components/PlayerInfoContent.tsx
```diff
@@ -57,15 +58,22 @@ export const PlayerInfoContent: React.FC = () => {
...
                   <NationFlag
                     flagUrl={findNationFlagUrl(variant.nations, member.nation)}
                     alt={member.nation}
                     size="lg"
                     className="size-8"
                     color={findNationColor(variant.nations, member.nation)}
                   />
-                )}
+                ) : member.isGameMaster && hasNationFlags ? (
+                  <NationFlag
```
johnpooch, line 70: It feels wrong to me that we are using the `NationFlag` component for an image that is not a flag and does not belong to a nation.

packages/web/src/components/PlayerInfoContent.tsx
```diff
@@ -119,7 +127,9 @@ export const PlayerInfoContent: React.FC = () => {
                         )}
                       </span>
                     </div>
-                  )}
+                  ) : member.isGameMaster ? (
+                    <div className="text-sm text-muted-foreground mt-1">Game Master</div>
+                  ) : null}
```
johnpooch, line 132: Look at lines 94 to 99. We already have a badge which shows the game master. I think we need to tidy up the logic here. It would be solved by my suggestion above, creating two separate components, `PlayerCard` and `GameMasterCard`.

packages/web/src/screens/GameDetail/ChannelAvatar.tsx
```diff
@@ -22,13 +22,13 @@ const imgY = (d: number, ya: YAlign, yb: number): number => {
   return yb - d / 2;
 };
 
-interface FlagImg { url: string; x: number; y: number; d: number }
+interface FlagImg { url: string | null; color: string; x: number; y: number; d: number }
```
johnpooch, line 25: The changes in this file feel messy and convoluted. 

Do we really need to support having the Game Master be in private channels, or would it be sufficient to just have the Game Master be in the public press channel?

packages/web/src/screens/GameDetail/ChannelCreateScreen.tsx
```diff
@@ -90,19 +90,19 @@ const ChannelCreateScreen: React.FC = () => {
...
+                  <React.Fragment key={member.id}>
                     <Item
                       className="cursor-pointer hover:bg-accent/50"
                       onClick={() => handleToggle(member.id)}
                     >
                       <ItemMedia>
                         <Avatar>
                           <AvatarImage src={member.picture ?? undefined} />
-                          <AvatarFallback>{member.nation?.[0]}</AvatarFallback>
+                          <AvatarFallback>{(member.nation ?? member.name)?.[0]}</AvatarFallback>
```
johnpooch, line 101: In the `PlayerInfoContent` file, I proposed creating `PlayerCard` and `GameMasterCard` components. I think these components could be re-used in this file as well. We would need to make the cards selectable. 

My suggestion then would be to render the game master first and then render all of the other game members. 

I think we should consistently render the Game Master separate from the other members of the game.

packages/web/src/screens/GameDetail/ChannelListScreen.tsx
```diff
@@ -34,7 +34,7 @@ const getLatestMessagePreview = (
   const latestMessage = messages[messages.length - 1];
   const senderLabel = latestMessage.sender.isCurrentUser
     ? "You"
-    : latestMessage.sender.nation.name;
+    : (latestMessage.sender.nation?.name ?? "Game Master");
```
johnpooch, line 37: I don't love this conditional logic. We are inferring that the player is the game master based on the player's nation not having a name. It feels a bit fragile. We would be better off checking whether the sender is the game master explicitly.

packages/web/src/screens/GameDetail/ChannelListScreen.tsx
```diff
@@ -47,8 +47,9 @@ const ChannelListScreen: React.FC = () => {
   const { data: channels } = useGamesChannelsListSuspense(gameId);
   const { data: variants } = useVariantsListSuspense();
 
-  const currentNationName =
-    game.members.find(m => m.isCurrentUser)?.nation ?? undefined;
+  const currentMember = game.members.find(m => m.isCurrentUser);
+  const currentNationName = currentMember?.nation ?? undefined;
+  const currentIsGm = !!(currentMember?.isGameMaster && !currentMember.nation);
```
johnpooch, line 52: Throughout this pull request, we repeatedly make these checks to determine if the member is both the game master and doesn't have a nation, i.e., is a non-playing game master. 

I feel like we should separate the concept of the game master and the game creator. Mixing the two is introducing a lot of conditional logic.

packages/web/src/screens/GameDetail/ChannelScreen.tsx
```diff
@@ -259,13 +262,14 @@ const ChannelScreen: React.FC = () => {
                           <div className="w-8 flex-shrink-0 flex justify-center">
                             <NationFlag
                               flagUrl={
-                                variant
-                                  ? findNationFlagUrl(variant.nations, item.sender.nationName)
-                                  : null
+                                item.sender.nationName === null
+                                  ? (item.sender.picture ?? null)
+                                  : (hasNationFlags && variant ? findNationFlagUrl(variant.nations, item.sender.nationName) : null)
```
johnpooch, line 267: As mentioned before, we are having to introduce some ugly logic here. When the player is a non-playing game master, they don't have a nation and they don't have a flag, and yet we are using the nation flag component.

packages/web/src/screens/GameDetail/ChannelScreen.tsx
```diff
@@ -259,13 +262,14 @@ const ChannelScreen: React.FC = () => {
...
                               }
-                              alt={item.sender.nationName}
+                              alt={item.sender.nationName ?? "Game Master"}
```
johnpooch, line 269: This should be handled on the backend. Again, we are making a fragile claim here, inferring that if the sender does not have a nation, they must be the game master. I don't really like this.

packages/web/src/screens/GameDetail/channelUtils.ts
```diff
@@ -20,16 +20,28 @@ export const brightnessByColor = (hex: string): number => {
...
+  currentNationName: string | undefined,
+  currentIsGm: boolean
+): boolean => {
+  if (name === currentNationName) return false;
+  if (currentIsGm && name === "Game Master") return false;
+  return true;
+};
+
 // Private channel names are formatted by the backend as "Nation1, Nation2, ..."
 export const getChannelDisplayName = (
```
johnpooch, line 34: I don't really know why this function exists at all, given that the display name comes from the backend. Either way, I think we are letting the Game Master logic pollute the entire codebase.

packages/web/src/screens/GameDetail/DrawProposalsScreen.tsx
```diff
@@ -49,6 +49,7 @@ interface ProposalItemProps {
   game: GameRetrieve;
   currentMember: Member | undefined;
   variant: Variant | undefined;
+  hasNationFlags: boolean;
```
johnpooch, line 52: I don't really like this `hasNationFlags` property. 

It feels like messy pollution throughout the codebase. 

I feel like we should assume that all variants have nation flags. Variants which don't have nation flags will have weird rendering quirks, and that's fine. I really don't like the conditional logic being littered around everywhere.

packages/web/src/screens/GameDetail/DrawProposalsScreen.tsx
```diff
@@ -74,23 +76,27 @@ const ProposalItem: React.FC<ProposalItemProps> = ({
     .map(id => game.members.find(m => m.id === id))
     .filter((m): m is Member => Boolean(m));
 
-  const nationColor = variant
-    ? findNationColor(variant.nations, proposal.createdBy.nation)
-    : null;
+  const isGm = game.nonPlayingGm && proposal.createdBy.nation == null;
```
johnpooch, line 79: The game model should not have a non-playing GM field. It can be trivially inferred from the game members. 

However, as I read through this pull request, I'm pretty sure that by the end of it I'm gonna propose a different approach to the modeling entirely.

packages/web/src/screens/GameDetail/DrawProposalsScreen.tsx
```diff
@@ -74,23 +76,27 @@ const ProposalItem: React.FC<ProposalItemProps> = ({
...
+  const flagUrl = isGm
+    ? (hasNationFlags ? (proposal.createdBy.picture ?? null) : null)
+    : (variant ? findNationFlagUrl(variant.nations, proposal.createdBy.nation) : null);
+  const nationColor = isGm
+    ? null
+    : (variant ? findNationColor(variant.nations, proposal.createdBy.nation) : null);
```
johnpooch, line 85: This is more nasty conditional logic polluting the codebase. 

It is a code smell that points towards problems with our modeling. We need to get the modeling right on the backend and try to contain the messiness as much as possible.

packages/web/src/screens/GameDetail/DrawProposalsScreen.tsx
```diff
@@ -74,23 +76,27 @@ const ProposalItem: React.FC<ProposalItemProps> = ({
...
+        alt={proposal.createdBy.nation ?? proposal.createdBy.name}
         size="md"
         className="size-8 self-start"
         color={nationColor}
       />
       <ItemContent>
         <div className="flex items-center justify-between">
           <div className="flex items-center gap-2">
-            <ItemTitle>Proposed by {proposal.createdBy.nation}</ItemTitle>
+            <ItemTitle>Proposed by {proposal.createdBy.nation ?? proposal.createdBy.name}</ItemTitle>
```
johnpooch, line 99: Do you see what I mean about the conditional logic polluting the codebase? We're having to insert conditions everywhere to support this feature. 

That's usually a giveaway that there is something wrong at a more fundamental level.

packages/web/src/screens/GameDetail/ProposeDrawScreen.tsx
```diff
@@ -45,6 +46,20 @@ const ProposeDrawScreen: React.FC = () => {
...
+    if (game.nonPlayingGm && member.isGameMaster) {
+      return hasNationFlags ? (member.picture ?? null) : null;
+    }
+    return variant ? findNationFlagUrl(variant.nations, member.nation) : null;
+  };
+
+  const getMemberFlagColor = (member: Member) => {
+    if (game.nonPlayingGm && member.isGameMaster) return null;
+    return variant ? findNationColor(variant.nations, member.nation) : null;
+  };
```
johnpooch, line 61: This is more conditional logic polluting the codebase.

service/channel/models.py
```diff
@@ -43,15 +43,19 @@ def with_unread_counts(self, user):
         )
 
 
+def channel_name_for_members(members):
+    nations = sorted([m.nation.name if m.nation else "Game Master" for m in members])
```
johnpooch, line 47: This is a little bit ugly, and it does make me question whether we really want the Game Master to be a member of private chats.

johnpooch (changes requested):
This is a very large pull request, and it introduces a large number of conditional logic branches. I have a few high-level thoughts about this pull request. 

**Break it down**

My first suggestion would be to break this pull request down into the minimal viable product, and then follow up with enhancements. 

To me, the initial implementation of this feature should not allow the game master to participate in chat channels at all, or to propose draws. These can be easily added in follow-up PRs. 

If you remove these two capabilities, this initial PR will be much more reviewable and much safer. 

**Dishonest/ambiguous modeling**

There is a fundamental problem about what we are doing in this PR. The modeling is either dishonest or ambiguous for multiple reasons.

We are stretching the definition of the term "member" to include potentially non-playing users. The knock-on effects of this cause pollution throughout the codebase. 

We are also conflating the existing game master functionality (which is really just "Game Creator") with a new non-playing game master concept.

I think we should make the modeling more honest and explicit:

- The person who creates the game should no longer be considered the "game master" by default. Instead, they should just be considered the "game creator" Which may grant them some capabilities like being able to remove other players.

- The non-playing game master should not be considered a member of the game. Instead, we should add a new Game Master foreign key field to game. 

- For consistency, we might want to create a created_by far and key field on game.

If we do these three suggestions, we end up with three clear entities:

- Member: A player in the game - always has a nation
- Game Creator: The user that created the game. When the game has a game master, "Game Creator" and "Game Master" will always be the same user.
- Game Master: Non-playing game master. Not a member.

**Contain the ugliness**

When we get around to implementing the draw proposal stuff and the chat channel stuff, there is inevitably some ugliness introduced. 

Right now, chat messages and chat channels can only be created by game members, but we will be extending the rules to allow the game master to create these entities. That inevitably introduces a little bit of messiness, but the messiness should be isolated to the backend and should not pollute the rest of the codebase. 

Same with draw proposals. It is inherently messy that draw proposals can now be created by game masters and members, but can only include members. But the impact of this messiness can be reduced by isolating it to the backend controller. 

---

To get the modeling into the right place as I've described will require doing some initial preparatory PRs.

We need to create migrations to backfill previous games so that they are in the expected state. For example, changing game masters to game creators, etc. 

Only when the modeling is in a really good state can the rest of the work be carried out. 

This is just inherently one of those complex features that requires a careful, iterative approach. I see it being broken down into five or six PRs, each of which can be easily reviewed.

Issue #690, opened 2026-06-13
Epic: Player replacement (voluntary substitution + takeover)


johnpooch:
## Summary

Allow a player to be substituted out of an active game and replaced by another user via a shareable link. Two entry points lead to one takeover mechanism:

- **Voluntary** — a player marks their seat `seeking_replacement` (e.g. they want to find a sub via Discord). They keep playing normally until someone takes over.
- **Involuntary** — a player already in `civil_disorder` (existing system) becomes claimable too.

Discovery is **link-only** (not surfaced in Find Games), takeover is **pure self-service** (no approval, no creator force), and eligibility is simply "not already a member of this game".

## Key design decisions (from design discussion)

- **New member, retire old.** On takeover we create a fresh `Member` for the newcomer and mark the outgoing member with a `replaced_by` FK to the successor. This preserves per-tenure history and keeps the existing derivation-based reliability/stats feature (`user_profile.utils.get_player_stats`) correct — reassigning `member.user` would silently re-attribute the predecessor's NMR/CD record to the newcomer (exploitable). The `replaced_by` chain also doubles as the seat's tenure lineage.
- **Distinct voluntary field.** New `Member.seeking_replacement` boolean, kept separate from `civil_disorder` (voluntary is not punitive). Derived umbrella property `replaceable = (civil_disorder or seeking_replacement) and not (eliminated or kicked or replaced_by)`.
- **"Active member" filtering.** `eliminated`/`kicked` already mark non-active rows; we add `replaced_by__isnull=True` to the same active-member querysets and permission checks so a retired member isn't treated as the seat's occupant.
- **Voluntarily-replaceable player keeps playing** — real orders, normal deadlines, normal draw votes — until actually handed over.
- **Chat history transfers** to the newcomer (new `ChannelMember` rows across the game's channels).
- **On takeover:** clear `civil_disorder`/`seeking_replacement`, reset `nmr_extensions_remaining` to the game default, carry the in-progress current-phase orders over (re-openable).

## Milestones

Leanest demonstrably-useful first milestone is voluntary replacement end-to-end: sub-issues **1–4** (backend) + **7–8** (frontend). CD takeover (5) and private-link hardening (6) are follow-ups.

Sub-issues are listed below.

johnpooch:
Superseded by RFC #782 (https://github.com/johnpooch/diplicity-react/discussions/782), which captures this epic and all of its sub-issues. Closing in favour of the RFC.

PR #700, opened 2026-06-13
[1/9] Backend: Member model fields + serializer (seeking_replacement, replaced_by, replaceable)

johnpooch:
Part of #690.

Foundation only — no behavior change.

## Changes
- Add to `service/member/models.py`:
  - `seeking_replacement = BooleanField(default=False)`
  - `replaced_by = ForeignKey("self", null=True, blank=True, on_delete=SET_NULL, related_name="replaces")`
  - derived property `replaceable = (civil_disorder or seeking_replacement) and not (eliminated or kicked or replaced_by_id)`
- Migration.
- Expose `seeking_replacement` and `replaceable` on `MemberSerializer`.
- Update `service/openapi-schema.yaml` and regenerate `packages/web/src/api/generated/endpoints.ts` so the frontend `Member` type gains `seekingReplacement` and `replaceable`.

## Tests
- API-level tests in `TestReplaceableSerialization` covering key flag combinations for `replaceable`, following the `TestCivilDisorderSerialization` pattern.

## Ships alone
Yes — inert fields, read side only.


johnpooch:
## PR Review

### What the PR does

The PR adds two new fields to the `Member` model (`seeking_replacement` boolean and `replaced_by` FK to self) plus a computed `replaceable` property that is `True` when a member has civil disorder or is seeking replacement but is not eliminated, kicked, or already replaced. These are wired through the serializer and codegen. The PR also removes the `/devices/` FCM endpoints and `/api/test-sentry/` from the OpenAPI schema and the generated TypeScript client.

---

### Does it make sense?

**Member replacement tracking:** The model change is logically coherent. `seeking_replacement` and `replaced_by` are clean additions. The `replaceable` property correctly uses `self.replaced_by_id is not None` (avoiding a DB hit) and combines the conditions correctly. The migration is auto-generated and safe (both fields are nullable/default-false). The serializer wiring is correct and matches the existing `civil_disorder` pattern exactly.

**FCM/test-sentry removal:** This is where the PR's intent is unclear. The changes are *only* in `service/openapi-schema.yaml` and `packages/web/src/api/generated/endpoints.ts` — no Django `urls.py`, `settings.py`, or view files are touched. `CLAUDE.md` explicitly documents that codegen in a cloud session without Firebase configured produces *exactly* this diff as an **environmental artifact**: "A clean git diff after codegen in a cloud session (no Firebase, DEBUG off) shows only the /devices/ + FCMDevice removal; that is environmental, not a stale-checkout signal." So the question is: are these changes a real, permanent removal of FCM support (in which case Django-side code changes are missing), or are they just an environmental codegen artifact that shouldn't be committed? Currently this looks like the latter — the schema committed here would be inconsistent with what `docker compose up codegen` produces in a Firebase-configured environment.

**Single purpose:** The PR bundles a new feature (replacement tracking) with a cleanup (FCM removal). Under the project rubric these are unrelated pieces of work. However, if the FCM removal is just the environmental codegen output generated as a side-effect of running codegen for the member fields, that's a different (and more worrying) concern — it would mean incorrect schema is being committed.

---

### Does it meet the bar?

**Blocking:**

1. **Inline import in test** (`service/member/tests.py`, `_make_member` helper): `from member.models import Member` is inside the method body. `CLAUDE.md` is explicit: "Always place imports at the top of the file — Do not use inline imports inside methods."

2. **Class docstring** (`service/member/tests.py`, `TestReplaceableProperty`): `"""Truth table for Member.replaceable property."""` violates "Do not add docstrings or comments. This applies to test code too."

3. **Tests test an internal property, not an API response.** The parallel test class already in this file — `TestCivilDisorderSerialization` — is the canonical pattern: create a member via the DB, hit `GET /game/:id/`, assert `response.data["members"][0]["civil_disorder"]` is `True`. The new tests instead create *unsaved* `Member(...)` objects and call `.replaceable` directly. This means:
   - The serialization path for `seeking_replacement` and `replaceable` is never exercised.
   - No test proves these fields actually appear in the API response.
   - The philosophy says "Tests assert on observable outcomes (HTTP responses…) not internal state."

   The fix is straightforward: replace `TestReplaceableProperty` with tests that follow `TestCivilDisorderSerialization`'s shape — create a saved member via `game.members.create(seeking_replacement=True)`, call `GET /game/:id/`, assert the response includes `"replaceable": True`. That simultaneously covers the truth-table intent and tests the serialization boundary.

**Needs discussion:**

4. **FCM removal completeness.** If this is a real, permanent removal of FCM device management from the app, the Django-side code should be explicitly removed rather than relying on the absence of `FIREBASE_PROJECT_ID`. If it's not a real removal and is instead just a codegen artifact from running in a cloud session without Firebase, the schema/client changes should be reverted to avoid committing a schema that doesn't match the Firebase-enabled production environment.

---

### Verdict: **Needs changes**

- **Inline import** and **docstring** in `TestReplaceableProperty` are clear rule violations — fix before merging.
- **Test strategy**: replace the unsaved-object unit tests with API-level tests in the style of `TestCivilDisorderSerialization`; the truth-table coverage can be expressed as a small set of API tests covering the key flag combinations.
- **FCM intent**: clarify whether the schema/client changes are a real permanent removal (requiring Django-side code changes too) or an environmental codegen artifact (which should be reverted to keep the schema consistent with the Firebase-enabled production environment).

---

johnpooch:
## Code Review

### What the PR does

The PR adds two new fields to the `Member` model — `seeking_replacement` (BooleanField) and `replaced_by` (FK to self) — plus a `replaceable` computed property that returns `True` when a member is eligible to be replaced by another player. Both fields are exposed through `MemberSerializer`, the OpenAPI schema is updated, and the generated TypeScript `Member` interface gains `seekingReplacement` and `replaceable`. Five API-level tests cover the `replaceable` serialization logic across the key combinations.

---

### Does it make sense?

The member replacement logic is internally consistent: `replaceable` is `True` when the member has `civil_disorder` or `seeking_replacement`, and is `False` when `eliminated`, `kicked`, or already `replaced_by` someone. The `replaced_by_id is not None` check (using the ID field directly) avoids a superfluous DB join — correct.

**The diff does not match the PR description.** The title and description claim the PR also removes FCM device endpoints (`/devices/` GET/POST/PUT), the `/api/test-sentry/` endpoint, the `FCMDevice` schema, and `TypeEnum` from generated types. None of that is in the diff:

- `FCMDevice` and `TypeEnum` are still present in `packages/web/src/api/generated/endpoints.ts` (lines 201–227)
- `FCMDevice` references are still present in `service/openapi-schema.yaml`
- `fcm_django` and `/devices/` are still referenced in `service/project/settings.py` and `service/project/urls.py`
- `test-sentry` is still wired in `service/project/urls.py` and `service/project/views.py`

The PR description should be updated to describe only the member replacement tracking work.

---

### Does it meet the bar?

**Blocking:**

**A. PR description claims removals that aren't in the diff**
The description and checklist both reference FCM device removal that does not exist in this diff. The description needs to be corrected to reflect what the PR actually does.

**B. `replaceable` property placed before `class Meta` in `service/member/models.py`**
The existing `Member` model follows a clear convention: fields → `class Meta` → properties (`name`, `picture`, `__str__`). The new `replaceable` property is inserted *before* `class Meta`, breaking this ordering. It should be moved to after `class Meta` alongside the other properties.

**Non-blocking observations:**

- **`replaced_by` FK is not serialized.** The FK tracks which member replaced whom but is never exposed via the API. If a future screen needs to identify the replacement player, this field will need to be added to the serializer at that point. Fine for now if the use case only needs the boolean `replaceable`.
- **Test approach is consistent with existing patterns.** `TestReplaceableSerialization` mirrors `TestCivilDisorderSerialization` exactly — same `Game.objects.create(...)` pattern, same session-scoped fixtures, same structure. Correct.
- **`replaced_by_id is not None`** is the right way to check a nullable FK without triggering a DB join.

---

### Verdict: Needs changes

- Update the PR description to remove the FCM/test-sentry claims that aren't in the diff.
- Move the `replaceable` property to after `class Meta` in `models.py`, consistent with the existing property ordering convention.

---

johnpooch:
## PR Review

### What the PR does

Adds two new fields to `Member` — a `seeking_replacement` boolean and a `replaced_by` self-referential FK — plus a `replaceable` computed property (`True` when a member has civil disorder or is seeking replacement, and is not yet eliminated, kicked, or already replaced). The new fields are wired through `MemberSerializer`, the OpenAPI schema is updated, the TypeScript client is regenerated, and mock fixtures are updated. Six API-level tests cover key `replaceable` combinations. `CLAUDE.md` gains a note recommending `npx tsc -b --noEmit` after codegen.

---

### Does it make sense?

**Member replacement tracking:** The model design is coherent. `seeking_replacement` and `replaced_by` are clean additions. The `replaceable` property uses `self.replaced_by_id is not None` (avoiding a DB round-trip for the FK check), which is the correct pattern. The migration is auto-generated and safe — both new fields have database-level defaults. The serializer exposes `seeking_replacement` as a plain `BooleanField` (readable directly from the model field) and `replaceable` the same way (readable via the property), matching the existing `civil_disorder` treatment exactly.

**PR title vs. diff:** The PR title says "Add member replacement tracking and **remove FCM device endpoints**". No FCM removal is present in the diff — `endpoints.ts`, `openapi-schema.yaml`, `settings.py`, and `urls.py` all retain FCM content. This is a title/description error left over from planned-but-not-included scope. It should be corrected before merging to avoid confusion in git history.

**Scope:** Excluding the FCM title claim, the PR is single-purpose and cohesive. The `CLAUDE.md` update documents a workflow lesson earned during this PR's development, which is appropriate per the convention of keeping that file current.

---

### Does it meet the bar?

**Blocking:**

1. **PR title claims scope not in the diff.** `"remove FCM device endpoints"` appears in the title but nothing in the diff removes FCM devices from the schema, generated client, or Django configuration. The title and body should be updated to describe only what was actually done.

**Non-blocking observations:**

- **`replaceable` property position is correct.** It is inserted after `class Meta` and before the existing `name` property — consistent with the `fields → Meta → properties` convention. A previous auto-review incorrectly flagged this; the ordering is fine.

- **Tests follow the right pattern.** `TestReplaceableSerialization` mirrors `TestCivilDisorderSerialization` exactly: create a saved member via `game.members.create(...)`, hit `GET /game/:id/`, assert on `response.data`. Six cases cover the key truth-table combinations. No unsaved-object unit tests, no docstrings, no inline imports.

- **`replaced_by` is not serialized.** The FK exists on the model but is not exposed through the API — only `replaceable` (the boolean) is. This is intentional and documented as foundation-only in the description.

- **`CLAUDE.md` update is accurate and useful.** The added guidance ("always run `npx tsc -b --noEmit` after codegen") reflects a real pitfall (required fields silently breaking inline object literals) and the documented fix. Well-targeted.

---

### Verdict: **Needs changes**

- **Fix the PR title and description** to remove the "remove FCM device endpoints" claim. The code does only member replacement tracking; the FCM language belongs to a different PR.
- Everything else — model, migration, serializer, codegen output, mocks, tests, CLAUDE.md — meets the bar and can be merged once the description is corrected.

---

PR #703, opened 2026-06-14
[UX A DAY]: Sort started games by next deadline

JorenC:
## Summary

The "Started" tab of My Games now orders games by how soon their next deadline is — the most urgent game first. Games set to resolve manually ("resolve when ready") and sandbox games sort to the bottom, so the games actually waiting on you are easiest to find.

- **Frontend**: the Started tab passes a new `ordering=deadline` query param (the `Staging`/`Finished` tabs keep the default `-created_at`). No codegen needed — `ordering` already exists in the generated client.
- **Backend**: `GameFilter` gains a `deadline` ordering that annotates each game with its active phase's `scheduled_resolution` via a correlated subquery and orders by:
  1. `sandbox` ascending → sandboxes last
  2. `next_deadline` ascending, NULLs last → soonest deadline first; manual-resolution games after timed ones
  3. `-created_at` → deterministic tiebreaker

### Why server-side?

The My Games list is paginated (`page_size = 20`). A client-side sort would only reorder the games already fetched into the current pages — a game on page 2 with a sooner deadline would still render below page-1 games. Ordering has to happen in the DB before pagination slices the result. Ordering by the absolute `scheduled_resolution` ascending is equivalent to "shortest time left first" and avoids per-request `deadline - now` math.

## Test plan

- [x] `service/game/tests/test_game_list_deadline_ordering.py` (new): asserts `soon < later < manual(null) < sandbox` for `ordering=deadline`, and that tabs without the param still order by `-created_at`. Assertions are scoped to each test's own games because a `UserProfile` post-save signal seeds a "Practice Game" sandbox per user.
- [x] Existing list tests (`test_games_list_perf`, `test_anonymous_games`) — green, no regression.
- [x] `npx tsc -b --noEmit` — clean.

## Screenshots

**No screenshot.** The visual effect is produced entirely by backend ordering; under MSW mocks the games handler returns fixtures in registry order and does not replicate the deadline sort, so a mock screenshot would not demonstrate the change. The behaviour is verified by the backend tests above.

---

## Self code review (`/review-pr`)

**Does it make sense?** Yes. Server-side is the correct layer (pagination would break a client sort). The subquery picks the active phase unambiguously (`status=ACTIVE`, `order_by("-ordinal")[:1]` — an active game has exactly one active phase). The three-key ordering matches the requirement and is verified by tests on both the new and default paths. `ordering` already exists in the generated client, so no codegen is needed. Single-purpose; RELEASE_NOTES updated.

Checked the two likely pitfalls: the cross-app `from phase.models import Phase` import is consistent with `game/models.py` and introduces no cycle; `filter_mine`'s `.distinct()` combined with `order_by` on the subquery annotation works in Postgres (annotation + ordering fields are in the SELECT list) and is exercised by the `mine=true` tests.

**Does it meet the bar?** Yes, no blocking issues. The new branch is structurally identical to the adjacent `slots_remaining` branch; imports at top, no comments, tsc clean, no suppressions. Tests are API-level and assert on observable response ordering, correctly scoped around the seeded "Practice Game".

**Non-blocking suggestions:**
- `ordering: undefined` on the non-Started tab configs is deliberate (keeps `s.ordering` type-safe across the `as const` union), not an oversight.
- No `assertNumQueries` test for the new ordering. The subquery is a single scalar correlated subquery (not an N+1) and `test_games_list_perf` still passes, so this is optional — add one only if the list query budget is actively guarded.

**Verdict:** Looks good.

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -1,5 +1,13 @@
 # Diplicity React - Release Notes
 
+## Started Games Sorted by Deadline (June 2026)
+
+### Change: "Started" games are ordered by next deadline
+
+In My Games, your started games are now ordered by how soon their next deadline is, with
+the most urgent game at the top. Games set to resolve manually ("resolve when ready") and
+sandbox games sort to the bottom, so the games actually waiting on you are easiest to find.
+
 ## Game Creator Rename (June 2026)
 
 ### Change: "Game Master" badge is now "Game Creator"
--- a/packages/web/src/screens/Home/MyGames.tsx
+++ b/packages/web/src/screens/Home/MyGames.tsx
@@ -14,12 +14,13 @@ import { useGamesListInfinite } from "@/hooks/useGamesListInfinite";
 import { useInfiniteScroll } from "@/hooks/useInfiniteScroll";
 
 const statuses = [
-  { value: "pending", label: "Staging", statusFilter: "pending" },
-  { value: "active", label: "Started", statusFilter: "active" },
+  { value: "pending", label: "Staging", statusFilter: "pending", ordering: undefined },
+  { value: "active", label: "Started", statusFilter: "active", ordering: "deadline" },
   {
     value: "completed",
     label: "Finished",
     statusFilter: "completed,abandoned",
+    ordering: undefined,
   },
 ] as const;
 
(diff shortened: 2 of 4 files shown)


johnpooch (approved):
This looks good. We need to watch the performance of this endpoint over the next few days to see if this additional query meaningfully slows down the endpoint. I have a feeling that it might, in which case we might consider denormalizing the scheduled resolution time onto the game model itself.

PR #705, opened 2026-06-14
[UX A DAY] Add unread message indicator to game cards

JorenC:
## What this does

Game cards in the games list now show a pill with an envelope icon and the **total number of unread chat messages** for that game, summed across all of its channels. The badge appears next to the game name and disappears once everything is read. This makes it easy to spot which games have new messages without opening each one.

### Why it was needed

The unread count already powered the chat-tab dot and per-channel counters, but it was only computed on the game **retrieve** endpoint (`GameRetrieveSerializer`). Game cards render a `GameList` from the **list** endpoint, and `GameListSerializer` had no unread field — so the count was genuinely absent from the card's data.

## Changes

**Backend**
- `GameQuerySet.with_total_unread_counts(user)` — annotates `total_unread_message_count` via a single correlated subquery over `ChannelMessage` (grouped by game), reusing the existing `last_read_at` logic and excluding the user's own messages. Unauthenticated users get `0`.
- `GameListView` chains the annotation; `GameListSerializer` exposes the field.
- New API + UI tests.

**Frontend**
- `GameCard` renders a `Badge` pill with a `Mail` icon + count next to the title (non-sandbox only).
- Mock fixtures updated for the now-required field.

**Codegen**
- Regenerated `endpoints.ts` + `openapi-schema.yaml` with just the `totalUnreadMessageCount` field.

## Screenshots

**Desktop** — Started tab; "Spring Awakening" shows the unread pill:

my games desktop

**Mobile** (390×844):

my games mobile

## Self-review notes (`/review-pr`)

Ran the review against this diff. Verdict was **Looks good**. Findings, for reviewer context:

1. **List uses an annotation + plain `IntegerField`; retrieve uses a `SerializerMethodField` + per-object `.count()`.** This divergence is intentional: a method field on the list would be N queries across the games list, so the list path annotates instead (single correlated subquery, flat query count). Noting it so the inconsistency doesn't read as an oversight.

2. **Cross-file import style:** `game/models.py` imports the channel models directly (`from channel.models import ...`), matching the other direct cross-app imports there (`member`, `unit`, `supply_center`, `victory`). `game/serializers.py` references the same models via `apps.get_model("channel", ...)`. The direct import is the right choice in `models.py` — module-level `apps.get_model` fails there because the module loads too early (`AppRegistryNotReady`), and `channel/models.py` doesn't import `game`, so there's no cycle. Flagging only because the two files now reference the same models two different ways.

3. **Codegen handling (transparency):** the generated files were hand-applied with *only* the `totalUnreadMessageCount` delta rather than committing raw codegen output. Running codegen in an environment without `FIREBASE_PROJECT_ID` also strips `/devices/` + `FCMDevice` (documented in `CLAUDE.md` as environmental); committing the raw output would have wrongly dropped `/devices/`. The applied lines are byte-faithful to what production-config codegen emits.

## Verification

- 242 backend game/channel tests pass (incl. 4 new list tests for present / zero-for-anonymous / excludes-own-messages / resets-after-mark-read).
- 20 GameCard tests pass; `tsc -b --noEmit` and ESLint clean.
- Perf/anonymous game tests unaffected — the annotation is a single correlated subquery, so the games-list SQL count stays flat.

### DIFF
--- a/packages/web/src/components/GameCard.tsx
+++ b/packages/web/src/components/GameCard.tsx
@@ -112,7 +112,7 @@ const GameCard: React.FC<GameCardProps> = ({ game, variant, map }) => {
                   {!game.sandbox && (
                     <NationBadge nations={variant.nations} nation={playerNation} />
                   )}
-                  {!game.sandbox && game.totalUnreadMessageCount > 0 && (
+                  {game.totalUnreadMessageCount > 0 && (
                     <Badge variant="default" className="gap-1">
                       <Mail className="size-3" />
                       <span className="relative top-px">
--- a/packages/web/src/components/GameCard.tsx
+++ b/packages/web/src/components/GameCard.tsx
@@ -13,7 +13,7 @@ import {
 } from "@/components/ui/card";
 import { GameDropdownMenu } from "./GameDropdownMenu";
 import { NationBadge } from "./NationBadge";
-import { UserPlus, Lock, MessageSquareOff } from "lucide-react";
+import { UserPlus, Lock, MessageSquareOff, Mail } from "lucide-react";
 import { Badge } from "@/components/ui/badge";
 import { Avatar, AvatarFallback, AvatarImage } from "@/components/ui/avatar";
 import { RemainingTimeDisplay } from "./RemainingTimeDisplay";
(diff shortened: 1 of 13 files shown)


packages/web/src/components/GameCard.tsx
```diff
@@ -107,6 +107,14 @@ const GameCard: React.FC<GameCardProps> = ({ game, variant, map }) => {
                   {game.sandbox && (
                     <Badge variant="secondary">Sandbox</Badge>
                   )}
+                  {!game.sandbox && game.totalUnreadMessageCount > 0 && (
```
johnpooch: Not important really but this sandbox check is not necessary as sandbox games cannot have messages

johnpooch (commented):
I think this pr might also slow down the my games request significantly. 

I recommend we wait a day until we are sure that the ordering pr hasn't had too much of a negative impact before merging this one. That way we can distinguish between the two prs when we look at latency

JorenC:
@johnpooch this is now mergable

PR #706, opened 2026-06-15
Sort chat channels by most recent activity

JorenC:
## What this does

Adds a deterministic order to the chat channel list. A new `ChannelQuerySet.order_for_list()` annotates each channel with `last_activity = Max(messages__created_at)` and orders by:

1. `private` ascending — the public **Public Press** channel (`private=False`) is always pinned to the top.
2. `last_activity` descending with `nulls_last=True` — most recently active channels next, channels with no messages last.
3. `-created_at` as a stable tiebreaker.

`ChannelListView` chains the new method onto its existing queryset. The sort lives entirely in the backend; the frontend keeps rendering the API order verbatim, so no frontend changes are needed.

## Why it makes sense

- Public Press sorts first regardless of its own activity because `private` ascending dominates the ordering, and there is exactly one public channel per game (`game/serializers.py`).
- `Max(messages__created_at)` captures the latest activity whether sent or received (no sender filter).
- The new `Max` annotation coexists safely with the existing filtered `Count(distinct=True)` from `with_unread_counts` — both share a single join/GROUP BY; `distinct=True` protects the count and `Max` is unaffected by row multiplication.

## Tests

Added `TestChannelListOrdering` with three API-level tests asserting on the response body:
- Public channel stays first even when a private channel has newer activity.
- Private channels are ordered by most recent message.
- Channels without messages sort last.

Full channel suite passes (49 tests).

## No codegen / no screenshots

- **No codegen**: the API response *shape* is unchanged — only the order of the array elements differs — so no OpenAPI/client regeneration is required.
- **No screenshots**: there are no visual changes. The reordering is a pure data-ordering change in the backend; the channel list UI renders identical rows in a different order, and there is no way to convey "these are now sorted" in a static screenshot without a contrived before/after of specific fixture data. Mock mode (MSW) serves channels in fixture order and does not exercise the backend sort, so screenshots would not reflect this change anyway.

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -1,5 +1,13 @@
 # Diplicity React - Release Notes
 
+## Chat Channels Sorted by Activity (June 2026)
+
+### Change: Chat channels are ordered by most recent activity
+
+In a game's chat, your channels are now ordered by how recently each one had activity, with
+the most recently active channel near the top. The public "Public Press" channel is always
+pinned to the top, and channels with no messages yet sort to the bottom.
+
 ## Nation Badge (June 2026)
 
 ### Feature: Your nation is shown at a glance
--- a/service/channel/models.py
+++ b/service/channel/models.py
@@ -1,5 +1,5 @@
 from django.db import models
-from django.db.models import Q, Count, Subquery, OuterRef, IntegerField, Value
+from django.db.models import Q, Count, Subquery, OuterRef, IntegerField, Value, Max, F
 from django.contrib.auth import get_user_model
 from django.utils import timezone
 from common.models import BaseModel
@@ -22,6 +22,13 @@ class ChannelQuerySet(models.QuerySet):
     def with_related_data(self):
         return self.prefetch_related(
             "messages", "messages__sender", "messages__sender__user", "members", "members__user"
+        )
+
+    def order_for_list(self):
+        return self.annotate(last_activity=Max("messages__created_at")).order_by(
+            "private",
+            F("last_activity").desc(nulls_last=True),
+            "-created_at",
         )
 
     def with_unread_counts(self, user):
@@ -67,6 +74,9 @@ class ChannelManager(models.Manager):
 
     def with_unread_counts(self, user):
         return self.get_queryset().with_unread_counts(user)
+
+    def order_for_list(self):
+        return self.get_queryset().order_for_list()
 
 
 class Channel(BaseModel):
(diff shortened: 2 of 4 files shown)


JorenC:
## Self-review (`/review-pr`)

### What the PR does
Gives the chat channel list a deterministic order entirely in the backend. A new `ChannelQuerySet.order_for_list()` annotates each channel with `last_activity = Max(messages__created_at)` and orders by `private` ascending (pinning the public "Public Press" channel on top), then last activity descending with empty channels last, then `-created_at` as a tiebreaker. `ChannelListView` chains it onto the existing queryset. Frontend is untouched.

### Does it make sense?
Yes, and it's complete:
- **Public-on-top** is guaranteed: `private` ascending (`False`=0 before `True`=1) dominates, and there is exactly one public channel per game (`game/serializers.py`). Covered by `test_public_channel_first_even_with_older_activity`.
- **Activity ordering** uses `Max(messages__created_at)` (no sender filter, so sent or received both count) with `nulls_last=True` to push message-less channels down. Covered by the other two tests.
- **Aggregate safety**: the new `Max` coexists with the existing filtered `Count(distinct=True)` in `with_unread_counts` over the same `messages` join — single GROUP BY, `distinct` protects the count, `Max` is unaffected by row multiplication. Full channel suite (49 tests) passes.
- **Completeness**: `RELEASE_NOTES.md` entry added (matching the "Started Games Sorted by Deadline" precedent). No API shape change, so no codegen needed; no frontend references to update.
- **Single purpose**: diff touches only `models.py`, `views.py`, the test file, and `RELEASE_NOTES.md`. No scope creep.

### Does it meet the bar?
Yes, against `CLAUDE.md` and the code philosophy's five tenets:
1. **Match existing patterns** — `order_for_list()` follows the QuerySet→Manager delegation pattern like `with_unread_counts`/`with_related_data`; the view stays thin and just chains queryset methods.
2. **Simplicity** — one queryset method, no abstraction or configurability beyond the requirement.
3. **Observable over internal** — all three tests assert on the HTTP response body (channel-name order), not internal query state.
4. **Evidence over assertion** — each ordering rule has a dedicated test; full suite green.
5. **Fix, don't suppress** — no suppressions introduced.

Project-specific: imports at top of file; no docstrings/comments; query optimization lives in the QuerySet not the view; tests are API-level using session-scoped fixtures; inline scenario setup matches the existing style of the test file. No schema change, so no migration required.

**Blocking:** none.

**Suggestions (non-blocking):**
- No `assertNumQueries` test for the list endpoint — but none existed before and the change adds an annotation to the same query (no extra queries), so consistent with the current bar.
- Mock mode (MSW) serves channels in fixture order and won't exercise the backend sort; fine to leave.

### Verdict
**Looks good** — no blocking findings.

---

johnpooch (approved):

Issue #637, opened 2026-06-12
Unit type not appearing in orders list when named coast


johnpooch:
Hello, sorry I'm reporting another (very small) issue. 😅 The order tab always tells which kind of unit (army/fleet) is doing what. But in the specific case of a fleet positioned on a multi-coasts province, it doesn't tell it. (and also doesn't tell on which coast it is positioned)

See image, there is a fleet in Spain North Coast.

JorenC:
This is a similar issue to the support/movement not going to the right coast. If supporting a fleet, the system should know where the fleet is and use that as location.

PR #707, opened 2026-06-15
[UX A DAY] Render named coasts in orders and on the map

### LINKED ISSUE #630: Support from Named-coasts comes from main province
Image

JorenC:
## What & why

Closes #630 and #637.

Multi-coast provinces (Spain, St. Petersburg, Bulgaria) are modelled as separate provinces (`spa/nc`, "Spain (NC)") with a `parent` FK to the main province. Units sit on the named coast, but orders store `source`/`target`/`aux` as the **parent** province. As a result the app showed the parent everywhere except for moves and builds:

- **#637** — a fleet on a named coast showed no unit type and the parent name in the orders list.
- **#630** — support orders from / into a named coast showed (and drew arrows to) the parent province.

This PR makes orders render the correct named coast **wherever it is known**, in the orders list and on the map. It is **display-only** — the adjudicator already validates coasts, so no game logic changes.

## Changes

**Backend (`service/order/`)**
- `models.py` — `aux_coast` (the coast the aux unit currently occupies, mirrors the existing `source_coast`), `_resolved_target_coast`, and a read-only `target_coast` property. Support `summary`/`title` now use the coast-aware display names.
- `utils.py` — `build_move_coast_lookup()` builds `{(source, target): named_coast}` from the move orders already in the response.
- `serializers.py` — new `target_coast` field; `to_representation` attaches the lookup so the model can resolve coasts.
- `views.py` — `OrderListView` builds the lookup from the already-fetched visible orders (no extra queries). Not added to create, to keep its query count unchanged.

**Frontend (`packages/web/`)**
- `OrdersScreen.tsx` — match a unit to its orderable/source province by parent too, and show the unit's own coast name.
- `toRenderState.ts` — support/convoy arrows originate from `sourceCoast`; support arrows point at the resolved `targetCoast`.
- Regenerated client adds `targetCoast` (the `/devices/`+`FCMDevice` schema entries are untouched).

## "Known" semantics & edge cases

- The list returns only orders **visible to you**, so an opponent's move isn't in the lookup → support falls back to the parent. In a completed phase all orders are visible → the coast is shown. This is the desired behaviour and falls out for free.
- Ambiguous / no coast specified / single-coast target → parent.
- Support-to-hold shows the held unit's current coast.

## Tests

- Backend: 4 new tests (coast known / unknown / support-from-coast / support-hold); existing N+1 and query-count guards unchanged. `order/tests.py` → 169 passed.
- Frontend: new `OrdersScreen` list test + updated/added `toRenderState` map tests; `tsc` + eslint clean; map artifact snapshot unchanged.

## Screenshots

Captured against a new `active-named-coast` mock fixture (`Iberian Gambit`).

**Orders list (mobile)** — unit type + coast on the source, coast on the support target:

orders mobile

**Orders list + map (desktop)**:

desktop

**Map (mobile)** — support arrow from St. Petersburg (NC):

map mobile


JorenC:
## PR Review: Render named coasts in orders and on the map

🤖 Automated review via `/review-pr` (code-philosophy + CLAUDE.md rubric). Posting findings only — no approve/merge.

### What the PR does

Multi-coast provinces are modelled as separate child provinces (`spa/nc`) with a `parent` FK, but orders store `source`/`target`/`aux` as the parent. This PR makes the orders list and map render the *specific* coast wherever it can be determined, falling back to the parent when it can't. It's display-only — adjudication is untouched. Backend adds `aux_coast`/`target_coast` model properties plus a `build_move_coast_lookup()` helper that derives the `(source, target) → named_coast` mapping from the move orders already present in the list response; the `OrderListView.list()` override threads that lookup through serializer context. Frontend matches units to their orderable province by parent and points support arrows at the resolved coast.

### Does it make sense?

Yes. The approach is sound and the "known semantics" are the elegant part: the lookup is built only from orders visible to the requesting user, so an opponent's hidden move naturally yields a parent fallback, while a completed phase (all orders visible) shows the real coast. The lookup keying is correct — `build_move_coast_lookup` keys on `(order.source.province_id, order.target.province_id)` and `_resolved_target_coast` reads it with `(self.aux.province_id, self.target.province_id)`, correctly matching a support's aux→the supported unit's move source. `aux_coast` is a faithful mirror of the existing `source_coast` property (`models.py:181` vs `:189`), and both rely on the `phase.units` prefetch already in `with_related_data()` (`models.py:43-48`), so no N+1 is introduced. Edge cases (support-to-hold uses the held unit's current coast; ambiguous/unknown → parent) are each covered by a test.

### Does it meet the bar?

Strong adherence to project conventions. No blocking issues.

**Backend**
- `OrderListView.list()` override (`views.py:20`) bypasses pagination, but `OrderListView` has no `pagination_class` (pagination is configured only on a game view), so behaviour is unchanged — and `variant/views.py:54` already establishes this exact `list()`-override pattern. ✅
- New properties added in the correct model body position; no docstrings/comments in backend code per the style guide; imports at top. ✅
- Serializer stays on the `Serializer` base with an explicit `target_coast` field mirroring `source_coast` (`allow_null=True` → generates `Province | null`). ✅
- Codegen output is a clean one-line addition (`endpoints.ts` + `openapi-schema.yaml`); no `/devices` churn, no hand-edits. ✅

**Frontend**
- `findUnitForProvince` (`OrdersScreen.tsx:82`) preserves the existing dislodged-first preference while adding parent matching — behaviour-preserving refactor of two duplicated call sites. ✅
- `buildOptimisticOrder.ts` updated for the new required field. ✅
- The `useXxxSuspense` / Suspense-wrapper conventions are untouched; no new lint suppressions, no `any`. ✅

**Tests & docs**
- 4 backend tests covering known/unknown/support-from-coast/support-hold, plus a new frontend list test and reworked `toRenderState` map tests. Assertions are on observable output (response `summary`/`target_coast`, rendered text). ✅
- `RELEASE_NOTES.md` entry present for this user-facing change. ✅
- New `active-named-coast` mock fixture is in direct support of the screenshots/tests, not scope creep. ✅

**Non-blocking observations**
- The lookup is intentionally not wired into `OrderCreateView` ("to keep its query count unchanged"). Consequence: immediately after creating a support order into a named coast, the returned/optimistic order shows the parent until the list refetches. Documented and acceptable for a display-only change — flagging only so it's a conscious trade-off.
- `toRenderState` now uses `sourceCoast` for Convoy sources (previously ignored). Harmless in practice — convoying fleets are at sea and never carry a named coast, so this resolves to the parent — and the test was updated to reflect it.

### Verdict

**Looks good** — coherent display-only fix for #630/#637 that matches existing patterns (notably the `source_coast` mirror and the `list()` override), introduces no N+1, ships with observable-outcome tests on both ends, and updates release notes and codegen cleanly. The create-view and convoy-source trade-offs are minor, intentional, and documented.

PR #745, opened 2026-06-16
Player reliability tier gating on games (#353, #355, #356)

johnpooch:
Implements Phase 2 + the two frontend tickets of the reliability-tiers epic (#348) in one PR.

Closes #353. Addresses #355 and #356 (left open — follow-ups below).

## What this adds

A new `min_reliability` field on `Game` (`open` / `reliable_and_new` / `reliable_only`, default `open`) that controls who can join, based on each player's reliability tier. The tier is the existing on-the-fly derivation in `get_player_stats` (`reliable` / `new` / `None`); no tier-storage changes were needed.

**Eligibility:** `open` → all tiers · `reliable_and_new` → reliable + new (excludes unreliable) · `reliable_only` → reliable only.

## Backend (#353)
- `MinReliability` constant + `min_reliability` field on `Game` (migration `0019`, backfills existing rows to `open`).
- `tier_allows_min_reliability(tier, min_reliability)` helper in `user_profile/utils.py`, reusing `get_player_stats`.
- `MeetsReliabilityRequirement` permission appended to `MemberCreateView` (fast-paths `open` before computing stats; returns 403 with a clear message otherwise).
- **Creator enforcement:** `GameCreateSerializer.validate` rejects creating a game whose `min_reliability` the creator's own tier doesn't meet — but only when the creator auto-joins as a player (skipped in Game Master mode, where the creator isn't a player).
- `min_reliability` exposed read-only on the list + retrieve serializers.
- `eligible_only` boolean filter on the game list (computes the user's tier once per request — no N+1).

## Frontend
- **#355:** "Player Reliability" dropdown on the Create Game Advanced step with per-option help text, defaulting to Open.
- **#356:** Find Games only ever shows games the player is eligible to join (`eligible_only: true`). The earlier "show games I can't join" override was removed per review — players find a friend's game via its invite link rather than browsing ineligible games.
- Per-game join-button disabled state is intentionally **out of scope** — deferred to the reliability-badge issue #354. An ineligible join attempt returns 403.

API client regenerated via orval; the committed OpenAPI schema diff is purely additive (the `/devices`/`FCMDevice` removal seen with full codegen in this environment is environmental and was not committed).

## Follow-ups from review (#355/#356 left open)
- #756 — redesign the Player Reliability control (slider / segmented) instead of a dropdown.
- #757 — enforce the reliability requirement on the game-info screen reached via a direct/invite link.
[...]

PR #750, opened 2026-06-16
Add UX design principles to CLAUDE.md

johnpooch:
## Summary

- Adds a new `### UX Principles` subsection under `## UI Guidelines` in CLAUDE.md
- Anchors UX decisions to Material Design 3 as the primary reference system
- Documents a decision table for selection controls (switch, checkbox, radio buttons, segmented button, dropdown, chips, slider) with the MD3 threshold of 5 or fewer options for inline controls
- Documents progressive disclosure rules: what to surface inline vs. hide behind an interaction, including the "right split" rule

These principles were derived from JorenC's feedback on PR #745 and grounded in the official MD3 component guidelines (radio buttons, segmented buttons, menus, chips, switch, sliders).

## No visual changes

This is a documentation-only change to CLAUDE.md.
EOF
)

---
_Generated by [Claude Code](https://claude.ai/code/session_01S3ykaJcm2wCQ1vvqLQZThS)_

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -459,6 +459,10 @@ Follow the MD3 decision tree when choosing how to present a set of options:
 
 **Never use a dropdown when 2–5 fixed options are available and space allows** — hiding choices behind an extra interaction increases cognitive load without saving meaningful space. If the option count is dynamic and could exceed 5, a dropdown is appropriate from the start.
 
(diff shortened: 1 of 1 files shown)


JorenC:
@johnpooch I'm fine to add to this - TBH I don't think we should use Material as baseline - it has some good perks but is also a bit outdated. I can take a look for something else - semantic comes to mind, but there are literally dozens.

More important is that the Claude.md is already passed the Sonnet context window size - so I'm worried that adding more and more here will bloat the guidelines. 

Would it not make sense to add this as an external reference that the claude.md can reference (like 'if there is a front end to be made, reference front-end-design.md or something?

JorenC:
Shouldn't we just close this one and not merge?

johnpooch:
@JorenC yes, based on our discussion earlier, I think we should close this and instead create a design skill which encodes the UX/UI design principles of the project. It will be an evolving skill that is updated by a scheduled job which scans recent PRs for UI feedback.

Issue #479, opened 2026-05-22
PRD: Naming of Variants


JorenC:
When a user uploads a variant, the names are platform-wide. 
However, it's possible that multiple users have the same name for a variant. 

If the variant is not published, the name should only be checked against the own creator. 
E.g. 'joren' 'youngstown' should become 'joren-youngstown'. 
E.g. 'john' 'youngstown' should become 'john-youngstown'. 

That way, users can name their maps what they want.
Only if we publish the game to the official list, should it become singular.

This also means people can upload their 'own' versions of official maps - e.g. "joren" "classical". As long as it is in my playground, happy.

----

Also: uploaded/draft maps should not be propagated to the 'official' list (e.g. game filters).
Only official, published maps should be.

johnpooch:
What is the current state? Can't a user create a "Classical" variant right now? Doesn't seem like there's a problem here.

JorenC:
Right now a user can't make a classical variant because it already exists. The id in the dVAR needs to be unique, so conflicts arise cross-creators.

Adding the suggested user prefix (or better yet, user id (number)) will prevent this issue, and then users can only not conflict with themselves, not between each other.

JorenC:
Will be closed by https://github.com/johnpooch/diplicity-react/pull/764

PR #764, opened 2026-06-17
[Community Variants] User-specific Variant IDs (prefix)

JorenC:
## Summary

- When a user uploads a variant with slug e.g. `europe-1939`, the server now stores the primary key as `{user_id}-europe-1939` (e.g. `25-europe-1939`). System variants (no owner) keep their bare ids unchanged.
- Two different users can now upload variants with the same slug without conflict.
- A `slug` property on `Variant` strips the owner prefix, and `variant_to_canonical_dict` emits the slug so download → re-upload round-trips cleanly.
- `validate_safe_replacement` and the update serializer both compare against `instance.slug` for consistency.
- Migration `0022` widens `Variant.id` `max_length` from 36 to 64 to accommodate the prefix.
- No data migration needed: existing owned drafts keep their current ids; the `slug` property handles them by returning the full id when no prefix is found.

## Test plan

- [ ] All 142 variant tests pass (`docker compose run --rm service python3 -m pytest variant/ -v`)
- [ ] Upload same slug as two different users → both succeed with distinct prefixed ids
- [ ] Upload same slug as same user twice → second upload rejected with "already exists"
- [ ] Download DVAR for owned variant → `id` field is the bare slug, not the prefixed id
- [ ] Download → re-upload round-trip succeeds


JorenC:
## Code review (self-review per CONTRIBUTING convention)

**Verdict: Looks good** — approach is correct, edge cases hold up under tracing, and all 24 affected backend tests pass locally.

### Does it make sense?
Yes. The change set logically and completely addresses the collision problem. Verified key paths:

- **Create** (`serializers.py`, `utils.py:variant_id_for` + `create_variant_from_dvar`): the same `variant_id_for(owner, slug)` is used for both the duplicate check and the actual `create(...)`, so the validated id and the stored id can't drift. Uniqueness is guaranteed by `(owner_id, slug)` since `owner_id` is an integer (no dashes).
- **Slug round-trip** (`models.py` `slug` property): the strip is delimited (`f"{owner_id}-"`), so `"12-x"` isn't mis-stripped by owner `1`. Legacy owned drafts (bare id, with owner) created before this PR are handled — `startswith(prefix)` fails and the full id is returned, so `instance.slug` still equals their DVAR id and updates don't break. Confirms the "no data migration needed" claim.
- **Update / round-trip**: `update_variant_from_dvar` never reassigns `variant.id`, so the prefixed PK survives an update; the slug-vs-slug comparison lets a downloaded DVAR (bare slug) be re-uploaded against its prefixed PK.
- **Downstream**: `create_variant_from_dvar` is also called in `adjudication/tests.py` with `owner=primary_user`; it uses the returned object (not the literal id), so it's correctly untouched.

Ran `variant/tests/test_variant_crud.py` + `test_variants_list_performance.py` → **24 passed**.

### Does it meet the bar?
Follows project conventions: `variant_id_for` is a thin module-level helper, the `slug` property is in the correct model body position, no comments/docstrings, imports at top, migration is an ORM `AlterField` (Postgres-safe). Tests use the session-user/factory fixtures and assert on observable HTTP responses. Scope is tight.

### Minor follow-ups (non-blocking)
- Published variants keep the numeric owner prefix in their public id (admin publish only flips `status`). This is the explicit documented tradeoff — flagging to confirm it's intended that published community variants surface a numeric prefix in URLs/game records. The download/`slug` path masks it, so it's cosmetic.
- `max_length=64` vs an unbounded slug (the JSON schema has a `pattern` but no `maxLength`). Comfortable headroom over the old 36 and not a regression, just noting the column isn't formally bounded.

(Added a `RELEASE_NOTES.md` entry in 67fd5b28 since this is user-facing for variant authors.)

johnpooch:
I don't think I really understand the use case for this. 

The pr description doesn't help me understand either. 

I have two questions:

- why do we want to allow users to upload a variant with the same slug as an existing variant?

- if we want to prefix the slug with the users id or some other disambiguation, why don't we use the variant creator to handle this? The endpoints for variants is publicly accessible, so if we wanted, we could make the variant creator check that the variant id is unique. But even that is overkill. 

One alternative suggestion if we really want this disambiguation: when the user uploads the dvar on Diplicity, extract the id into an input field and validate it on the fly against existing ids. Then the user can edit the id, e.g. europe-1939-johnpooch or whatever they want. 

How many users are really uploading variants with ids that conflict with existing variants?

I don't see a strong justification for why we want this username prefixing on variants. The proposed behaviour is a bit implicit and weird.

JorenC:
Right now slugs are GLOBAL.
We want to make drafts PERSONAL and HIDDEN (until published).

Use case:
- You make a variant. You call it 'coolmap'. You keep it in draft forever. 
- I add a variant. I call it 'coolmap' too. Now I upload and I get an error - but, there is no 'coolmap' in the maps, there is no indicator that it already existed, so there is no reason to have this error and no guideline of why it is there. 
- Another use case is that I have my own version of classical - basically I make a 'backstabbr' version that is much more stylized. I want to call it 'classical'. However, I'm not allowed - because classical already exists. 

This becomes an unnecessary cesspool of 'trying to guess which names are somewhere already in use'. 
Adding this slug
- Mean that I can name it whatever I want (it's a slug anyway that will never get read).
- It means that we would never need to rename things when we move into official.

I strongly believe this is a QoL improvement that doesn't cost anything, but gives freedom to variant-creator to name it whatever they want. This error is stupid (name already exists, but not showing me what that name is) and should not happen. 

-

Secondly, this could not be in the variant-creator; it does not know the user, and there can be time between the dVAR creation and uploading, in which another user could have 'claimed the name'. 

-

Thirdly, editing the slug mean I need to open the dVAR and open it in text to change the JSON. We want it to be usable by non-coders. The alternatively is that I need to re-run the whole dVAR creator AGAIN just to change the name - and then hope that it's not also already taken. Unnecessary UX issues.

-

Thirdly,  this allows 'giving' variants. Let's say I have a variant in progress, and I want you to be able to play it - I just give you the dVAR/dSVG and  you upload it to your private game, upload and done. That way, an 'unofficial' discord collection can be made. 

-

Finally, this is just good practice and doesn't cost anything - no data migration, no issues with old variants, it's just cleaner. The slug is never exposed to the user - and even if it is, it just adds a number in front of it. 

So - why not, with all the benefits?

JorenC:
Closes #479

PR #767, opened 2026-06-17
Add official flag to variants

JorenC:
## Summary

- Adds `official: BooleanField(default=False)` to the `Variant` model
- Seeds classical, vietnam-war, italy-vs-germany, and hundred as official via data migration
- Exposes `official` as a read-only field in `VariantSerializer` (and therefore the API / generated TS type)
- Adds `official` to the admin list display, filter, and inline toggle; adds bulk mark/unmark actions

## No visual changes

This is groundwork only — the `official` field is now on the API but no UI filtering is wired yet. The create-game filter (default to official-only, checkbox to show all) is the follow-up PR.


johnpooch (approved):

JorenC:
@johnpooch — PR review summary (`/review-pr`):

## Verdict: Looks good ✅

### Does it make sense?
Yes — the changeset matches the stated intent end-to-end:
- **Model → migration**: `0022` adds the field; `0023` seeds the four official IDs with a properly reversible `RunPython` (forward `True`, reverse `False`).
- **API exposure**: serializer field is `read_only=True`, correct since `official` is an admin/editorial concern, not client-settable.
- **Codegen consistent**: `endpoints.ts` and `openapi-schema.yaml` both gained the field (added to `required`), and the two inline `Variant` literals in `mocks/legacy.ts` were updated. `npx tsc -b --noEmit` passes clean, confirming no other inline `Variant` objects were missed.
- **Scope**: single purpose — every file traces directly to "add an official flag".
- **Release notes**: correctly omitted (groundwork, nothing user-visible).

### Does it meet the bar?
Strongly. Matches existing patterns throughout:
- Field placed in the fields block before `@property` methods, per model body-order convention.
- Serializer uses the plain `Serializer` base with an explicit read-only field.
- The two new admin actions mirror the existing `publish_selected` / `archive_selected` actions exactly.
- Test update asserts the observable API response and the persisted value.

### Fixed in this branch
- The `actions` tuple in `variant/admin.py` exceeded the configured Black `line-length = 120` — wrapped to multiline (commit `f610413e`).

PR #769, opened 2026-06-17
[Quick Bug Fix] Map renders twice on landscape mode

JorenC:
## Summary

- `GameIndexRoute` was reading `window.innerWidth < 1024` once at render time to decide whether to show the map or orders in the content column
- The layout CSS switches to two-pane (content + right map panel) at 768px (`md:`)
- On portrait→landscape rotation, the CSS immediately revealed the right map panel while the stale `GameIndexRoute` render kept `MapScreen` in the content column → **two maps rendered simultaneously**
- Fix: replace the one-shot `window.innerWidth` read with `useIsMobile()` (reactive `matchMedia` listener, 768px breakpoint) — the same hook used by `GamePhaseRedirect`, `GameCard`, and `getGameLandingPath`

## Before / After

| Portrait (390×844) | Landscape (900×500) |
|---|---|
| Single map, bottom nav | Orders on left, map on right — no duplication |

**Portrait — unchanged (single map):**
portrait

**Landscape — fixed (orders + map, not map + map):**
landscape

## Test plan

- [x] `GameIndexRoute` unit tests added (`src/Router.test.tsx`): asserts MapScreen on mobile, OrdersScreen on desktop, using the same `@/hooks/use-mobile` mock pattern as `GamePhaseRedirect.test.tsx`
- [ ] Manual: rotate a phone browser from portrait → landscape on `/game/:id/phase/:phaseId` — content column should switch to Orders, not show a second map


JorenC:
## PR Review (automated, via `/review-pr`)

### What the PR does
`GameIndexRoute` picks between `MapScreen` and `OrdersScreen` based on screen width. It previously did a one-shot `window.innerWidth < 1024` read at render time. This PR replaces that with the reactive `useIsMobile()` hook (768px breakpoint) and adds a test asserting the correct screen renders for each mode.

### Does it make sense?
Yes — the root cause is real and the fix addresses it directly:

- The layout (`GameDetailLayout.tsx`) switches entirely on Tailwind's `md:` = **768px**. The right-hand `GameMap` panel is `hidden md:flex` (visible ≥768px).
- The old code used **1024px** for the content column. So in the **768–1024px** band (a phone in landscape), `isMobile` was `true` → content column rendered `MapScreen`, *while* the right panel already showed `GameMap` ≥768px → the map appeared twice.

Aligning to `useIsMobile()` fixes both at once: the threshold now matches the layout (768px), and because the hook subscribes to a `matchMedia` listener, rotation re-renders instead of leaving a stale one-shot value. `grep` confirms no other stray `1024`/`innerWidth` breakpoint logic remains. The new test passes and asserts on observable output (which screen renders).

### Does it meet the bar?
Largely yes.

**Suggestions (non-blocking):**
- The test covers the branch (`mobile → MapScreen` / `desktop → OrdersScreen`) but mocks `useIsMobile` wholesale, so it doesn't guard the reactivity that actually fixes the bug. Acceptable — reactivity is the hook's responsibility — but it won't catch a future regression back to a one-shot read.
- `GameIndexRoute` is exported solely for the test. Reasonable and low-cost.

**Non-issue, for the record:** `useIsMobile` returns `false` on its first render before the effect runs, so a mobile user briefly gets `OrdersScreen` before `MapScreen`. This is pre-existing hook behavior shared by every sibling that uses `useIsMobile`, not introduced here.

A `RELEASE_NOTES.md` entry has since been added for this user-facing fix.

### Verdict: **Looks good**
- Root cause correctly identified and verified against the layout's actual 768px breakpoints; fix removes the breakpoint mismatch *and* the staleness in one change.
- Single-purpose, minimal diff, with a passing test asserting observable behavior.

JorenC:
This is pullable - release notes conflict. Can we make an CI/CD auto-script? Whenever you do a PR, all other PRs are now confliceted because of release_notes conflicting - there has to be an easier way to append release notes, no?

PR #789, opened 2026-06-18
Eliminate map zoom flicker via viewBox-driven rendering (Phase 1)

JorenC:
## Problem
The interactive map flickers during pan/zoom on **every** device (independent of FPS), worst on dense maps like Cold War and Spice Islands. Root cause: react-zoom-pan-pinch CSS-scales the whole SVG, so the browser must re-rasterise an oversized GPU layer each frame and drops tiles under memory pressure.

## Approach
Split each map into two stacked layers:
- **`MapVisual`** — visible layers in a fixed, container-filling SVG whose `viewBox` is driven imperatively from the zoom transform. No CSS transform ⇒ the rasterised surface stays bounded to the viewport at any zoom ⇒ no oversized texture, no re-raster flicker.
- **`MapHitLayer`** — intrinsic-sized, fully transparent hit layer that RZPP CSS-transforms for gestures. Transparent ⇒ re-rasterising is free and never flickers; geometry-based hit-testing still works.

RZPP's `{scale, positionX, positionY}` is converted to a `viewBox` in `onTransformed` via the shared `viewBoxZoom.ts` helper, reusing each surface's existing fit/centre/bounds logic. `translateZ(0)` removed.

## Scope — all three zoom surfaces migrated
- **`InteractiveMapZoomWrapper`** (GameMap) — fullscreen toggle preserved.
- **`TutorialMap`** — focus-on-provinces `setTransform` animation preserved (getBBox now targets the hit layer).
- **`ExpandableMapPreview` `ZoomableMap`** — read-only; non-interactive transparent hit layer purely for RZPP bounds/gestures.

No consumer is left on the CSS-transform path. `InteractiveMap.tsx` is removed and its Storybook stories rewritten to render `MapVisual` directly. Shared layer-split helpers live in `mapLayers.ts`; PR #634 diagnostic logging removed.

## Verification
- `tsc -b --noEmit` clean, eslint clean, map tests pass (`GameMap`, `ExpandableMapPreview`, `mapRenderer`, `dsvgParser`).
- Static render confirmed correct on desktop + mobile (GameMap), the tutorial (focus zoom), and the expandable preview dialog (contained/centred).
- **On-device:** flicker confirmed eliminated on Pixel 8a (Android) via remote localhost on the local network. No flicker observed during drag and pinch on Cold War and Spice Islands.


JorenC:
## `/review-pr` findings — PR #789

### What the PR does
Replaces the CSS-transform zoom/pan approach (which forced the browser to re-rasterise an oversized GPU layer every frame, causing flicker) with a two-layer split: a viewport-filling **`MapVisual`** SVG whose `viewBox` is mutated imperatively to mirror the gesture transform, plus a transparent intrinsic-sized **`MapHitLayer`** that react-zoom-pan-pinch CSS-transforms purely for gesture/hit-testing. All three zoom surfaces (GameMap, TutorialMap, ExpandableMapPreview) are migrated; the old `InteractiveMap` component is removed and the shared layer-split/viewBox helpers extracted.

### Does it make sense?
Yes. I traced the core transform math in `viewBoxZoom.ts` and it is correct. The hit layer renders 1 user-unit = 1px (its `viewBox` aspect matches its intrinsic `width`/`height`, so RZPP's `scale`/`position` apply cleanly), and `transformToViewBox` solves for the visual `viewBox` that maps the same user coordinate to the same screen pixel: `minX = mapViewBox.minX − positionX/scale`, `width = container.width/scale`. Because the resulting `viewBox` aspect ratio always equals the container aspect ratio and the SVG fills the container at 100%, `preserveAspectRatio="xMidYMid meet"` adds no letterboxing or distortion. Province fills, borders, unit graphics **and** the hover highlight all live in the same `viewBox`-driven SVG, so they scale together exactly as before — no per-element visual regression. The `useLayoutEffect` seed-before-paint and the `requestAnimationFrame`-deferred focus zoom in `TutorialMap` (now reading `getBBox` off the hit layer) are sound.

### Does it meet the bar?
Solid. Findings are non-blocking:

- **Description vs diff mismatch (minor, CLAUDE.md rule 10):** the body says *"`InteractiveMap.tsx` is now only used by Storybook stories"*, but the file is in fact **deleted** and `InteractiveMap.stories.tsx` was rewritten to render `MapVisual` directly. The diff is the better state; just update the wording.
- The `InteractiveMapZoomWrapper` prop type was tightened from a `ComponentProps`-derived `Omit` to an explicit `InteractiveMapContentProps` — good, matches the prop-interface convention. The sole call site (`GameMap.tsx:275`) matches.
- Scope is cohesive: migration of all three surfaces, removal of the now-dead `InteractiveMap`, extraction of shared helpers, and removal of the PR #634 diagnostic logging are all in service of the one performance goal — not scope creep.

### Verification (run locally on this branch)
- `tsc -b --noEmit`: clean
- `eslint` on all changed files: clean
- `vitest`: `viewBoxZoom` (10), `dsvgParser` (11), `mapRenderer` (40), `ExpandableMapPreview` (2), `GameMap` (5) — all pass
- New `viewBoxZoom.test.ts` covers the conversion math, string formatting, and the null-svg guard — good observable-outcome coverage.

### Verdict: **Looks good**
- Approach is well-reasoned and the math checks out under tracing.
- All three surfaces migrated consistently; no consumer left on the CSS-transform path.
- Type/lint/test gates all green; new unit tests cover the conversion helper.
- Only nit: fix the description line about `InteractiveMap.tsx` to reflect that it was deleted.

*(Per author: flicker fix already confirmed on-device on Pixel 8a; screenshots intentionally omitted as this is a performance-only change.)*

JorenC:
@johnpooch this is ready for review — flicker fix verified on-device (Pixel 8a). PTAL 🙏

johnpooch:
https://github.com/user-attachments/assets/98c5dfdf-7002-4684-b554-feba241db587

johnpooch:
The video I have linked in the comment above shows the performance of the same variant first on your PR in the staging environment and second in production. 

The performance is clearly degraded on my device. 

I have added a comment to the other PR to describe the steps that I think we should take going forward.

PR #790, opened 2026-06-18
Strip unicode-range from embedded map fonts (WebKit Font Embed fix)

JorenC:
## Problem
On iOS Safari / WKWebView, the small place-name labels on some variants (e.g. **Spice Islands**) render in a serif fallback instead of their intended cursive font. The affected font is **Italianno** — the only one of the map's four embedded faces that carries a `unicode-range` descriptor. Chrome/Android renders it correctly, which is why it only showed on iOS.

## Cause
WebKit mishandles `@font-face` with a `unicode-range` applied to SVG `<text>` and silently falls back. Established by elimination:
- The Italianno woff2 data is valid (correct `wOF2` magic + length).
- In Blink the font applies to every label and removing `unicode-range` changes nothing (measured).
- The Libre Baskerville faces (same data-URI embedding, same injected `<style>`, no `unicode-range`) render fine on the same device.
- ⇒ `unicode-range` is the only differentiator.

This is **not** related to the viewBox refactor — the raw standalone SVG reproduces it identically.

## Fix
The variant-creator subsets each embedded font to exactly the glyphs the map uses and emits a matching `unicode-range`, so for a single inlined face the descriptor is redundant — every used character is covered by construction. Strip it during defs collection (`collectDefs`):
- no-op in Blink (verified by measurement),
- fix in WebKit,
- no tofu risk (no character in use is outside the subset).

## Verification
- `tsc -b` clean, eslint clean, 12 `dsvgParser` tests pass (incl. a new case asserting the range is stripped but the face kept).
- **WebKit confirmation pending on-device** — I couldn't run Playwright WebKit headless locally (missing Linux libs).

### DIFF
--- a/packages/web/src/components/InteractiveMap/dsvgParser.test.ts
+++ b/packages/web/src/components/InteractiveMap/dsvgParser.test.ts
@@ -47,6 +47,17 @@ describe("parseDsvg", () => {
     expect(parsed.defs).toContain("<style");
   });
 
+  test("strips unicode-range from embedded @font-face but keeps the face", () => {
+    const svg = `<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 10 10">
+      <style>@font-face { font-family: 'Italianno'; src: url('data:font/woff2;base64,AAAA') format('woff2'); unicode-range: U+41-57, U+61-7a; }</style>
+      <g id="background"></g>
+    </svg>`;
+    const { defs } = parseDsvg(svg);
+    expect(defs).not.toContain("unicode-range");
+    expect(defs).toContain("font-family: 'Italianno'");
+    expect(defs).toContain("format('woff2')");
+  });
+
   test("returns empty maps when the position layers are absent", () => {
     const withoutPositions = `<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 10 10">
       <g id="background"></g>
--- a/packages/web/src/components/InteractiveMap/dsvgParser.ts
+++ b/packages/web/src/components/InteractiveMap/dsvgParser.ts
@@ -77,6 +77,16 @@ const collectPoints = (layer: Element | null): Map<string, Point> => {
 const stripRedundantNamespace = (markup: string): string =>
   markup.split(' xmlns="http://www.w3.org/2000/svg"').join("");
 
+// The variant-creator subsets each embedded font to exactly the glyphs the map
+// uses and emits a matching `unicode-range`. With a single inlined face that
+// gate is redundant in Blink, but WebKit (iOS Safari / WKWebView) mishandles
+// `unicode-range` on `@font-face` applied to SVG <text> and silently falls back
+// to serif. Stripping it makes the font apply to every character that requests
+// it — a no-op in Blink, a fix in WebKit — with no tofu risk, since every used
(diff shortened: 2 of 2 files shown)


JorenC:
## Review note — recommend merging

This PR should be pulled in to prevent the serif fallback rendering of the affected cursive labels on iOS/WebKit.

Removing `unicode-range` can never cause a regression here: the variant-creator only embeds the glyphs that are actually rendered in the SVG, so every character in use is guaranteed to be inside the Base64-embedded subset *by construction* — there can never be a glyph that requests the face but falls outside the embed. Stripping the (redundant) `unicode-range` descriptor is therefore the safe way to go: a no-op where it already worked, a fix where it didn't.

### Review summary

- **Root-cause fix, not a symptom mask.** The change targets exactly the one descriptor identified by elimination as the differentiator between the broken face (Italianno) and the working ones.
- **Regex is correct.** `/\s*unicode-range\s*:[^;}]*;?/gi` stops at `;`/`}`, handles the last-declaration-before-`}` case, leaves no double-space, and covers multiple faces via the `g` flag. The base64 `woff2` payload only contains `[A-Za-z0-9+/=]`, so the literal `unicode-range` cannot be matched inside the font data.
- **Tightly scoped** — the fix plus a test asserting the range is stripped while the face is kept. Test asserts observable output via the public `parseDsvg` API.

**One loose end:** on-device WebKit verification is still pending (per the PR's own Verification section). The change is provably a no-op in Blink and the only changed variable in the elimination argument, so the downside risk is essentially zero — but confirming on the iPhone before/after would close the loop.

**Verdict: Looks good.**

johnpooch:
image

johnpooch:
The screenshot in the comment above shows the Spice Islands map being rendered on Safari on my iPhone in production. The issue you're describing doesn't appear to be happening on my phone. 

I feel like we've gone off course with this interactive map component.

The component performs terribly. The pinch and zoom is sluggish, and we have weird rendering errors on different devices. 

It feels like we're playing whack-a-mole. We're just throwing Claude at the problem with loose instructions and hoping that things get better. 

We need a more systematic way to approach this.

I have a few suggestions:
1. We need to be able to clearly understand what is happening with these components. For example, we need to understand things like frame rate, latency, et cetera. We need to try and understand what's happening to real users so that we can get a better understanding of what's going wrong. It's not impossible at all that we could gather diagnostic data and device IDs and work out what's going wrong. We need to understand how the complexity of the variant SVG interacts with this component. Instinctively, one would assume that smaller SVGs would perform better, but we don't know that empirically. We need to get much more clarity. 

2. It's not easy to see at a glance what these SVGs are comprised of. I think we need to be able to see:
- the total size
- the number of layers
- the complexity of the paths, etc.
to understand which of those factors, if any, is contributing to the poor performance.

3. The problem that we're trying to solve here has been solved a million times before. This is an interactive map graphic that is pan-able and zoom-able. Yet I feel like we are home-rolling a component or a set of components that is littered with complex `useMemo` and `useEffect` chains. I want to take a step back with this and consider whether there might be better solutions if we take a slightly broader look at it.

To summarize, here are my suggestions:
1. Add diagnostic logs to add diagnostic data to the variants.
2. Enable Claude to validate performance improvements.
3. Rebuild the component from scratch using an established library or approach.

PR #795, opened 2026-06-19
Add ux-patterns Claude Code skill

### LINKED ISSUE #748: Create UX design principles documentation

JorenC:
## Summary

- Adds `.claude/skills/ux-patterns/` — a loadable Claude Code skill that encodes resolved UX design decisions for this codebase
- Covers: form controls, feedback patterns, navigation/layout, progressive disclosure, microcopy, and interaction patterns
- Closes #748

## No visual changes

This is a developer tooling change only — no production code was modified.

### DIFF
--- a/.claude/skills/ux-patterns/SKILL.md
+++ b/.claude/skills/ux-patterns/SKILL.md
@@ -1,0 +1,43 @@
+---
+name: ux-patterns
+description: UX design decisions for diplicity-react — which controls to use, how to handle feedback, progressive disclosure, microcopy, accessibility, and interaction patterns. Load when building or reviewing any UI, making design decisions, or choosing between controls.
+allowed-tools: Task, Read, Glob, Grep, Write, Edit, Bash, TodoWrite
+---
+
+# UX Patterns Skill
+
+This skill encodes the design decisions for the diplicity-react application. These are not suggestions — they are resolved decisions to apply consistently across the codebase.
+
+The stack is **React 19 + Tailwind v4 + Radix UI (shadcn/ui style) + react-hook-form + Zod**. No Chakra, no Semantic UI.
+
+## Quick Reference: Which file to consult?
+
+- Choosing a form control (select vs radio vs slider)? → [./form-controls.md](./form-controls.md)
+- Loading states, errors, toasts, confirmations? → [./feedback-patterns.md](./feedback-patterns.md)
+- Sheets vs dialogs, screen structure, navigation? → [./navigation-layout.md](./navigation-layout.md)
+- When and how to explain things to users? → [./progressive-disclosure.md](./progressive-disclosure.md)
+- Button labels, error copy, help text, tone? → [./microcopy.md](./microcopy.md)
+- Optimistic updates, undo, live data, mobile? → [./interaction-patterns.md](./interaction-patterns.md)
+
+## Non-Negotiable Principles
+
+1. **Tabs are navigation only.** Never use `<Tabs>` to drive form selection or binary choices — use radio buttons instead.
+2. **Dialogs are for interruptions only.** A `<Dialog>` or `<AlertDialog>` appears only for destructive confirmations or unexpected required information. Secondary content, player profiles, and settings go in a `<Sheet>`.
+3. **Dropdowns are a last resort for form fields.** 2–4 short-label options use segmented control or radio group. `<Select>` is for 5+ options or long lists.
+4. **No adaptive UI.** The interface is identical for new and experienced players. It does not hide or alter content based on usage history.
+5. **Rare features are deprioritized, not hidden.** Sandbox mode and Game Master controls are always accessible but visually subordinate.
+
+## Suspense Fallbacks
+
+All screens that fetch data must use a skeleton fallback — not `<Suspense fallback={<div></div>}>`. See [./feedback-patterns.md](./feedback-patterns.md) for the correct pattern.
+
+Apply this to every new screen you build and to any existing screen you touch as part of the current issue. Do not fix screens the current PR is not already touching — boy scout rule.
+
+## Detailed Guidance
+
+- **[./form-controls.md](./form-controls.md)** — Decision tree for every input type: select, radio, segmented control, slider, stepper, combobox, checkbox, switch
+- **[./feedback-patterns.md](./feedback-patterns.md)** — Loading skeletons, toast rules, inline errors, confirmation dialogs, button loading states
+- **[./navigation-layout.md](./navigation-layout.md)** — Screen structure, sheet vs dialog, action placement, stepper navigation, focus management
+- **[./progressive-disclosure.md](./progressive-disclosure.md)** — Two-tier explanation model, first-time UX, conditional fields, complexity escalation, the (i) tooltip pattern
+- **[./microcopy.md](./microcopy.md)** — Tone, button labels, placeholder text, help text, error copy, confirm dialog copy
+- **[./interaction-patterns.md](./interaction-patterns.md)** — Optimistic updates, undo, live updates, inline editing, empty states, pagination
--- a/.claude/skills/ux-patterns/SKILL.md
+++ b/.claude/skills/ux-patterns/SKILL.md
@@ -29,7 +29,7 @@ The stack is **React 19 + Tailwind v4 + Radix UI (shadcn/ui style) + react-hook-
 
 ## Suspense Fallbacks
 
-All screens that fetch data must use a skeleton fallback — not `<Suspense fallback={<div></div>}>`. See [./feedback-patterns.md](./feedback-patterns.md) for the correct pattern.
+Keep the Suspense wrapper — it is the correct pattern. Only the fallback changes: replace the empty-div fallback (`<Suspense fallback={<div></div>}>`) with a skeleton that matches the loaded content's dimensions exactly, so the swap causes no layout shift. See [./feedback-patterns.md](./feedback-patterns.md) for the full pattern.
 
 Apply this to every new screen you build and to any existing screen you touch as part of the current issue. Do not fix screens the current PR is not already touching — boy scout rule.
 
--- a/.claude/skills/ux-patterns/feedback-patterns.md
+++ b/.claude/skills/ux-patterns/feedback-patterns.md
@@ -1,0 +1,127 @@
+# Feedback Patterns
+
+## Loading States
+
+**Always use skeleton screens.** When data is loading, render placeholder shapes that match the layout of the content that will appear. Do not use centered spinners. Do not render a blank `<div>`.
+
+The current pattern `<Suspense fallback={<div></div>}>` is an anti-pattern. Replace with a skeleton that mirrors the screen structure.
+
+---
+
+## Toast Notifications (Sonner)
+
+### When to use a toast
+
+- Transient mutations: join game, leave lobby, send message, copy link
+- Network errors mid-session (with a Retry action)
+- Success confirmation for game orders (see special rule below)
+- Undo opportunity after a reversible destructive action
+
+### When NOT to use a toast
+
+- Form submission errors → use inline error banner instead
+- State changes the UI already makes obvious (toggling a checkbox, inline edit save)
+
+### Toast duration
+
+Calculate based on message length. A person reads ~200 words per minute, ~3–4 words per second. Formula: `max(3000, wordCount * 300)` ms. A 5-word message stays for 3s. A 15-word message stays for ~4.5s. Never hard-code a single duration for all toasts.
+
+### Undo toasts
+
+For reversible destructive actions (e.g. leaving a lobby that you can rejoin), execute immediately and show: `"You left the lobby" [Undo]`. The Undo window is ~5 seconds.
+
+---
+
+## Form Errors
+
+### Field-level errors
+
+Show below the field AND apply a destructive-colored border to the input. Both signals together — the border draws the eye, the message explains.
+
+```tsx
+<FormItem>
+  <FormLabel>Game Name</FormLabel>
+  <FormControl>
+    <Input className={cn(fieldState.error && "border-destructive")} {...field} />
+  </FormControl>
+  <FormMessage /> {/* appears below */}
+</FormItem>
+```
+
+### Submission errors
+
+When a form submission fails (server error), show an inline error banner inside the form — not a toast. Place the banner just above the submit/next button so the user sees it without scrolling up to find which field is highlighted.
+
+```tsx
+{submitError && (
+  <Alert variant="destructive">
+    <AlertDescription>{submitError}</AlertDescription>
+  </Alert>
+)}
+<Button type="submit">Create Game</Button>
+```
+
+---
+
+## Confirmation Dialogs
+
+Use `<AlertDialog>` only for:
+
+1. **Destructive/irreversible actions**: Delete account, concede game, end game
+2. **Unexpected required information**: Something the user must acknowledge before proceeding that could not have been anticipated
+
+Do not use `<AlertDialog>` for: secondary information, player profiles, settings, or anything that could be a `<Sheet>`.
+
+### Copy format
+
+**Title**: Verb-first, names the action exactly. No "Are you sure?"
+
+**Description**: States the specific consequence. No vague "This cannot be undone."
+
+```
+Title:       "Delete account"
+Description: "Your profile, game history, and all associated data will be permanently deleted."
+
+Title:       "Concede game"
+Description: "You will be eliminated. Your units are disbanded and your supply centers become neutral."
+```
+
+---
+
+## Button Loading State
+
+When an action is in progress, replace the button label with a spinner icon. The button remains disabled and its width does not change (use a fixed-width container if needed).
+
+```tsx
+<Button disabled={isPending}>
+  {isPending ? <Loader2 className="size-4 animate-spin" /> : "Create Game"}
+</Button>
+```
+
+Do not show "Creating..." text. Do not show a spinner alongside text.
+
+---
+
+## Urgency Signals
+
+For approaching deadlines, change the color of the time display:
+
+- Normal: default text color
+- ~6 hours remaining: amber/warning color
+- ~2 hours remaining: red/destructive color
+
+Do not add pulsing animations or banner alerts. Push notifications handle proactive alerting — the in-app UI signals urgency through color only.
+
+---
+
+## Network Errors (Mid-Session)
+
+When a request fails while the user is using the app (not initial load), show an error toast with a Retry action:
+
+```
+"Failed to load games"  [Retry]
+```
+
+TanStack Query retries automatically with backoff. Only surface the error to the user after all automatic retries are exhausted.
+
+For initial-load failures (data the screen requires to render at all), use `<QueryErrorBoundary>` which handles the error state inline in the content area, keeping the `<ScreenHeader>` visible.
--- a/.claude/skills/ux-patterns/feedback-patterns.md
+++ b/.claude/skills/ux-patterns/feedback-patterns.md
@@ -2,9 +2,11 @@
 
 ## Loading States
 
-**Always use skeleton screens.** When data is loading, render placeholder shapes that match the layout of the content that will appear. Do not use centered spinners. Do not render a blank `<div>`.
+**Always use skeleton screens.** When data is loading, render placeholder shapes that match the layout of the content that will appear. Do not use centered spinners.
 
-The current pattern `<Suspense fallback={<div></div>}>` is an anti-pattern. Replace with a skeleton that mirrors the screen structure.
+Keep the Suspense wrapper — it is good practice. Only the fallback changes: replace `<Suspense fallback={<div></div>}>` with a skeleton that mirrors the screen structure.
+
+**The skeleton must match the loaded content's dimensions exactly** so the swap causes no vertical shift ("hiccup"). Where content has a variable height (game cards, player cards), give the component a fixed height so the skeleton and the loaded content occupy the same space.
 
 ---
 
@@ -26,9 +28,20 @@ The current pattern `<Suspense fallback={<div></div>}>` is an anti-pattern. Repl
 
 Calculate based on message length. A person reads ~200 words per minute, ~3–4 words per second. Formula: `max(3000, wordCount * 300)` ms. A 5-word message stays for 3s. A 15-word message stays for ~4.5s. Never hard-code a single duration for all toasts.
 
+A toast that carries an action (Undo, Retry) must stay long enough to use that action. Its duration is **whichever is longer** — the formula above or the action's window (see Undo toasts). An action toast must never disappear before its action can be used.
+
 ### Undo toasts
 
-For reversible destructive actions (e.g. leaving a lobby that you can rejoin), execute immediately and show: `"You left the lobby" [Undo]`. The Undo window is ~5 seconds.
+This is the canonical Undo pattern; other files reference it rather than restating it.
+
+For reversible destructive actions (e.g. leaving a lobby you can rejoin):
+
+1. Execute the action immediately — it is a real mutation, the user has left.
+2. Show a toast with an Undo action: `"You left the lobby" [Undo]`.
+3. The undo window is ~5 seconds. The toast stays for whichever is longer: the duration formula or the undo window.
+4. Tapping Undo issues a *separate* reversing request (e.g. rejoin). That request can itself fail — if it does, show an error toast and leave the original action committed.
+
+For irreversible actions, use a confirmation dialog instead (see Confirmation Dialogs below).
 
 ---
 
@@ -36,17 +49,7 @@ For reversible destructive actions (e.g. leaving a lobby that you can rejoin), e
 
 ### Field-level errors
 
-Show below the field AND apply a destructive-colored border to the input. Both signals together — the border draws the eye, the message explains.
-
-```tsx
-<FormItem>
-  <FormLabel>Game Name</FormLabel>
-  <FormControl>
-    <Input className={cn(fieldState.error && "border-destructive")} {...field} />
-  </FormControl>
-  <FormMessage /> {/* appears below */}
-</FormItem>
-```
+The message appears below the field and the input shows a destructive border — both signals together. **This is automatic in the standard `FormField` pattern**: `FormControl` sets `aria-invalid` when the field has an error, `Input` styles its border destructive on `aria-invalid`, and `FormMessage` renders the message. Do not add a manual `border-destructive` class.
 
 ### Submission errors
 
@@ -65,12 +68,7 @@ When a form submission fails (server error), show an inline error banner inside 
 
 ## Confirmation Dialogs
 
-Use `<AlertDialog>` only for:
-
-1. **Destructive/irreversible actions**: Delete account, concede game, end game
-2. **Unexpected required information**: Something the user must acknowledge before proceeding that could not have been anticipated
-
-Do not use `<AlertDialog>` for: secondary information, player profiles, settings, or anything that could be a `<Sheet>`.
+See [./navigation-layout.md](./navigation-layout.md) for when a confirmation belongs in an `<AlertDialog>` versus a `<Sheet>`. When you do use an `<AlertDialog>`, the copy follows this format:
 
 ### Copy format
 
@@ -90,27 +88,26 @@ Description: "You will be eliminated. Your units are disbanded and your supply c
 
 ## Button Loading State
 
-When an action is in progress, replace the button label with a spinner icon. The button remains disabled and its width does not change (use a fixed-width container if needed).
+When an action is in progress, show a spinner before the button's label and disable the button. Keep the label — do not change it to "Creating..." or "Loading...".
 
 ```tsx
 <Button disabled={isPending}>
-  {isPending ? <Loader2 className="size-4 animate-spin" /> : "Create Game"}
+  {isPending && <Loader2 className="size-4 animate-spin" />}
+  Create Game
 </Button>
 ```
-
-Do not show "Creating..." text. Do not show a spinner alongside text.
 
 ---
 
 ## Urgency Signals
 
-For approaching deadlines, change the color of the time display:
+Signal urgency through color escalation only. Do not add pulsing animations or banner alerts — push notifications handle proactive alerting; the in-app UI escalates color as the situation becomes more urgent.
+
+Example — an approaching deadline, by colouring the time display:
 
 - Normal: default text color
 - ~6 hours remaining: amber/warning color
 - ~2 hours remaining: red/destructive color
-
-Do not add pulsing animations or banner alerts. Push notifications handle proactive alerting — the in-app UI signals urgency through color only.
 
 ---
 
--- a/.claude/skills/ux-patterns/form-controls.md
+++ b/.claude/skills/ux-patterns/form-controls.md
@@ -1,0 +1,105 @@
+# Form Controls
+
+## The Decision Tree
+
+### Single-select fields
+
+Start with the number of options and their label length:
+
+| Options | Label length | Control |
+|---------|-------------|---------|
+| 2 | Any | Segmented control (horizontal equal-weight pills) |
+| 3–4 | Short (1–2 words) | Segmented control |
+| 3–4 | Long (3+ words) | Vertical radio group |
+| 3–4 | Each has a description | Radio group with description under each label — OR slider if options form a spectrum (see below) |
+| 5+ | Any | `<Select>` dropdown |
+| 10+ | Predictable/searchable names | Combobox (typeahead) |
+
+**Spectrum options → Slider.** When options represent a continuum from permissive to restrictive (e.g. player reliability: Open → Reliable+New → Reliable Only), use a slider. Left = most open, right = most restrictive. Label the endpoints and current value.
+
+**CRITICAL: `<Tabs>` is never a form control.** It is navigation only. The current codebase incorrectly uses `<Tabs>` for `deadlineMode` and `variantCategory` — those are anti-patterns to fix, not to follow.
+
+---
+
+### Binary toggles (exactly 2 options)
+
+- **On/off semantics** (enabling a feature): `<Switch>`. Example: Private game, Push notifications.
+- **Two named alternatives** (neither is obviously "on"): Radio buttons. Example: Fixed Time vs Duration deadlines, Random vs Ordered nation assignment.
+
+Never use `<Tabs>` for binary form choices.
+
+---
+
+### Boolean fields
+
+Use `<Checkbox>` with `FormLabel` to the right. For settings rows (Account screen), use `<Switch>` with the label to the right of the toggle.
+
+---
+
+### Number inputs
+
+| Range / context | Control |
+|----------------|---------|
+| Small integers, ≤5 options | Stepper (− display +) |
+| Spectrum with visual meaning | Slider |
+| Large, precise, or open-ended | Plain numeric `<Input type="number">` |
+
+---
+
+### Multi-select fields
+
+Use a vertical checkbox list (`<Checkbox>` group). This applies when ≤8 options are available. For longer lists that need search, use a Popover-based combobox with checkboxes inside.
+
+---
+
+### Searchable selects
+
+Use a Combobox (typeahead input + filtered dropdown) when the list has 10+ options with predictable names that benefit from typing. Do not use a Combobox just because the list is long — only when searching is genuinely useful.
+
+The timezone selector currently has enough options for a Combobox, but the options are too similar (all "City (Region)" format) to benefit from search — keep it as a `<Select>`. A community variant selector with hundreds of variants would warrant a Combobox.
+
+---
+
+### Paired fields
+
+When two fields logically belong together (e.g. time + timezone, movement duration + retreat duration), place them in a 2-column grid on the same row:
+
+```tsx
+<div className="grid grid-cols-2 gap-4">
+  <FormField name="fixedDeadlineTime" ... />
+  <FormField name="fixedDeadlineTimezone" ... />
+</div>
+```
+
+This is the only case where form fields share a row. All other fields are full-width stacked.
+
+---
+
+### Option label overflow
+
+Truncate long labels with ellipsis after 1 line. Always add a tooltip that shows the full label on hover (desktop) or long press (mobile). Never wrap labels across 2 lines in a segmented control.
+
+---
+
+## Component Reference
+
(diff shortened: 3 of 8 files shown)


JorenC:
@johnpooch I don't know how you plan to include this - it works standalone. Major risk thing is that I told it we need to use skeleton frames - it will start doing that and boyscout ruling previous screens it touches, but not touching screens that are not related to the current PR. We should have skeleton frames, so I think this is the right approach - rather than do a full refactor.

.claude/skills/ux-patterns/feedback-patterns.md
```diff
@@ -0,0 +1,127 @@
...
+- Undo opportunity after a reversible destructive action
+
+### When NOT to use a toast
+
+- Form submission errors → use inline error banner instead
+- State changes the UI already makes obvious (toggling a checkbox, inline edit save)
+
+### Toast duration
+
+Calculate based on message length. A person reads ~200 words per minute, ~3–4 words per second. Formula: `max(3000, wordCount * 300)` ms. A 5-word message stays for 3s. A 15-word message stays for ~4.5s. Never hard-code a single duration for all toasts.
```
johnpooch, line 29: This feels a little bit over-engineered to me. I feel like simpler guidance would be to simply keep the toast message short and use some reasonable default toast duration unless otherwise specified.

.claude/skills/ux-patterns/SKILL.md
```diff
@@ -0,0 +1,43 @@
...
+
+1. **Tabs are navigation only.** Never use `<Tabs>` to drive form selection or binary choices — use radio buttons instead.
+2. **Dialogs are for interruptions only.** A `<Dialog>` or `<AlertDialog>` appears only for destructive confirmations or unexpected required information. Secondary content, player profiles, and settings go in a `<Sheet>`.
+3. **Dropdowns are a last resort for form fields.** 2–4 short-label options use segmented control or radio group. `<Select>` is for 5+ options or long lists.
+4. **No adaptive UI.** The interface is identical for new and experienced players. It does not hide or alter content based on usage history.
+5. **Rare features are deprioritized, not hidden.** Sandbox mode and Game Master controls are always accessible but visually subordinate.
+
+## Suspense Fallbacks
+
+All screens that fetch data must use a skeleton fallback — not `<Suspense fallback={<div></div>}>`. See [./feedback-patterns.md](./feedback-patterns.md) for the correct pattern.
```
johnpooch: 💡 **suggestion**: This (and feedback-patterns.md) declares the empty-`<div>` Suspense fallback an anti-pattern and mandates skeletons, but CLAUDE.md's Suspense Wrapper Pattern and review checklist explicitly require the fallback be "an empty div (not a spinner or skeleton — keep it minimal)". That leaves two contradictory sources of agent guidance. Reconcile by updating CLAUDE.md in this PR (or aligning the skill with it).

---

.claude/skills/ux-patterns/interaction-patterns.md
```diff
@@ -0,0 +1,173 @@
...
+## Empty States
+
+All empty states include: icon + title + description + primary CTA (action to resolve the empty state).
+
+```tsx
+<Notice
+  icon={Users}
+  title="No games found"
+  message="Try a different filter or create your own game."
+  action={<Button>Create Game</Button>}
```
johnpooch: 🛑 **blocking**: `Notice` takes an `actions` prop (plural), not `action` — `src/components/Notice.tsx` has no `action` prop, so an agent following this writes code that won't typecheck. Same error in microcopy.md's empty-state examples. Use `actions={<Button>Create Game</Button>}`.

---

.claude/skills/ux-patterns/feedback-patterns.md
```diff
@@ -0,0 +1,127 @@
...
+
+### Undo toasts
+
+For reversible destructive actions (e.g. leaving a lobby that you can rejoin), execute immediately and show: `"You left the lobby" [Undo]`. The Undo window is ~5 seconds.
```
johnpooch: If I'm understanding correctly, the undo button will appear as part of a success toast. For example, "You have joined the game. [Undo]" 

If this is the case, then I think that the undo window being around five seconds is contradicting the guidance about toast durations. 

We may want to advise Claude to make the toast duration longer when the toast includes an action, e.g., Undo.

.claude/skills/ux-patterns/feedback-patterns.md
```diff
@@ -0,0 +1,127 @@
...
+
+---
+
+## Form Errors
+
+### Field-level errors
+
+Show below the field AND apply a destructive-colored border to the input. Both signals together — the border draws the eye, the message explains.
```
johnpooch: I don't disagree with this guidance at all, but just pointing out that the shad cn components provide this behavior by default.

.claude/skills/ux-patterns/feedback-patterns.md
```diff
@@ -0,0 +1,127 @@
...
+---
+
+## Confirmation Dialogs
+
+Use `<AlertDialog>` only for:
+
+1. **Destructive/irreversible actions**: Delete account, concede game, end game
+2. **Unexpected required information**: Something the user must acknowledge before proceeding that could not have been anticipated
+
+Do not use `<AlertDialog>` for: secondary information, player profiles, settings, or anything that could be a `<Sheet>`.
```
johnpooch: I 100% agree with this guidance. I generally don't like using dialogues except for in these exact scenarios.

.claude/skills/ux-patterns/feedback-patterns.md
```diff
@@ -0,0 +1,127 @@
...
+
+Title:       "Concede game"
+Description: "You will be eliminated. Your units are disbanded and your supply centers become neutral."
+```
+
+---
+
+## Button Loading State
+
+When an action is in progress, replace the button label with a spinner icon. The button remains disabled and its width does not change (use a fixed-width container if needed).
```
johnpooch: I don't 100% agree with this guidance. I think that adding a loading spinner before the button label seems more natural. See the example on ShadCN docs here.  

https://ui.shadcn.com/docs/components/radix/button#spinner

.claude/skills/ux-patterns/feedback-patterns.md
```diff
@@ -0,0 +1,127 @@
...
+
+```tsx
+<Button disabled={isPending}>
+  {isPending ? <Loader2 className="size-4 animate-spin" /> : "Create Game"}
+</Button>
+```
+
+Do not show "Creating..." text. Do not show a spinner alongside text.
```
johnpooch: I agree with not showing the "creating..." text, but why would we not show the spinner alongside the text? It feels like a very common pattern.

.claude/skills/ux-patterns/feedback-patterns.md
```diff
@@ -0,0 +1,127 @@
...
+
+---
+
+## Urgency Signals
+
+For approaching deadlines, change the color of the time display:
+
+- Normal: default text color
+- ~6 hours remaining: amber/warning color
+- ~2 hours remaining: red/destructive color
```
johnpooch: I feel like the guidance here should be more generic and then use the approaching deadlines as an example.

.claude/skills/ux-patterns/feedback-patterns.md
```diff
@@ -0,0 +1,127 @@
+# Feedback Patterns
+
+## Loading States
+
+**Always use skeleton screens.** When data is loading, render placeholder shapes that match the layout of the content that will appear. Do not use centered spinners. Do not render a blank `<div>`.
```
johnpooch: I am broadly in agreement with the idea of using skeletons. Initially, I didn't use them because I wanted that initial load to be so fast that adding the skeletons would almost feel like jank. But now I feel like they should be added. 

But I have a very small pet peeve with skeletons, and that is when the skeletons don't match the loaded content exactly in height and width. This is especially common when the content that is being represented using the skeleton has a variable height depending on what's inside it. 

The most important example for Diplicity is the game card and maybe the player cards. It would be great if we could enforce a fixed height for these components so that the skeleton to content transition is smooth

.claude/skills/ux-patterns/feedback-patterns.md
```diff
@@ -0,0 +1,127 @@
...
+
+The current pattern `<Suspense fallback={<div></div>}>` is an anti-pattern. Replace with a skeleton that mirrors the screen structure.
```
johnpooch: I think this guidance goes against guidance that appears elsewhere in our documentation. I'm OK with replacing the empty div with the skeleton, but the use of the suspense pattern is not an anti-pattern. It is actually a very good practice, and we should continue to use it.

.claude/skills/ux-patterns/form-controls.md
```diff
@@ -0,0 +1,105 @@
...
+
+---
+
+### Option label overflow
+
+Truncate long labels with ellipsis after 1 line. Always add a tooltip that shows the full label on hover (desktop) or long press (mobile). Never wrap labels across 2 lines in a segmented control.
+
+---
+
+## Component Reference
```
johnpooch: I think we should remove this component reference section. I think it actually duplicates knowledge that Claude can trivially infer from reading the codebase. Claude already knows to use shad-cn components, et cetera. If we change the component that we use, this table will go out of date immediately.

.claude/skills/ux-patterns/form-controls.md
```diff
@@ -0,0 +1,105 @@
...
+
+---
+
+## Anti-patterns in the current codebase (do not follow)
+
+- `mode` field (Standard/Gunboat/Sandbox): currently `<Select>` with 3 options → should be radio group (labels are short-ish but options have meaningfully different descriptions)
+- `nmrExtensionsAllowed` (None/1 per player/2 per player): currently `<Select>` with 3 short options → should be segmented control
+- `minReliability` (Open/Reliable+New/Reliable Only): currently `<Select>` with descriptions → should be slider (spectrum)
+- `deadlineMode` (Fixed Time/Duration): currently `<Tabs>` used as form selector → should be radio buttons
+- Theme selector (Light/Dark/System): currently `<Select>` with 3 options → should be segmented control
```
johnpooch: I would remove this section and instead use these examples as inline examples for each of the rules you've described above. As soon as we fix these anti-patterns, this section will be out of date.

.claude/skills/ux-patterns/interaction-patterns.md
```diff
@@ -0,0 +1,173 @@
...
+### Standard optimistic (messaging, non-critical actions)
+Update the UI immediately. If the server rejects, revert silently and show an error toast. No success toast — the UI state change is the confirmation.
+
+### Orders: optimistic + success confirmation
+Game orders are the exception. Update the UI immediately (optimistic), then show a success toast when the server confirms. This is non-negotiable: a missed order can lose a game. The player needs to know their order was received.
+
+```tsx
+// After mutateAsync resolves:
+toast.success("Orders submitted")
+```
```
johnpooch: This code example can be removed.

.claude/skills/ux-patterns/interaction-patterns.md
```diff
@@ -0,0 +1,173 @@
...
+
+---
+
+## Undo
+
+For reversible destructive actions (actions the user can redo without consequence, e.g. leaving a lobby they can rejoin):
+
+1. Execute the action immediately
+2. Show a toast: `"You left the lobby" [Undo]`
+3. Undo window: ~5 seconds
```
johnpooch: There is a little bit of duplication here with the feedback-patterns.md file. The duplication is probably okay, but I mention it because I think that the timing here contradicts other guidance in that file.

.claude/skills/ux-patterns/interaction-patterns.md
```diff
@@ -0,0 +1,173 @@
...
+4. If the user doesn't undo, the action is committed
```
johnpooch: The meaning of steps one and four is ambiguous to me. Step one, "execute the action immediately," implies that the mutation or API call happens immediately. For example, the user immediately joins the game, but step four implies that actually the action doesn't happen until the undo window has expired. 

I think the most robust approach is to execute the action immediately and have a separate request for undoing the action, but then again, there is always the possibility that the undo request fails.

.claude/skills/ux-patterns/interaction-patterns.md
```diff
@@ -0,0 +1,173 @@
...
+---
+
+## Live Updates
+
+When new data arrives while the user is viewing a screen (e.g. a new game phase starts):
+
+- Show a toast describing what changed
+- Automatically refresh the relevant list/screen
+
+Do not use "New content available — tap to refresh" banners. The toast + auto-refresh handles it.
```
johnpooch: There's one exception to this that is currently in the application. When a user is looking at a previous phase and the current phase resolves, we want to indicate to the user that the current phase has resolved. It may be disorienting to forcibly shift them to the newest phase. That being said, forcibly shifting them to the newest phase when they're on the current phase seems like reasonable behavior. Maybe for simplicity we should just always forcibly shift them into the newest phase when a phase resolves. The slightly disorienting feeling of being shifted while looking at historical phases is probably negligible and probably pretty rare.

.claude/skills/ux-patterns/interaction-patterns.md
```diff
@@ -0,0 +1,173 @@
...
+For irreversible actions, use a confirmation dialog instead (see [./feedback-patterns.md](./feedback-patterns.md)).
+
+---
+
+## Live Updates
+
+When new data arrives while the user is viewing a screen (e.g. a new game phase starts):
+
+- Show a toast describing what changed
+- Automatically refresh the relevant list/screen
```
johnpooch: I agree with this guidance, except I would maybe say something like "automatically reload the relevant screen content." I know it's a pedantic thing, but to me this reads like actually refreshing the page. 

For example, when a user creates an order, what happens behind the scenes is that the list orders endpoint is called again, and the new data is rendered, so we don't refresh the whole screen. We just reload the updated content.

.claude/skills/ux-patterns/interaction-patterns.md
```diff
@@ -0,0 +1,173 @@
...
+
+Do not use "New content available — tap to refresh" banners. The toast + auto-refresh handles it.
+
+---
+
+## Infinite Scroll
+
+For lists that can grow unboundedly (game history, message history), load more content as the user scrolls to the bottom. No "Load more" button. No pagination controls.
```
johnpooch, line 41: Agreed.

.claude/skills/ux-patterns/interaction-patterns.md
```diff
@@ -0,0 +1,173 @@
...
+
+**No-results state** (search/filter returned nothing): "No results" title + "Clear filters" CTA. No icon needed — keep it lightweight.
+
+---
+
+## Multi-Select with Checkboxes
+
+In a vertical checkbox list, when the user has applied selections and the state is "nothing found":
+
+Show the empty state with a "Clear filters" button that resets all selections to their default (unselected) state.
```
johnpooch: This section feels a little bit random. I'm not sure I really follow it.

.claude/skills/ux-patterns/interaction-patterns.md
```diff
@@ -0,0 +1,173 @@
...
+---
+
+## Notification Badges
+
+| Situation | Component |
+|-----------|-----------|
+| "Your turn" on a nav item | Dot indicator (colored dot, no count) |
+| Unread message count | Numeric badge |
+| Phase or game status | Semantic `<Badge>` (colored pill) |
+| Metadata count (3/7 players) | Muted text — no badge |
```
johnpooch: This doesn't really feel like UX guidance and feels more like the spec for a component. I feel like we should extract the underlying principles that drive these decisions rather than describing them like this. 

I.e., why do we use:
- a dot indicator
- a numeric badge
- a semantic badge
- muted text

.claude/skills/ux-patterns/interaction-patterns.md
```diff
@@ -0,0 +1,173 @@
...
+    Expand all
+  </button>
+  <button
+    className="text-muted-foreground hover:text-foreground underline-offset-2 hover:underline"
+    onClick={() => setExpandedItems([])}
+  >
+    Collapse all
+  </button>
+</div>
+```
```
johnpooch: I don't think this code example is required.

.claude/skills/ux-patterns/interaction-patterns.md
```diff
@@ -0,0 +1,173 @@
...
+
+## Game Order Submission Flow
+
+Direct manipulation on the map — tapping a unit shows order options contextually, inline on or near the map. No separate wizard panel. No sheet for order type selection.
+
+This is the existing pattern in the codebase. Do not change it to a side-panel form.
+
+---
+
+## Accordion: Orders Display
```
johnpooch: Again, this kind of feels like the specification for a component rather than generic UX patterns that Claude will understand moving forward.

.claude/skills/ux-patterns/interaction-patterns.md
```diff
@@ -0,0 +1,173 @@
...
+    onClick={() => setExpandedItems([])}
+  >
+    Collapse all
+  </button>
+</div>
+```
+
+---
+
+## First-Time Feature Encounter
```
johnpooch: These instructions feel a little bit vague and half-baked. I understand that this is something that you're planning to work on in the future. For now, I think it will confuse Claude because Claude has no way to know whether a player has encountered a feature or not.

.claude/skills/ux-patterns/interaction-patterns.md
```diff
@@ -0,0 +1,173 @@
...
+
+When a user encounters a feature for the first time (not covered in the new-player introduction):
+
+Show a subtle muted text block below the relevant UI element. It disappears after the user takes their first action. It does not reappear.
+
+This is the only contextual help mechanism for first encounters. Do not use highlighted "New" badges, modal announcements, or coach marks.
+
+---
+
+## App Updates
```
johnpooch: I don't disagree with this guidance, but I'd say it's probably unnecessary.

.claude/skills/ux-patterns/microcopy.md
```diff
@@ -0,0 +1,159 @@
...
+
+## Tone of Voice
+
+**Serious and direct.** This is a strategy game. Players are focused. They want efficiency, not charm.
+
+- No personality, warmth, or wit in UI copy
+- No thematic language ("dispatches", "nations" in button labels)
+- No exclamation marks
+- No first-person from the app ("We couldn't find that game")
+- Short sentences. Active voice.
```
johnpooch, line 12: And perhaps no AI coded language, like em-dashes. 

I also think marketing text is a dead giveaway for AI copy.

To be clear, I know that we generate most of our copy using AI, and that's fine. We stand by that as a decision, but I still feel that players will instinctively recoil at text that is obviously AI-generated.

.claude/skills/ux-patterns/microcopy.md
```diff
@@ -0,0 +1,159 @@
...
+// Bad
+<AlertDialogTitle>Are you sure?</AlertDialogTitle>
+<AlertDialogDescription>This action cannot be undone.</AlertDialogDescription>
+```
+
+**Action buttons**: Primary = the destructive action (verb-only: "Delete", "Leave", "Concede"). Secondary = "Cancel".
+
+---
+
+## Empty States
```
johnpooch, line 145: This is largely a duplicate of what we have in the feedback patterns section.

.claude/skills/ux-patterns/navigation-layout.md
```diff
@@ -0,0 +1,159 @@
...
+
+- `<ScreenHeader title="...">` → renders as `h1`
+- Section headings inside cards → `<h2 className="text-lg font-semibold">`
+- Sub-section headings → `<h3 className="text-sm font-medium">`
+
+Never go deeper than h3. Never use bold text to substitute for a heading level.
+
+---
+
+## Sheet vs Dialog: The Rule
```
johnpooch, line 33: This is kind of duplicating what we have in feedback patterns.

.claude/skills/ux-patterns/navigation-layout.md
```diff
@@ -0,0 +1,159 @@
...
+- Back on step 1 = native navigation (exit the flow)
+- Back on step 2+ = previous step
+- If the form has been modified and user attempts to exit from step 1, prompt before discarding
+- Stepper does not allow skipping steps — each step must be validated before advancing
+
+The wizard exists because all steps require conscious decisions. Do not skip steps programmatically or pre-validate on behalf of the user.
+
+---
+
+## Inline Editing
```
johnpooch: This feels like it doesn't belong in Navigation-Layout and also duplicates a lot of the same section in Form-Controls.

.claude/skills/ux-patterns/navigation-layout.md
```diff
@@ -0,0 +1,159 @@
...
+1. Show value + edit icon (`<Pencil>` button with `aria-label`)
+2. Tap edit: value is replaced with `<Input>` pre-filled with current value
+3. Show Save (`<Check>`) and Cancel (`<X>`) icon buttons to the right
+4. Save/Cancel appear immediately — never below the field
+5. On save success: return to display state
+6. On save failure: show error message inline below the input (not a toast)
+
+---
+
+## Settings: Local vs Global
```
johnpooch: This section is a little bit problematic. Firstly, theme preference is device local as far as I remember, but username is database backed. I think that separating the controls based on whether they are device-specific local or database-backed seems like a little bit of a weird rule. For example, we may have a control to enable push notifications, which is local by nature, and then have a control to enable email notifications, which would be database-backed by nature. Would we really want to separate them into two separate sections?

.claude/skills/ux-patterns/navigation-layout.md
```diff
@@ -0,0 +1,159 @@
...
+The Account screen contains both device-local settings (e.g. Push Notifications) and global account settings (e.g. username, theme preference). These must be visually separated with distinct section headings:
+
+- "This device" or "Notifications" — settings that apply to this device only
+- "Account" — profile settings that apply everywhere
+
+Never mix the two groups in the same section.
+
+---
+
+## Icon-Only Buttons
```
johnpooch: This doesn't feel like it belongs in navigation-layout.md. 

It's also not clear to me what this guidance is referring to. It says that an ARIA label and a visible tooltip are required, but required when?

.claude/skills/ux-patterns/navigation-layout.md
```diff
@@ -0,0 +1,159 @@
...
+Minimum 48×48dp for all interactive elements on mobile. If a visual element is smaller (e.g. a 16×16 icon), expand the hit area with padding. The `size="icon"` variant on `<Button>` handles this for icon buttons — do not reduce it.
+
+---
+
+## Lists: Cards vs Rows
+
+| Content type | Component |
+|-------------|-----------|
+| Rich items with preview (games, variants with map) | `<Card>` or `<ScreenCard>` |
+| Simple items (players, account settings, nav items) | `<Item>` / `<ItemGroup>` from `@/components/ui/item` |
```
johnpooch: I actually think players should gradually be revised into something more like the card as we start to expose more and more of the player information, for example:
- reliability
- rating
- etc.

.claude/skills/ux-patterns/navigation-layout.md
```diff
@@ -0,0 +1,159 @@
...
+
+Within a list, all items must use the same component. Never mix cards and rows.
+
+---
+
+## Notification Badges
```
johnpooch, line 98: This is totally duplicated from another section and doesn't belong in the navigation layout file.

.claude/skills/ux-patterns/navigation-layout.md
johnpooch: Additionally, this doesn't belong in the navigation-layout.md file.

.claude/skills/ux-patterns/navigation-layout.md
```diff
@@ -0,0 +1,159 @@
...
+      <Pencil className="size-4" />
+    </Button>
+  </TooltipTrigger>
+  <TooltipContent>Edit name</TooltipContent>
+</Tooltip>
+```
+
+---
+
+## Touch Targets
```
johnpooch: This section does not belong in the navigation-layout.md file.

.claude/skills/ux-patterns/progressive-disclosure.md
```diff
@@ -0,0 +1,166 @@
...
+
+---
+
+## First-Time Users
+
+New users encounter an onboarding flow before reaching the main UI. This flow explains what Diplomacy is and what the app does — context before action.
+
+The new-player introduction covers most feature types. If a feature is covered in the introduction, the UI adds no additional explanation.
+
+If a feature is NOT covered in the introduction: show a **subtle muted text block below the relevant UI element** until the user takes their first action with it. After the first action, the text does not reappear.
```
johnpooch, line 51: This appeared somewhere else in the documentation already. I think this is probably the place for this documentation to live.

johnpooch:
Reminder/note to self. 

Let's feed the comments I left back into our Claude guidance. Claude should be careful not to duplicate guidance, put guidance in the wrong place, or use code snippets unnecessarily.

JorenC:
I will make another pass on this. This should be growing anyway

.claude/skills/ux-patterns/feedback-patterns.md
JorenC: Discussed with Joren and we kept the length-based formula. The thinking: "a reasonable default unless specified" still needs "reasonable" defined somewhere, and a one-line length-based duration is a cheap way to stop long toasts from disappearing too fast. We did add a rule that action toasts (Undo/Retry) stay for max(formula, action window) so they never vanish before the action is usable. Happy to revisit if you feel strongly.

.claude/skills/ux-patterns/SKILL.md
JorenC: Reconciled. SKILL.md and CLAUDE.md now both say: keep the Suspense wrapper, only swap the empty-div fallback for a skeleton that matches the loaded content's dimensions.

.claude/skills/ux-patterns/interaction-patterns.md
JorenC: Fixed. The Notice prop is `actions` (plural) — corrected here and the same typo in microcopy.md.

.claude/skills/ux-patterns/feedback-patterns.md
JorenC: Resolved. Undo toast duration is now max(duration formula, undo window) so it can't disappear before the action is usable, and Undo is the canonical pattern that this file cross-links.

.claude/skills/ux-patterns/feedback-patterns.md
JorenC: Good point — rewrote it to state this is automatic in the FormField pattern (FormControl sets aria-invalid, Input styles the border, FormMessage renders the message) and dropped the manual snippet.

.claude/skills/ux-patterns/feedback-patterns.md
JorenC: Thanks — left as is.

.claude/skills/ux-patterns/feedback-patterns.md
JorenC: Agreed — changed to a spinner before the label, matching the shadcn example you linked.

.claude/skills/ux-patterns/feedback-patterns.md
JorenC: Agreed — keep the label and show the spinner before it. Updated in both feedback-patterns.md and microcopy.md.

.claude/skills/ux-patterns/feedback-patterns.md
JorenC: Done — now leads with the generic principle (signal urgency via colour escalation) and uses the approaching deadline as the example.

.claude/skills/ux-patterns/feedback-patterns.md
JorenC: Done — added that the skeleton must match the loaded content's dimensions exactly, with fixed heights for variable-height items (game/player cards) to avoid layout shift.

.claude/skills/ux-patterns/feedback-patterns.md
JorenC: Agreed — clarified that the Suspense wrapper stays; only the fallback changes (empty div to skeleton).

.claude/skills/ux-patterns/form-controls.md
JorenC: Removed the Component Reference table.

.claude/skills/ux-patterns/form-controls.md
JorenC: Done — removed the codebase anti-patterns section and folded the teaching into the decision-tree rules as timeless examples.

.claude/skills/ux-patterns/interaction-patterns.md
JorenC: Removed.

.claude/skills/ux-patterns/interaction-patterns.md
JorenC: Resolved — collapsed the duplication; this file now cross-links the canonical Undo pattern in feedback-patterns.md, and the timing contradiction is gone (duration = max(formula, undo window)).

.claude/skills/ux-patterns/interaction-patterns.md
JorenC: Rewrote the Undo steps: execute immediately (it is a real mutation), show a toast with Undo, and on Undo issue a separate reversing request that can itself fail.

.claude/skills/ux-patterns/interaction-patterns.md
JorenC: Added a rule: when the current phase resolves while the user is viewing an older phase, shift to the newest phase anyway and show a toast. One consistent rule beats the rare disorientation, and the toast explains the jump.

.claude/skills/ux-patterns/interaction-patterns.md
JorenC: Done — "automatically refresh" changed to "automatically reload the relevant screen content".

.claude/skills/ux-patterns/interaction-patterns.md
JorenC: Thanks.

.claude/skills/ux-patterns/interaction-patterns.md
JorenC: Removed — it just restated the no-results empty state covered above.

.claude/skills/ux-patterns/interaction-patterns.md
JorenC: Reworked into a principle (match the badge to the kind of signal: binary to dot, quantitative to count, status to semantic colour, metadata to muted text; never colour alone) and deduped to a single home in navigation-layout.md.

.claude/skills/ux-patterns/interaction-patterns.md
JorenC: Removed.

.claude/skills/ux-patterns/interaction-patterns.md
JorenC: Removed — it duplicated the "Read-Only Complexity" guidance in progressive-disclosure.md, which already frames it as a principle.

.claude/skills/ux-patterns/interaction-patterns.md
JorenC: Removed — first-time UX is covered in progressive-disclosure.md (which you also flagged as the right home).

.claude/skills/ux-patterns/interaction-patterns.md
JorenC: Removed.

.claude/skills/ux-patterns/microcopy.md
JorenC: Added a Tone of Voice bullet: no em-dashes, no marketing or LLM-flavoured phrasing.

.claude/skills/ux-patterns/microcopy.md
JorenC: Deduped — the full empty-state pattern lives in interaction-patterns.md; microcopy.md now keeps only the copy guidance and cross-links.

.claude/skills/ux-patterns/navigation-layout.md
JorenC: Deduped — navigation-layout.md is the canonical home. feedback-patterns.md's Confirmation Dialogs now cross-links it and keeps only the dialog copy format.

.claude/skills/ux-patterns/navigation-layout.md
JorenC: Moved the full inline-editing pattern to interaction-patterns.md and removed it from navigation-layout.md.

.claude/skills/ux-patterns/navigation-layout.md
JorenC: Agreed — reorganised to group settings by topic (what they control), with a small "This device" indicator on device-scoped settings instead of splitting by storage location. Your push-vs-email example is exactly the case that breaks the old rule.

.claude/skills/ux-patterns/navigation-layout.md
JorenC: Moved to microcopy.md and clarified the trigger: any button whose only content is an icon (no visible text label) needs both an aria-label and a visible tooltip.

.claude/skills/ux-patterns/navigation-layout.md
JorenC: Noted — added a one-line forward note that players are rows today but may shift toward cards as we surface reliability/rating. Didn't bake the change in yet.

.claude/skills/ux-patterns/navigation-layout.md
JorenC: Deduped to one home. Kept it in navigation-layout.md (Joren's call) but reworked into principles, and removed the interaction-patterns.md copy.

.claude/skills/ux-patterns/navigation-layout.md
JorenC: Moved Lists: Cards vs Rows to form-controls.md — it's a component-selection decision, same family as the other "which component" guidance there.

.claude/skills/ux-patterns/navigation-layout.md
JorenC: Moved Touch Targets to interaction-patterns.md, alongside the other mobile guidance.

.claude/skills/ux-patterns/progressive-disclosure.md
JorenC: Agreed — kept this as the canonical home and removed the duplicate from interaction-patterns.md.

JorenC:
## Review feedback addressed

Thanks for the thorough review, John. I've gone through all 36 comments with Joren and pushed the updates (force-pushed after rebasing onto `main`). Net change is **+84 / −207** — the recurring themes were *deduplicate*, *prefer principles over component specs*, and *drop unnecessary code snippets*, so a lot of this was trimming.

I've replied inline on every thread; high-level summary below.

### Structural / dedupe
- **Suspense fallbacks**: SKILL.md and CLAUDE.md now agree — keep the wrapper, swap the empty-div fallback for a skeleton matching the loaded content's dimensions (fixed heights for variable-height cards to avoid layout shift).
- **Sheet vs Dialog**: `navigation-layout.md` is now the single home; `feedback-patterns.md` cross-links it and keeps only the dialog copy format.
- **Empty states**: full pattern lives in `interaction-patterns.md`; `microcopy.md` keeps only copy guidance + a cross-link.
- **Undo**: now canonical in `feedback-patterns.md`; `interaction-patterns.md` cross-links it. Timing contradiction fixed.
- **Notification Badges**: deduped to one home (`navigation-layout.md`) and reworked into a principle (match the badge to the *kind* of signal; never colour alone).
- **Relocated out of `navigation-layout.md`**: Inline Editing → `interaction-patterns.md`; Touch Targets → `interaction-patterns.md`; Icon-Only Buttons → `microcopy.md`; Lists: Cards vs Rows → `form-controls.md`.

### Removed (out of scope / duplicates / Claude already knows this)
- `form-controls.md`: Component Reference table; codebase anti-patterns section (folded into the decision-tree rules as timeless examples).
- `interaction-patterns.md`: Multi-Select, Accordion (dup of progressive-disclosure), First-Time Encounter (dup), App Updates; plus the `toast.success` and expand/collapse-all snippets.

### Reworked per your notes
- Field-level errors: stated as automatic in the `FormField` pattern, manual snippet dropped.
- Button loading: spinner before the label, label stays (per the shadcn example you linked).
- Urgency: leads with the generic principle, deadline as the example.
- Settings: grouped by topic with a small "This device" indicator, instead of splitting by storage location (your push-vs-email example was the clincher).
- Icon-only buttons: clarified the trigger — any button whose only content is an icon needs both `aria-label` and a visible tooltip.
- `microcopy.md` tone: added "no em-dashes, no marketing/LLM-flavoured phrasing".
- Phase resolution: when the current phase resolves while viewing history, shift to newest + show a toast (one consistent rule).
- Fixed the blocking `action` → `actions` typo on `Notice`.

### One thing I pushed back on
- **Toast duration formula** (`feedback-patterns.md`): we kept the length-based formula rather than a single default. Reasoning in the thread — happy to simplify if you still prefer that.

.claude/skills/ux-patterns/feedback-patterns.md
johnpooch: I'm still not convinced about this, but it's not a hill I'm going to die on. Some toast messages come from the backend. How is Claude meant to know what the length of those messages is? I think just going with something like two or three seconds is simple. But your call.

johnpooch (approved):
Approved with a small pushback that I still think the toast duration guidance is over-engineered. But I'm happy for this to be merged.

JorenC:
Hahahahhahahha!

I think we can add a fallback if the front end doesn't know. But I think by
default it's good practice - these notifications are important, and if they
are too fast it's very dangerous info gets missed (we rely on them quite
heavily).

Right now they are "Couldn't join game" (bad). They should be "There was a
problem reaching the server. Couldn't join game. Try again later". Or "A
player joined before you, the game is full. Couldn't join game. Find
another one" etc etc.

This needs more time.


On the other hand, a 10 second "Couldn't join game" feels unnecessary slow.

JorenC:
But thanks for merging.

We can also update as we go. Not all my calls are perfect and you have good
calls, so let's see. Maybe we learn more and change

PR #803, opened 2026-06-20
Reintroduce neutral (non_playable) nation support

### LINKED ISSUE #624: PRD: Reintroduce Neutral Nations Support
@johnpooch maybe we do want to slightly prioritize this - it unlocks quite some features and variants. 
One thing to think about is sandboxes - I think that in a sandbox we should just use whatever the default behaviour is - if they can't be given orders, that's fine (it can't be done in the variant/reality anyway). If they CAN be given orders, that's fine as well - it's to test any specific scenario/play around. It's all up to the player, but I don't think we need to put in extra effort for specific Sandbox behaviour.

## Background

Neutral nations were previously implemented in PR #573, which introduced a `non_playable` boolean field on the `Nation` model. The implementation correctly excluded neutral nations from all player-seating code paths, allowing variants to define nations that exist on the map but are never controlled by a player.

The feature was merged in commit `a88ba73`.

However, the implementation was later reverted as part of PR #616 ("Fix variant upload 500: drop orphaned non_playable column"), which introduced migration `0013_remove_non_playable`. This migration removed the `non_playable` field entirely from the database schema and the current codebase no longer contains support for neutral nations.

As a result, the Variant Creator cannot currently emit `non_playable: true`, even though the concept was already designed and implemented once. If emitted today, the backend would simply treat the nation as a normal playable power.

## Motivation

Several historical and scenario-based variants require neutral forces that occupy territory but are not controlled by a player.

Examples include:

- Sengoku
- Historical Diplomacy scenarios with independent garrisons
- Variants featuring neutral supply centers protected by standing units

Supporting neutral nations would unlock a number of previously-created Diplicity variants and expand the capabilities available to the growing variant-building community.

## Goal

Reintroduce the neutral nation system from PR #573 in a stable manner, resolving the issues that led to the variant upload 500 errors and subsequent rollback.

### Neutral Nations

- A nation may be marked as `non_playable`.
- Non-playable nations are excluded from all player-assignment logic.
- Neutral nations may own supply centers.
- Neutral nations may own units at game start.

### Neutral Units

- Neutral units never receive orders.
- Neutral units automatically hold every turn.
- Neutral units cannot be issued orders by players.
- If a neutral unit is dislodged, it is immediately removed from the board.
- Neutral units do not retreat.

### Adjustments

- Neutral nations never build units.
- Neutral nations never participate in adjustment phases.
- Neutral nations never voluntarily disband units.
- The only way a neutral unit leaves the board is by being dislodged.

## Variant Definition Support

- Restore the `non_playable` field to the `Nation` model.
- Ensure variant import/export preserves the field.
- Ensure variant validation accepts neutral nations.
- Allow Variant Creator to emit `non_playable: true`.
- Maintain backwards compatibility with existing variants.

## Acceptance Criteria

- Variants can define one or more neutral nations.
- Neutral nations are never assigned to players.
- Neutral units automatically hold every turn.
- Dislodged neutral units are immediately removed without retreat.
- Existing variants continue to function unchanged.
- Sengoku and similar variants can be represented without engine modifications.

## Notes

This feature is largely a restoration of functionality already implemented in PR #573 and merged in commit `a88ba73`, with the primary work being to resolve the issues that resulted in its removal via PR #616.

johnpooch:
Closes #624

Restores the neutral nation system originally implemented in PR #573 and merged in `a88ba73`, which was removed when PR #616 dropped the orphaned `non_playable` column (migration `0013_remove_non_playable`). The original revert was caused by a botched partial revert leaving a NOT NULL column with no default — a schema problem, not a logic one — so this change re-applies PR #573's code (adapted to the evolved `adjudication/service.py` skip logic) plus a fresh migration `0014_add_non_playable`.

### What it does
- **`Nation.non_playable`** boolean field, exposed through the nation serializer, the variant JSON schema, and the variant import/export round-trip (`variant_to_canonical_dict` / `_build_variant_children` / safe-replacement).
- **Seating** excludes neutral nations everywhere a game counts powers: `IsSpaceAvailable`, `Game.start`, sandbox creation, sandbox clone, and member auto-start.
- **Adjudication**: neutral nations never build or disband (`allowed_builds` / `required_disbands` return 0), and retreat/adjustment phases where *only* neutral nations have options are skipped (`skip_nations = cd_nations | non_playable_nations`). The skip path means a dislodged neutral unit is removed from the board without retreating, and unordered neutral units hold by default — satisfying the PRD's neutral-unit behaviour without engine-internal changes.

### Notes for the reviewer
- PR #573 shipped without tests; this PR adds engine-level tests for the build/disband gating and a variant-upload round-trip test for field preservation.
- The OpenAPI schema and generated `endpoints.ts` were updated for the single new `nonPlayable` field only. I deliberately avoided committing the full `spectacular`/`orval` regen because in this environment it also dropped unrelated, environment-dependent entries (`/devices/`, `/api/test-sentry/`, and an `OrderStatusEnum` value) — pre-existing drift unrelated to this change.

No frontend UI changes (the field is data-only; the Variant Creator emitting `non_playable: true` is out of scope here).

---
_Generated by [Claude Code](https://claude.ai/code/session_012odxzBkfoRFYzE5XQcoVAk)_

### DIFF
--- a/packages/web/src/api/generated/endpoints.ts
+++ b/packages/web/src/api/generated/endpoints.ts
@@ -75,6 +75,7 @@ export interface Nation {
   nationId: string;
   name: string;
   color: string;
+  nonPlayable: boolean;
   /** @nullable */
   readonly flagUrl: string | null;
 }
--- a/packages/web/src/components/InteractiveMap/toRenderState.test.ts
+++ b/packages/web/src/components/InteractiveMap/toRenderState.test.ts
@@ -17,8 +17,8 @@ const province = (id: string, parentId: string | null = null): Province => ({
 });
 
 const nations: Nation[] = [
-  { nationId: "england", name: "England", color: "#2196F3", flagUrl: null },
-  { nationId: "france", name: "France", color: "#80DEEA", flagUrl: null },
+  { nationId: "england", name: "England", color: "#2196F3", nonPlayable: false, flagUrl: null },
+  { nationId: "france", name: "France", color: "#80DEEA", nonPlayable: false, flagUrl: null },
 ];
 
 const variant = { nations };
--- a/packages/web/src/mocks/legacy.ts
+++ b/packages/web/src/mocks/legacy.ts
@@ -13,13 +13,13 @@ import type {
 } from "@/api/generated/endpoints";
 
 export const mockNations: Nation[] = [
-  { nationId: "austria", name: "Austria", color: "#FF0000", flagUrl: null },
-  { nationId: "england", name: "England", color: "#0000FF", flagUrl: null },
-  { nationId: "france", name: "France", color: "#00FFFF", flagUrl: null },
-  { nationId: "germany", name: "Germany", color: "#000000", flagUrl: null },
-  { nationId: "italy", name: "Italy", color: "#00FF00", flagUrl: null },
-  { nationId: "russia", name: "Russia", color: "#FFFFFF", flagUrl: null },
-  { nationId: "turkey", name: "Turkey", color: "#FFFF00", flagUrl: null },
+  { nationId: "austria", name: "Austria", color: "#FF0000", nonPlayable: false, flagUrl: null },
+  { nationId: "england", name: "England", color: "#0000FF", nonPlayable: false, flagUrl: null },
+  { nationId: "france", name: "France", color: "#00FFFF", nonPlayable: false, flagUrl: null },
+  { nationId: "germany", name: "Germany", color: "#000000", nonPlayable: false, flagUrl: null },
+  { nationId: "italy", name: "Italy", color: "#00FF00", nonPlayable: false, flagUrl: null },
+  { nationId: "russia", name: "Russia", color: "#FFFFFF", nonPlayable: false, flagUrl: null },
+  { nationId: "turkey", name: "Turkey", color: "#FFFF00", nonPlayable: false, flagUrl: null },
 ];
 
 export const mockMembers: Member[] = [
--- a/packages/web/src/screens/Tutorial/useTutorialEngine.test.ts
+++ b/packages/web/src/screens/Tutorial/useTutorialEngine.test.ts
@@ -23,6 +23,7 @@ const nation = (nationId: string): Nation => ({
   nationId,
   name: nationId,
   color: "#ffffff",
+  nonPlayable: false,
   flagUrl: null,
 });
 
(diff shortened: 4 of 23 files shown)


JorenC:
https://github.com/JorenC/variant-creator/pull/28 - if we can merge this, I can pull my PR on the variant creator and neutrals will work - it's in sync.

johnpooch (commented):
Two findings below. The adjudicator-level logic and the variant round-trip test are solid. The seating-side changes are the gap.

---

**Finding 1 — Missing tests for the seating-side changes (significant)**

Four Django touchpoints changed their member/nation counting to exclude `non_playable` nations, but none have test coverage for that behaviour:

- `IsSpaceAvailable` (`permissions.py`) — now calls `.filter(non_playable=False).count()`
- `member/views.py` auto-start trigger — same filter
- `game/models.py` `create_sandbox` and `start` — Python list-comp filter
- `game/serializers.py` sandbox clone — same list-comp filter

The adjudicator tests verify that `allowed_builds` / `required_disbands` return 0 for non-playable nations, and the variant round-trip test covers field persistence. But there are no API-level tests that verify:

1. A join request is rejected (or accepted) at the correct threshold when the variant has a non-playable nation.
2. The game auto-starts when the right number of *playable* seats are filled.
3. Sandbox creation and clone seat only playable nations.

Per the project testing standard, these behavior changes should be covered by API endpoint tests in `game/tests/` and `member/tests/` that use a fixture variant containing at least one non-playable nation.

---

**Finding 2 — Schema description asserts a constraint that isn't validated (minor)**

`service/variant.schema.yaml` says:

> Non-playable nations must have as many units as supply centers they own in initialState so the adjustment phase stays balanced.

The code does not validate this constraint at import time, and it doesn't need to: `_is_non_playable()` gates both `allowed_builds` and `required_disbands` to 0 regardless of the unit/SC ratio, so any initial imbalance simply persists permanently (neutral units never build or disband). The "must" wording is therefore wrong and may confuse variant creators who could worry needlessly about balancing their neutral forces.

Suggest either removing the sentence or rewording it to describe what *will* happen ("units in excess of supply centers will never disband; the engine handles any ratio").


---

JorenC:
This is what I meant with unclarity for status in the accelerated retreats.

Did you merge it? Did the bot (pull into main?) You pulled in a commit into main - the tests? Neutrals? Both? 

Especially since this needs me to update the variant-creator and I clarified that, it's important I get a tag.
@johnpooch

PR #815, opened 2026-06-20
Make email login case-insensitive

### LINKED ISSUE #814: Make email login case-insensitive
## Goal

Email login should be case-insensitive, matching standard email behavior.

## Current behaviour

Email lookups during login, registration, and password reset are case-sensitive. A user who registered with `Player@Example.com` cannot log in with `player@example.com`.

## Expected behaviour

Email lookups should be case-insensitive. Users can log in with any case variation of their registered email address.

## Where

`service/login/serializers.py` — `EmailLoginSerializer.create()`, `RegisterSerializer.validate_email()`, and `PasswordResetSerializer.create()` all use case-sensitive email filters.

johnpooch:
Closes #814

Changes `email=` to `email__iexact=` in the three serializer methods that look up users by email: `EmailLoginSerializer.create`, `RegisterSerializer.validate_email`, and `PasswordResetSerializer.create`. Users who registered with any case variation of their email can now log in, reset their password, or be rejected for duplicate registration regardless of case.

A test was added to each of the three affected test files to assert the case-insensitive behaviour directly.

---
_Generated by [Claude Code](https://claude.ai/code/session_01VURXawa9KzpHjuBR2mF6yV)_

### DIFF
--- a/service/login/serializers.py
+++ b/service/login/serializers.py
@@ -23,7 +23,7 @@ class EmailLoginSerializer(serializers.Serializer):
     refresh_token = serializers.CharField(read_only=True)
 
     def create(self, validated_data):
-        user = User.objects.filter(email=validated_data["email"]).first()
+        user = User.objects.filter(email__iexact=validated_data["email"]).first()
         if user is None or not user.check_password(validated_data["password"]):
             raise exceptions.AuthenticationFailed("Invalid email or password.")
         if not user.is_active:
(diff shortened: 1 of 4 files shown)


johnpooch (commented):
Clean, minimal fix. Three one-line changes, one test per changed code path, all 19 tests green.

One gap worth noting for future reference (out of scope for this issue): `AuthUserManager.create_from_google_id_info` and `create_from_apple_id_info` in `login/models.py` still use `email=` (case-sensitive `get_or_create`). If an OAuth provider ever sent a mixed-case email matching an existing user stored with different casing, it would create a duplicate account rather than find the existing one. Google and Apple normalise to lowercase in practice, so this isn't an immediate problem — but it's a logical complement to this fix.

---

PR #819, opened 2026-06-21
Accelerated retreat/adjustment phases

### LINKED ISSUE #775: Accelerated Retreat/Adjustment Phases
## Goal

Implement Accelerated Retreat/Adjustment Phases end-to-end in a single PR: in fixed-time games, let non-movement phases resolve early when all eligible players confirm within a creator-set window, while structurally guaranteeing a minimum movement-turn length. Includes the backend mechanic/API and the player-facing UI.

## Context

Full design and rationale are in the epic #546. Today fixed-time games never resolve early (one clause in `PhaseQuerySet.filter_due_phases`). The early-resolution plumbing already exists end-to-end via `resolve_phase` → `resolve_if_due`; this work relaxes that clause for non-movement phases inside the window and surfaces the setting and countdown in the UI.

## Approach

**Backend (`service/`)**
- Add `Game.accelerated_phase_window_seconds` (PositiveIntegerField, null=True; NULL = disabled, positive = window `W` in seconds).
- Add `Phase.early_resolve_window_end` (DateTimeField, null=True). Do not add an `is_accelerated` flag or new phase status.
- In `PhaseManager.create_from_adjudication_data`, set `early_resolve_window_end` only for fixed-time, non-movement phases when the feature is enabled: `T0 + W` where `T0` is the previous movement phase's `scheduled_resolution`; inherit the value unchanged when chaining minor phases; clamp to the new phase's own `scheduled_resolution`. Movement phases stay NULL.
- Add an `accelerated_due` branch to `filter_due_phases`: `deadline_mode == FIXED_TIME` AND `type != MOVEMENT` AND `early_resolve_window_end` not null AND `early_resolve_window_end >= now` AND all eligible confirmed. No change to the confirm endpoint, `get_scheduled_resolution`, or task plumbing.
- Expose `early_resolve_window_end` on `PhaseRetrieveSerializer`.
- A lapsed window is a no-op — never an NMR/extension. Pause/resume needs no special handling in v1.

**Frontend (`packages/web/`)**
- **`CreateGame.tsx`, Fixed Time tab only** (hidden when `deadlineMode === "duration"`): checkbox "Accelerated Retreat/Adjustment Phases" with a `?` tooltip (hover desktop / tap mobile); a discrete slider selecting `W` with readouts left "Acceleration window" = `W` and right "Minimum movement duration" = `movement_cycle − W`. Stops are linear/computed (no curated list): `step = movement_cycle / 12`; minimum selectable window is one step, maximum is `movement_cycle − one step` (full-cycle excluded), 11 positions; slider keeps constant length and rescales on frequency change. Multi-frequency clamp: `max W = min(retreat_cycle, adjustment_cycle, movement_cycle − step)`. **Enabled by default**: on a new game the checkbox is pre-checked and the default window `W = 2 × step` is applied (e.g. 4h daily → 20h floor); unchecking disables it. The `Game.accelerated_phase_window_seconds` model default stays NULL (= disabled) so non-UI creation paths are unaffected — the on-by-default behaviour lives in the create-game form, which sends the computed `W` whenever the box is checked.
- **`DeadlineSummary.tsx`**: unchanged when off; when enabled, "Movement resolves {frequency} at {time} {tz}. Retreat and adjustment phases can resolve early if everyone confirms within {W}, but movement turns are always at least {movement_cycle − W}."
- **`OrdersScreen.tsx`**: for non-movement fixed-time phases with the feature enabled, replace the `deadlineMode === "fixed_time"` early-return so the confirm button shows. While `now < early_resolve_window_end`: active button + "Acceleration window closes in {time} — confirm to resolve early." After it closes: button shown disabled + "Acceleration window closed — this phase will run its full length." Movement phases keep today's behaviour.
- Use the exact copy strings from §7 of #546.

**Cross-cutting**
- Run `docker compose up codegen` after the serializer/field changes, then `npx tsc -b --noEmit` in `packages/web`; add the new field to any inline game objects in mocks/tests.
- Tests: backend covers the new `filter_due_phases` branch (open vs closed window, movement excluded, duration mode unaffected), `early_resolve_window_end` computation incl. chaining and clamp, the `movement_cycle − W` floor guarantee, and that a lapsed window causes no NMR; frontend covers the create-game setting (slider stops/rescaling/readouts/default), both `DeadlineSummary` strings, and the `OrdersScreen` confirm states.
- Screenshots are mandatory: the create-game setting (checkbox + tooltip + slider with both readouts), the in-game confirm button in both window-open and window-closed states, and the updated `DeadlineSummary`. Add a `RELEASE_NOTES.md` entry.

johnpooch:
Closes #775

Non-movement phases in fixed-time games can now resolve early when all eligible players confirm within a creator-set window. Movement phases always run their full cycle.

## Backend

- `Game.accelerated_phase_window_seconds` field (nullable int) + migration
- `Phase.early_resolve_window_end` field (nullable datetime) + migration
- `PhaseQuerySet.filter_due_phases()` gains an accelerated-resolution branch: resolves a non-movement fixed-time phase early when its `early_resolve_window_end` is in the future and all members have confirmed
- `Phase.create_from_adjudication_data()` sets `early_resolve_window_end = T0 + W` on retreat/adjustment phases (T0 = previous movement phase scheduled resolution, W = window from game), clamped to `scheduled_resolution`
- `GameCreateSerializer.validate()` strips `accelerated_phase_window_seconds` for non-fixed-time games

## Frontend

- **Create Game → Deadlines tab**: checkbox "Accelerated Retreat/Adjustment Phases" (enabled by default when fixed-time is selected) with a slider for the window size and readouts for the window duration and minimum movement floor
- **`DeadlineSummary`**: new acceleration branch describes the window and minimum floor in plain text
- **OrdersScreen confirm button**: hidden for movement phases in fixed-time games; shown active while window is open; shown disabled with label "Acceleration window closed" once the window lapses
- New `Slider` shadcn-style component backed by `@radix-ui/react-slider`
- New mock fixture `activeGameRetreatWindowClosed` for the window-closed state

## Screenshots

**Create Game – Deadlines tab (acceleration slider)**
create game deadlines

**Orders screen – window open (active confirm button)**
retreat window open

**Orders screen – window closed (disabled button)**
retreat window closed

---
_Generated by [Claude Code](https://claude.ai/code/session_011YAyVq8oppzfi4fG8MHSyj)_


johnpooch (commented):
## Review

The feature logic is solid — the backend filter, window-end propagation, and frontend confirm-button states are all correct. But there are a few issues that need addressing before merge.

---

### Critical: `GameListSerializer` missing the new field

The `openapi-schema.yaml` and the generated `GameList` TypeScript interface both declare `acceleratedPhaseWindowSeconds`, but `GameListSerializer` in `serializers.py` has **no `accelerated_phase_window_seconds` field**. The schema was manually patched rather than generated; running `docker compose up codegen` from a correct serializer would produce a different result.

Any frontend code that reads `game.acceleratedPhaseWindowSeconds` from a game object fetched via a list endpoint (e.g. home-screen game cards, `DeadlineSummary` inside a game list row) will receive `undefined` at runtime, not `null`. The TypeScript type hides this because the generated interface says `number | null`, but that assumes the serializer actually emits the field.

Fix: add `accelerated_phase_window_seconds = serializers.IntegerField(read_only=True, allow_null=True)` to `GameListSerializer`, then re-run codegen.

---

### Significant: inline imports in backend tests (CLAUDE.md)

Every test method in `TestFilterDuePhasesAcceleratedWindow` and `TestCreateFromAdjudicationDataEarlyResolveWindow` starts with inline imports:

```python
def test_returns_phase_when_window_open_and_all_confirmed(self, ...):
    from django.utils import timezone
    from datetime import timedelta
    from game.models import Game
    from common.constants import DeadlineMode, GameStatus, MovementPhaseDuration
```

CLAUDE.md is explicit: "Always place imports at the top of the file — Do not use inline imports inside methods." These should all be hoisted to the top of `tests.py`.

---

### Significant: duplicated `FREQUENCY_SECONDS` and `formatDurationSeconds`

Both `CreateGame.tsx` and `DeadlineSummary.tsx` define identical copies of these. Extract them to a shared utility (e.g. `src/utils/deadline.ts`) and import from there. The issue spec even references both files needing the same duration formatting — this is a clear extraction candidate.

---

### Moderate: `screenshot-interact.mjs` should not be committed

`packages/web/scripts/screenshot-interact.mjs` is a hard-coded, one-off automation script that navigates to `/create-game`, clicks Next, and screenshots it. There is no mechanism to run it reproducibly (no npm script entry, no documentation), and it duplicates the pattern of the existing `screenshot.mjs`. Either delete it, or generalise it into the existing screenshot tooling.

---

### Minor: missing `<FormMessage />` in checkbox `FormField`

CLAUDE.md requires every `FormField` to include `FormItem`, `FormLabel`, `FormControl`, and `FormMessage`. The new `acceleratedPhaseEnabled` checkbox `FormField` is missing `<FormMessage />`. Add it for consistency even though the `z.boolean()` schema will never fail.

---

### Observation: `activeGameRetreat` fixture modified

The existing `activeGameRetreat` fixture now has `deadlineMode: "fixed_time"` and `acceleratedPhaseWindowSeconds: 14400` added. This changes the fixture used by any existing stories or tests that relied on it. Confirm that any affected stories still render correctly and that the fixture description accurately reflects the new scenario. (Adding a new fixture and leaving the original unchanged is the lower-risk approach.)

---

PR #826, opened 2026-06-22
Add rebuilds flag for non-playable nations

### LINKED ISSUE #825: Neutral nations rebuild units on undefended supply centres
## Goal

Add an opt-in variant option that makes neutral (non-playable) nations automatically rebuild a unit on any of their owned supply centres that has no unit, during each adjustment phase.

## Context

Neutral nation support (PR #803) gates neutrals out of builds and disbands entirely, so a neutral unit that is disbanded without its SC being captured never returns. This new variant flag restores those units each adjustment phase. The option must default to false when absent from the variant definition, preserving existing variants' behaviour.

## Approach

Add a boolean field to the variant definition (default false, carried through the variant JSON schema and import/export round-trip alongside `non_playable`). When enabled, the adjudicator's adjustment phase auto-builds a unit on each neutral-owned supply centre lacking a unit. If a neutral nation has more units than supply centres, no build is attempted (either skipping or letting the adjudicator reject the excess is acceptable).

johnpooch:
Closes #825

Adds a `rebuilds` boolean flag to non-playable nations. When `true`, the adjustment phase automatically places a unit on each owned supply centre that has no unit. Defaults to `false`, so all existing variants are unaffected.

## What changed

**Data model** — `rebuilds` field added to the `Nation` Django model (with migration), the adjudicator's `Nation` domain dataclass, the variant JSON schema, the import/export round-trip in `variant/utils.py`, and the `NationSerializer`.

**Adjudicator** — new `ApplyNonPlayableRebuilds` action and reducer inserted between `FinalizeStatuses` and `ApplyAdjustmentOutcomes` in the adjustment-phase pipeline. The reducer finds all non-playable nations with `rebuilds=True`, compares their owned SCs against occupied positions, and records `Unit` objects in a new `auto_rebuild_units` state field. `ApplyAdjustmentOutcomesReducer` appends these units to `next_units`. Unit type follows province type: sea → Fleet, otherwise → Army.

**Tests** — four new tests cover the happy path (unit rebuilt on empty SC), no-rebuild when unit already present, no-rebuild when `rebuilds=False`, and no-rebuild when unit count already meets or exceeds SC count. All 372 adjudicator tests pass.

---
_Generated by [Claude Code](https://claude.ai/code/session_01EjozPFfJBy4HKW155YbY9b)_

### DIFF
--- a/service/adjudicator/engine.py
+++ b/service/adjudicator/engine.py
@@ -860,35 +860,6 @@ class ApplyCivilDisorderReducer(Reducer):
         return state.replace(civil_disorder_disbands=tuple(sorted(civil)))
 
 
-class ApplyNonPlayableRebuildsReducer(Reducer):
-    """For each non-playable nation with rebuilds=True, find owned supply
-    centres with no standing unit and record them in auto_rebuild_units
-    for ApplyAdjustmentOutcomes to append."""
-
-    ACTION = Actions.ApplyNonPlayableRebuilds
-
-    @classmethod
-    def reduce(cls, state: StateView, action: Action) -> StateView:
-        rebuild: List[Unit] = []
-        for nation in state.variant().nations:
-            if not nation.non_playable or not nation.rebuilds:
-                continue
-            nation_view = state.nation(nation.id)
-            owned = nation_view.owned_supply_centers()
-            occupied = frozenset(
-                u.location
-                for u in state.units().all()
-                if u.nation == nation.id and not u.dislodged
-            )
-            if len(occupied) >= len(owned):
-                continue
-            for sc in sorted(owned - occupied):
-                province = state.variant().provinces[sc]
-                unit_type = Unit.FLEET if province.type == ProvinceType.SEA else Unit.ARMY
-                rebuild.append(Unit(nation=nation.id, type=unit_type, location=sc))
-        return state.replace(auto_rebuild_units=tuple(rebuild))
-
-
 class FinalizeStatusesReducer(Reducer):
     """Promote any order whose status is still None to its final value.
     Support orders read their final status from support_cut and
--- a/service/adjudicator/engine.py
+++ b/service/adjudicator/engine.py
@@ -7,6 +7,7 @@ from typing import ClassVar, Dict, List, Tuple, Type
 from .convoy import convoy_path_exists, is_convoy_redundant
 from .domain import (
     Phase,
+    ProvinceType,
     Resolution,
     State,
     SupplyCenter,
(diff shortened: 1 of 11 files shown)


JorenC:
@johnpooch what do you think?

service/adjudicator/engine.py
```diff
@@ -852,6 +860,35 @@ def reduce(cls, state: StateView, action: Action) -> StateView:
         return state.replace(civil_disorder_disbands=tuple(sorted(civil)))
 
 
+class ApplyNonPlayableRebuildsReducer(Reducer):
```
johnpooch: `ApplyNonPlayableRebuildsReducer` is defined twice in this file — once here (around line 863) and again around line 1152. Python silently uses the last definition, so this first one is dead code. The two bodies are functionally identical (the only difference is a local `nation_view` variable in the first vs. an inline call in the second). Delete this first definition and keep the one at line 1152.

---

service/adjudicator/types.py
```diff
@@ -1071,6 +1077,12 @@ def _is_non_playable(self) -> bool:
                 return nation.non_playable
         return False
 
+    def _is_rebuilds(self) -> bool:
```
johnpooch: `_is_rebuilds` is never called — the reducer accesses `nation.rebuilds` directly from `state.variant().nations` rather than going through this `NationView` helper. Delete it.

---

johnpooch (commented):
Two cleanups needed before merge:

1. **Duplicate `ApplyNonPlayableRebuildsReducer`** — the class is defined twice in `engine.py`. Python silently overwrites the first with the second, making the first dead code. Delete the first definition (line ~863).

2. **Unused `_is_rebuilds` method** in `types.py` — added to `NationView` but never called; the reducer reads `nation.rebuilds` directly from the variant. Delete it.

The logic itself looks correct: the pipeline placement (after `ApplyCivilDisorder`, before `FinalizeStatuses`), the SC/unit comparison, the sea→Fleet / other→Army heuristic, the `auto_rebuild_units` state field, and the four test cases all look right. The migration, serializer, schema, and import/export round-trip are all consistent.

---

JorenC:
What is the next step now - will the bot pick it up again?

johnpooch:
@JorenC I think Claude screwed this up. The issue clearly calls out the field should be "per-variant" and not "per-nation". I also don't like the name `rebuilds` as a field. It reads weirdly.

johnpooch:
Closing this because I think that there is a simpler fix.

PR #829, opened 2026-06-22
Refactor CLAUDE.md: monolith → lean core + domain docs

JorenC:
## Summary

- `CLAUDE.md` was 10,717 words and loaded into **every conversation**, burning budget on content only needed in specific domains (iOS builds, Android credentials, Metabase SQL, production debugging, etc.)
- Replaced with a 987-word core that covers only always-relevant content, pointing to seven domain docs under `.claude/docs/`
- Domain docs are read on demand — a session doing frontend work never pays for the iOS or Metabase content

## What changed

**`CLAUDE.md` (always loaded):** trimmed to ~90% smaller — keeps the evidence-based reasoning requirement, external service UI warning, project overview, dev setup, key commands, development guidelines, GitHub workflow, and the five core code philosophy tenets. Strips all code examples, "what to check in review" blocks, and the PR checklist.

**New `.claude/docs/` files (read on demand):**

| File | Contents |
|---|---|
| `frontend.md` | React patterns, state management, Suspense wrapper, forms, mocks, fixture registry, screenshot workflow |
| `backend.md` | Django views, serializers, models, permissions, constants, transactions, codegen |
| `testing.md` | Frontend + backend test patterns, fixture naming conventions, conftest organisation, performance tips |
| `production.md` | Railway CLI token conflict pattern, pgweb queries, write safety rules, staging environments |
| `android.md` | Build workflow, SHA-1 fingerprints, keystore credentials, Fastlane, GitHub secrets |
| `ios.md` | Xcode build, APNs/ASC credentials, Fastlane signing strategy, GitHub secrets |
| `metabase.md` | Native SQL encoding pattern, roundtrip-assert, schema gotchas |

## Token impact

| | Words |
|---|---|
| Before (CLAUDE.md, every conversation) | 10,717 |
| After (CLAUDE.md, every conversation) | 987 |
| Domain docs (on demand, 1–2 per session) | ~700–1,500 |

Baseline cost per conversation reduced by ~91%.

## No visual changes


JorenC:
@johnpooch please pull this so I can continue with a normal burn rate on my end... :)

johnpooch:
Thanks for tackling this. The baseline `CLAUDE.md` was overdue for a trim, and splitting it up is the right direction. Before we merge, I'd like us to take this one step further and encode the domain guidance as **Claude Code skills** rather than plain docs referenced from `CLAUDE.md`. Here's my reasoning.

### Both approaches keep the bulk out of baseline — the difference is *triggering*

The ~91% baseline reduction in this PR is real and would hold either way, because in both cases the bulk content lives outside the always-loaded `CLAUDE.md`. The meaningful difference is **how reliably the right content gets pulled in at the right moment**:

- **This PR (`.claude/docs/*.md`):** the lean `CLAUDE.md` carries a prose pointer ("Read the relevant doc before starting work in that area"). For it to work, a session has to read `CLAUDE.md`, remember the table, and *choose* to `Read` the file before working. That's a soft instruction — easy to skip mid-task.

- **Skills:** the harness automatically injects every skill's `name` + `description` into context each session, and the model is trained to invoke a skill when its `description` matches the task. The `description: ... Use when ...` field is a purpose-built "load this conditionally when relevant" signal — exactly the behaviour we're reaching for here.

So skills give us the same token savings *plus* a far more reliable conditional-load mechanism, at a negligible extra always-on cost (one short description line per skill).

### We already have a precedent — and this PR would duplicate it

We already ship a `react` skill at `.claude/skills/react/` (with bundled `react-components.md`, `react-data-loading.md`, `react-hook-form.md`, etc.). This PR's new `.claude/docs/frontend.md` overlaps it heavily, which would leave frontend guidance with **two sources of truth**. Standardising on skills lets us fold the existing `react` skill into the new `frontend` skill and remove that duplication rather than entrench it.

### Suggested shape

- Keep the lean `CLAUDE.md` from this PR.
- Convert the seven domain docs into skills under `.claude/skills/` (`android`, `ios`, `backend`, `frontend`, `metabase`, `production`, `testing`), each with a tight `name` + `Use when…` `description`, bundling sub-files like the existing `react` skill does.
- Fold the existing `react` skill into `frontend` so there's one source of truth.

Net: same baseline savings, more reliable on-demand loading, and we kill the existing duplication instead of adding to it. Happy to pair on the conversion if useful.

### While we're at it

When you push the revised PR, I would like to dig through all of the Claude guidance and make sure that all of it is still accurate or relevant. It has accumulated over time and probably needs a haircut.

JorenC:
Updated per your feedback. The seven `.claude/docs/` files are now skills under `.claude/skills/` — the existing `react` skill is folded into the new `frontend` skill and removed.

Main difference from the original PR: instead of CLAUDE.md carrying a prose pointer ("read this doc before working in this area"), the harness auto-injects each skill's name + description into every session, so the right guidance is triggered automatically without relying on Claude reading the table and choosing to act on it.

Happy to address anything that looks off in the content review you mentioned.

JorenC:
Needs a review though - also from my part

JorenC:
This helps me a lot to fix - need to do something about it

PR #831, opened 2026-06-23
Auto-select first variant when switching category tabs on Create Game screen

### LINKED ISSUE #804: Switching variant category clears the variant selection instead of picking the first one

johnpooch:
Closes #804

When the user switches between Official and Community tabs on the Create Game screen, the variant select now auto-selects the first variant in the new category instead of clearing the selection. When returning to a previously visited category, the last-selected variant in that category is restored.

**Implementation notes:**

`flushSync` is used to force the React category state update to complete before calling `form.setValue`. Without this, Radix UI's hidden native `<select>` element (which syncs the controlled value via a `useEffect`) resets the form field to `""` because the new variant's `<option>` isn't mounted yet when RHF writes the new value. `flushSync` ensures the new `SelectItem` options are mounted before `form.setValue` runs.

---
_Generated by [Claude Code](https://claude.ai/code/session_01PyhyQe6hrsWf7geSpCgLqj)_

### DIFF
--- a/packages/web/src/screens/Home/CreateGame.test.tsx
+++ b/packages/web/src/screens/Home/CreateGame.test.tsx
@@ -706,6 +706,56 @@ describe("CreateGame — variant category toggle", () => {
     ).not.toBeInTheDocument();
   });
 
+  it("auto-selects the first community variant when switching to the Community tab", async () => {
+    const user = userEvent.setup();
+    mockedUseVariantsListSuspense.mockReturnValue({
+      data: variantsWithCommunity,
+    } as unknown as ReturnType<typeof useVariantsListSuspense>);
+    renderCreateGame();
+
+    await user.click(screen.getByRole("tab", { name: /community/i }));
+    await user.click(screen.getByRole("button", { name: /next/i }));
+    await user.click(screen.getByRole("button", { name: /next/i }));
+    await user.click(screen.getByRole("button", { name: /create game/i }));
+
+    await waitFor(() => expect(createGameMutateAsync).toHaveBeenCalled());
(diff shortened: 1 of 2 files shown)


johnpooch (commented):
## Review

Clean fix. The `flushSync` approach is the right call here — it avoids a `useEffect` (which would cause an intermediate render with an invalid selection) and the comment clearly explains the non-obvious Radix constraint.

**`flushSync` batching** — both `setLastSelectedByCategory` and `setVariantCategory` are inside a single `flushSync` callback. In React 18, updates inside `flushSync` are batched into one synchronous re-render, so Radix's hidden `<select>` sees both state changes at once before `form.setValue` runs. This is the correct pattern.

**State initialization** — `lastSelectedByCategory` is initialized with only `{ [initialCategory]: defaultVariantId }`. The other category starts as `undefined`, so the first visit falls back to `targetVariants[0]?.id`. Correct, and the `?? ""` fallback for an empty variant list matches the old behaviour — no regression.

**Tests** cover both required behaviours through the observable outcome (submitted `variantId`), not internal state. Both were failing before the fix and pass after — good evidence.

No blocking issues. ✓

---

PR #839, opened 2026-06-23
Fix game card name truncation pushing overflow menu off-screen on mobile

### LINKED ISSUE #838: Game card name doesn't truncate, pushing overflow menu off-screen on mobile
## Goal

Fix the game card so the game name truncates with an ellipsis, always leaving room for the overflow menu.

## Current behaviour

On mobile web, when a game name is long, the overflow menu (⋯) is pushed off-screen because the game name doesn't truncate.

## Expected behaviour

The game name truncates with an ellipsis before reaching the overflow menu, so the menu is always visible and accessible.

## Where

`packages/web/src/components/GameCard.tsx` — the `button` wrapper around `CardTitle` (line 272) lacks the `overflow-hidden` or `min-w-0` constraints needed for the inner `truncate` span to activate correctly.

johnpooch:
Fixes game cards where a long name (or wide inline badge) would push the ⋯ overflow menu off-screen on mobile.

**Changes to `GameCard.tsx`:**
- `flex-1` on the title button so it fills exactly the space left by the actions div — the ⋯ menu is now always visible regardless of title length
- `overflow-hidden` on the button as a safety clip
- `flex-wrap` on `CardTitle` so inline badges (Sandbox / NationBadge) wrap to a second row when they don't fit alongside the title, instead of overflowing into the actions area
- `min-w-0` on the title `<span>` so it can shrink below its text width and trigger `truncate`'s ellipsis

**Behaviour:**
- Short title + small badge → both on the same row, no truncation
- Long title + small badge → both on the same row, title truncates with `…`
- Any title + badge that doesn't fit on one row → badge wraps to a second row, title fills the full row and truncates only if it's itself too long
- ⋯ menu always visible

Closes #838

---

home mobile

### DIFF
--- a/packages/web/src/components/GameCard.tsx
+++ b/packages/web/src/components/GameCard.tsx
@@ -271,9 +271,9 @@ const GameCard: React.FC<GameCardProps> = ({ game, variant, map }) => {
           <div className="flex items-center justify-between gap-2">
             <button
               onClick={handleClickGame}
-              className="text-left hover:underline min-w-0"
+              className="text-left hover:underline min-w-0 overflow-hidden"
             >
-              <CardTitle className="flex items-center gap-2 min-w-0">
+              <CardTitle className="flex flex-wrap items-center gap-2 min-w-0">
                 <span className="truncate min-w-0">{game.name}</span>
                 {game.sandbox && <Badge variant="secondary">Sandbox</Badge>}
                 {!game.sandbox && (isActive || isFinished) && (
--- a/packages/web/src/components/GameCard.tsx
+++ b/packages/web/src/components/GameCard.tsx
@@ -274,7 +274,7 @@ const GameCard: React.FC<GameCardProps> = ({ game, variant, map }) => {
               className="text-left hover:underline min-w-0"
             >
               <CardTitle className="flex items-center gap-2 min-w-0">
-                <span className="truncate">{game.name}</span>
+                <span className="truncate min-w-0">{game.name}</span>
                 {game.sandbox && <Badge variant="secondary">Sandbox</Badge>}
                 {!game.sandbox && (isActive || isFinished) && (
                   <NationBadge nations={variant.nations} nation={playerNation} />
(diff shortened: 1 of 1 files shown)


JorenC:
If there are too many badges, the overflow is still pushed out. 

On mobile, if the length of all badges combined is more than 50% of the screen width, add it to a new row.
If the length of all badges combined is less than 50% of the screen width, truncate the title.

JorenC:
Needs updated screenshots, Claude

johnpooch:
Updated — new screenshot in the PR description reflects the current behaviour. The ⋯ menu is always visible; when a badge doesn't fit alongside the title it wraps to a new row instead of overflowing into the actions area.

---

johnpooch (commented):
**Review: LGTM**

The fix is correct and minimal.

**Why it works:** The ⋯ menu lives in the sibling `<div class="shrink-0">`, so it can never be displaced — the only question is whether the title button behaves correctly. Adding `flex-1` to the button makes it fill exactly the space left by the `shrink-0` div, and adding `min-w-0` to the `<span>` is the critical piece that allows `truncate` to activate (flex items default to `min-width: auto`, which prevents shrinking below text content width). `flex-wrap` on `CardTitle` is a clean fallback for when a badge doesn't fit alongside the title.

Re JorenC's concern about "too many badges still pushing overflow off-screen" — this is fully addressed by the `flex-1`/`shrink-0` split. The ⋯ is in the `shrink-0` div and cannot be displaced regardless of badge count or title length. No further changes needed.

---

johnpooch:
@JorenC I think this fixes the problem described, so we should merge. 

But I think it still looks a bit shitty. I feel like on mobile, maybe all of the badges should appear on a row below the title. The inconsistency looks odd to me.

Suggesting a follow up issue for this. But I'll leave it with you.

Also, Claude needs to create a screenshot where the title is really long so that it causes a ...

JorenC:
Agreed. Let's update this. 

My main concern is the cards take up too much vertical-space, but let's keep it for now. 
In the long run a 'toggle images' thing might remove the variant image.

JorenC:
I think we can close this as it was updated by https://github.com/johnpooch/diplicity-react/pull/846 -  I think I re-assigned 'ready for implementation' thinking it would update. How do I do the update again :D?

PR #842, opened 2026-06-23
Fix inconsistent card heights caused by map aspect ratio differences

JorenC:
Replace md:h-auto on the desktop map panel with a fixed md:h-44 (176px),
matching the classical map's natural height at w-48 (1524×1357 viewBox
gives ~171px). All maps already use cover=true (preserveAspectRatio
xMidYMid slice), so they fill and crop into the fixed dimensions.

Co-Authored-By: Claude Sonnet 4.6 <noreply@anthropic.com>
Claude-Session: https://claude.ai/code/session_01RJpP2L77nLXomF3g3AGwQR

### DIFF
--- a/packages/web/src/components/GameCard.tsx
+++ b/packages/web/src/components/GameCard.tsx
@@ -260,7 +260,7 @@ const GameCard: React.FC<GameCardProps> = ({ game, variant, map }) => {
     <Card className="w-full flex flex-col md:flex-row overflow-hidden p-0">
       <button
         onClick={handleClickGame}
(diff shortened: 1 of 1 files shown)


JorenC:
Eeehm. I asked Claude web to try and make an issue, but i think i forgot to invoke the skill directly. It fixed it immediately rather than requesting the issue :/

Claude, please add screenshots

JorenC:
Claude, please add screenshots when you do the PR review

johnpooch:
For the review workflow to work, the labels need to be attached to the issue, not the PR. But we don't have an issue in this case. I suggest we close this and create the issue.

PR #846, opened 2026-06-23
Fix game card title truncation and move badges below title

### LINKED ISSUE #838: Game card name doesn't truncate, pushing overflow menu off-screen on mobile

johnpooch:
Closes #838

The game name now truncates with an ellipsis when it is too long, keeping the overflow menu always visible. The Sandbox and nation badges have been moved from the inline title row to the badge cluster row below the title.

**Changes:**
- `GameCard.tsx`: Added `overflow-hidden` to the title button and simplified `CardTitle` to use `truncate` directly. Removed Sandbox badge and NationBadge from the title row; added them at the top of `badgeCluster`. Updated the `badgeCluster` render condition to include sandbox games.
- `NationBadge.tsx`: Removed `leading-none` (which was making NationBadge shorter than other badges) and added `max-w-32 truncate` so very long nation names (e.g. "Kingdom of Tarnate") clip with an ellipsis instead of overflowing.

## Screenshots

**Desktop (1280×800)**
my-games desktop

**Mobile (390×844)**
my-games mobile

### DIFF
--- a/packages/web/src/components/GameCard.tsx
+++ b/packages/web/src/components/GameCard.tsx
@@ -217,7 +217,7 @@ const GameCard: React.FC<GameCardProps> = ({ game, variant, map }) => {
       {!game.sandbox && (isActive || isFinished) && (
         <NationBadge nations={variant.nations} nation={playerNation} />
       )}
-      {isActive && currentMember?.eliminated && (
+      {isActive && !game.sandbox && currentMember?.eliminated && (
         <Tooltip>
           <TooltipTrigger asChild>
             <Badge variant="secondary" className="gap-1">
--- a/packages/web/src/components/GameCard.tsx
+++ b/packages/web/src/components/GameCard.tsx
@@ -213,6 +213,10 @@ const GameCard: React.FC<GameCardProps> = ({ game, variant, map }) => {
 
   const badgeCluster = (
     <div className="flex flex-wrap items-center gap-2">
+      {game.sandbox && <Badge variant="secondary">Sandbox</Badge>}
+      {!game.sandbox && (isActive || isFinished) && (
+        <NationBadge nations={variant.nations} nation={playerNation} />
+      )}
       {isActive && currentMember?.eliminated && (
         <Tooltip>
           <TooltipTrigger asChild>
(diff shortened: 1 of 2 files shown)


JorenC:
While this is not what I intended, it does solve the issue that this does not increase vertical-height, so let's continue with this direction. 

Additional guidance:
- All badges need to be equal height, regardless of whether they have icon or not. 
- Badges with very long text (Kingdom of Tarnate) should truncate if the badge becomes so long that it will push out the content out of the window. 
- If there are more badges than fits in the card width, they should continue in a new row, like a flexbox.

johnpooch:
Done — pushed a second commit addressing your three points:

1. **Equal badge heights**: Removed `leading-none` from `NationBadge`. That class was reducing the line-height to match the font-size (12px instead of 16px), making NationBadge ~4px shorter than all other badges in the cluster.
2. **Long text truncation**: Added `max-w-32 truncate` to `NationBadge` so names like "Kingdom of Tarnate" clip with an ellipsis at 128px.
3. **Flex-wrap**: Already in place (`flex flex-wrap items-center gap-2` on the badge cluster).

PR description updated with refreshed screenshots.

---

johnpooch (commented):
## Review

The core truncation fix is correct. Adding `overflow-hidden` to the button is exactly what was needed — the `truncate` class on the inner `<span>` was already there, but without a clipping boundary on the containing button, text could overflow the button's edge and push the ⋯ menu off-screen. Simplifying `CardTitle` to use `truncate` directly (instead of a `flex` wrapper with an inner span) is cleaner.

**Badge relocation** — moving `SandboxBadge` and `NationBadge` out of the title row and into `badgeCluster` is a good call. Inline badges were a second source of overflow and their new position in the cluster row is more consistent. The updated render guard (`game.sandbox || isActive || isFinished`) correctly ensures sandbox games see the cluster.

**`NationBadge.tsx`** — removing `leading-none` (which was shortening badges relative to their siblings) is a fix in its own right. `max-w-32 truncate` correctly prevents overflow for long nation names; `overflow: hidden` (part of `truncate`) does the actual clipping even on an `inline-flex` element.

**Minor observation** — the `!game.sandbox` guard added to the Eliminated badge condition is a reasonable change but isn't described in the issue. It's a behaviour difference: active sandbox games no longer show the Eliminated badge even if `currentMember.eliminated` is true. If that's intentional (sandbox games don't have meaningful elimination state) it's fine, but worth noting explicitly in the PR description or a follow-up so it's traceable.

---

PR #852, opened 2026-06-23
Prevent selecting min reliability higher than user's own tier

### LINKED ISSUE #850: Create game allows selecting min reliability higher than the current user's own tier
## Goal

Prevent users from setting a minimum player reliability requirement that is higher than their own reliability tier.

## Current behaviour

The "Player Reliability" dropdown in the create-game flow shows all three options (`Open`, `Reliable + New Players`, `Reliable only`) regardless of the current user's reliability tier. A user with an "unreliable" or "new" rating can select `Reliable only`, which would exclude themselves from joining their own game.

## Expected behaviour

The control is a stepped slider with three stops (Open → Reliable + New → Reliable Only). The maximum selectable stop is capped at the user's own reliability tier — stops above it are rendered but greyed out and unselectable. The default selection is the highest stop available to the user. Helper text below the slider continues to describe the selected option, as it does now.

## Where

`packages/web/src/screens/Home/CreateGame.tsx` — the `minReliability` `FormField` (around line 909). The option list is defined in `packages/web/src/constants.ts` (`MIN_RELIABILITY_OPTIONS`). The current user's `reliabilityTier` is available on the `UserProfile` type returned by `useUserRetrieveSuspense`.

johnpooch:
The Player Reliability control in the create-game Advanced step is now a stepped button group (Open / Reliable + New / Reliable Only) instead of a dropdown. Stops above the current user's reliability tier are greyed out and unselectable, and the default selection is the highest stop the user is eligible for.

**Backend:** Added `reliabilityTier` to `UserProfileSerializer` so the `/user/` endpoint exposes it. Regenerated OpenAPI schema and TypeScript client.

**Frontend:** Replaced the `<Select>` with an inline button group. `getMaxReliabilityIndex` caps which stops are interactive based on the user's tier (`null` → index 0, `"new"` → index 1, `"reliable"` → index 2). `getDefaultMinReliability` sets the form default to the highest allowed stop.

Note: screenshots could not be taken in this cloud session — the pre-existing Firebase codegen difference causes a page-level JS error when `dev:mocks` runs, which blanks the page.

Closes #850

---
_Generated by [Claude Code](https://claude.ai/code/session_01KsNaKsqLo3FGvS1doVwd7w)_


JorenC:
Duplicate because of 'in-progress' bug - I did not realise it was created (in-progress was hanging)

PR #866, opened 2026-06-24
Resolve fixed-time phases early when all players confirm orders

### LINKED ISSUE #843: Resolve fixed-time phases early when all players confirm orders
## Goal

In fixed-time games, let any phase resolve as soon as every eligible player has confirmed their orders, advancing the game immediately instead of always waiting for the wall-clock deadline. Phase deadlines stay on the fixed schedule, so turns keep their full length and no one is caught out by an unexpectedly early deadline.

## Context

Today fixed-time games never resolve early, so a trivial retreat or adjustment phase — often a single dislodged unit — can hold the whole game for a full cycle while everyone waits. The early-resolution plumbing already exists for duration games; one clause in `PhaseQuerySet.filter_due_phases` (`service/phase/models.py`) excludes fixed-time games, and the confirm-orders button is hidden on them (`OrdersScreen.tsx` returns early when `deadlineMode === "fixed_time"`). First of two issues replacing #775, where the approach was agreed.

## Approach

Relax the fixed-time exclusion in `filter_due_phases` so a phase becomes due once all eligible players have confirmed, exactly as duration games already behave, for every phase type.

When a phase resolves early, anchor the next phase's `scheduled_resolution` to the previous phase's scheduled resolution rather than the moment of early resolution (`Game.get_scheduled_resolution`, `service/game/models.py`), so deadlines stay on the fixed grid and movement turns keep their full length.

Show the confirm-orders button on fixed-time phases like duration games do, with copy that makes clear confirming advances the game early while the deadline itself stays on the fixed schedule. Settle the wording while implementing.

johnpooch:
Closes #843

## What this does

In fixed-time games, phases now resolve immediately once all eligible players confirm their orders, rather than always waiting for the wall-clock deadline.

Three backend changes:

1. **`filter_due_phases`** — removed the `~Q(game__deadline_mode=DeadlineMode.FIXED_TIME)` exclusion from the `all_confirmed` branch, so fixed-time phases are eligible for early resolution when everyone has confirmed.

2. **`get_scheduled_resolution`** — added a `reference_time` parameter that is forwarded to `calculate_next_fixed_deadline`. This lets callers anchor the next deadline to the previous scheduled deadline rather than `now()`.

3. **`create_from_adjudication_data`** — passes `reference_time=previous_phase.scheduled_resolution` when computing the next phase's deadline, so an early resolution does not shift the game's fixed deadline grid forward.

One frontend change:

4. **`OrdersScreen`** — removed the early return that hid the confirm-orders button for fixed-time games. Fixed-time games now display the same "Confirm orders" button as duration games.

## Behaviour change: NMR extension criterion for fixed-time games

The NMR-extension logic for fixed-time games previously used "has submitted at least one order" as the criterion for skipping an extension. This PR unifies the logic across all deadline modes so that the criterion is now "has confirmed orders".

**Impact on in-flight games:** Players in active fixed-time games who have submitted orders but have never confirmed (the confirm button was hidden before this PR) will be treated as unconfirmed after deploy. If the NMR sweep runs before they confirm, they may receive an extension. This is a one-time transition effect; once they see and use the confirm button the behaviour stabilises.

## Screenshot

The confirm-orders button (previously hidden for fixed-time games):

Confirm orders button

---
_Generated by [Claude Code](https://claude.ai/code/session_0115PCApfMhMqA7iyK9dkkKk)_

### DIFF
--- a/service/phase/models.py
+++ b/service/phase/models.py
@@ -200,16 +200,9 @@ class PhaseManager(models.Manager):
         return self.resolve_due_phases(canary=True)
 
     def _check_and_apply_nmr_extensions(self, phase):
-        if phase.game.deadline_mode == DeadlineMode.FIXED_TIME:
-            unconfirmed = phase.phase_states.filter(
-                has_possible_orders=True,
-            ).annotate(order_count=Count('orders')).filter(
-                order_count=0,
-            ).select_related('member')
-        else:
-            unconfirmed = phase.phase_states.filter(
-                has_possible_orders=True, orders_confirmed=False
-            ).select_related('member')
+        unconfirmed = phase.phase_states.filter(
+            has_possible_orders=True, orders_confirmed=False
+        ).select_related('member')
 
         members_with_extensions = [
             ps.member for ps in unconfirmed
--- a/service/phase/models.py
+++ b/service/phase/models.py
@@ -85,7 +85,7 @@ class PhaseQuerySet(models.QuerySet):
             & Q(game__paused_at__isnull=True)
             & ~Q(game__status=GameStatus.COMPLETED)
             & ~Q(game__status=GameStatus.ABANDONED)
-            & (deadline_passed | (~Q(game__deadline_mode=DeadlineMode.FIXED_TIME) & all_confirmed))
+            & (deadline_passed | all_confirmed)
         )
 
 
@@ -674,7 +674,8 @@ class PhaseManager(models.Manager):
 
                 # Calculate next phase details
                 scheduled_resolution = previous_phase.game.get_scheduled_resolution(
-                    adjudication_data["type"]
+                    adjudication_data["type"],
+                    reference_time=previous_phase.scheduled_resolution,
                 )
                 new_ordinal = previous_phase.ordinal + 1
 
--- a/service/phase/tests.py
+++ b/service/phase/tests.py
@@ -4515,7 +4515,7 @@ class TestNMRExtensionsFixedTime:
         assert phase.scheduled_resolution > now + timedelta(hours=24)
 
     @pytest.mark.django_db
-    def test_fixed_time_with_any_order_skips_extension(
+    def test_fixed_time_confirmed_skips_extension(
         self,
         deadline_warning_game_factory,
         italy_vs_germany_italy_nation,
@@ -4525,6 +4525,10 @@ class TestNMRExtensionsFixedTime:
         game, italy, germany, phase = deadline_warning_game_factory(
             DeadlineMode.FIXED_TIME, now - timedelta(minutes=1)
         )
+        game.movement_frequency = PhaseFrequency.EVERY_2_DAYS
+        game.fixed_deadline_time = time(12, 0)
+        game.fixed_deadline_timezone = "UTC"
+        game.save()
         italy.nmr_extensions_remaining = 1
         italy.save()
         phase.units.create(
@@ -4532,8 +4536,7 @@ class TestNMRExtensionsFixedTime:
             type=UnitType.ARMY,
             nation=italy_vs_germany_italy_nation,
         )
-        ps = phase.phase_states.create(member=italy, has_possible_orders=True)
-        ps.orders.create(source=italy_vs_germany_venice_province, order_type=OrderType.HOLD)
+        phase.phase_states.create(member=italy, has_possible_orders=True, orders_confirmed=True)
 
         result = Phase.objects._check_and_apply_nmr_extensions(phase)
 
--- a/service/phase/tests.py
+++ b/service/phase/tests.py
@@ -2029,7 +2029,7 @@ class TestFilterDuePhasesBasicFiltering:
         assert phase in phases
 
     @pytest.mark.django_db
-    def test_filter_due_phases_fixed_time_all_confirmed_does_not_resolve_early(
+    def test_filter_due_phases_fixed_time_all_confirmed_resolves_early(
         self, phase_factory, classical_variant, classical_england_nation, classical_france_nation
     ):
         game = Game.objects.create(
@@ -2049,7 +2049,7 @@ class TestFilterDuePhasesBasicFiltering:
 
         phases = Phase.objects.filter_due_phases()
 
-        assert phase not in phases
+        assert phase in phases
 
     @pytest.mark.django_db
     def test_filter_due_phases_fixed_time_resolves_at_scheduled_time(
--- a/service/phase/tests.py
+++ b/service/phase/tests.py
@@ -4119,7 +4119,7 @@ class TestPhaseToCanonicalGameStatePerformance:
 class TestSendDeadlineWarnings:
 
     @pytest.mark.django_db
-    def test_fixed_time_all_orders_no_notification(
+    def test_fixed_time_all_orders_not_confirmed_sends_confirm_prompt(
         self,
         deadline_warning_game_factory,
         add_italy_germany_units,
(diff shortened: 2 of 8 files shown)


johnpooch (commented):
## Review

The overall approach is correct — removing the fixed-time exclusion from `all_confirmed` and anchoring the next deadline to `previous_phase.scheduled_resolution` rather than `now()` are the right two moves. Two findings below.

---

### 1. No test for deadline grid preservation

The most important correctness guarantee of this PR — that an early resolution does not shift the fixed deadline grid — is untested. The `reference_time=previous_phase.scheduled_resolution` parameter in `create_from_adjudication_data` is what enforces this, but no test exercises it.

A minimal case: game has daily 21:00 UTC deadlines; phase 1 scheduled for Jan 2 21:00; resolves early at Jan 2 18:00; phase 2 should land on **Jan 3 21:00**, not Jan 2 21:00 (which would be in the past) or some time derived from the actual resolution time.

`service/phase/test_background_resolution.py` is the right place for this. Without it the grid-preservation logic is uncheckable without running the full system.

---

### 2. `_check_and_apply_nmr_extensions` can silently block an early resolution

`PhaseManager.resolve()` (`service/phase/models.py`) calls `_check_and_apply_nmr_extensions` before doing anything else and returns early if any extension is applied. For fixed-time games the extension check uses `order_count=0`, not `orders_confirmed`:

```python
# service/phase/models.py _check_and_apply_nmr_extensions
if phase.game.deadline_mode == DeadlineMode.FIXED_TIME:
    unconfirmed = phase.phase_states.filter(
        has_possible_orders=True,
    ).annotate(order_count=Count('orders')).filter(order_count=0, ...)
```

So if a player confirms with zero orders submitted (possible via the frontend — the confirm button is shown whenever the player has orderable provinces, regardless of order count), the phase can satisfy `all_confirmed` in `filter_due_phases`, enter early resolution, but then be blocked by `_check_and_apply_nmr_extensions` and extended instead.

Before this PR the two code paths never intersected for fixed-time games. Now they can. Whether "extend despite confirmation" is the right policy is a design question, but the current code produces this silently and without surfacing it to the player.


---

johnpooch (commented):
**Finding 1 — addressed.** Added `TestDeadlineGridPreservation::test_next_deadline_anchored_to_scheduled_resolution_not_now` in `phase/test_background_resolution.py`. It creates a daily 21:00 UTC game, calls `get_scheduled_resolution("Movement", reference_time=datetime(2026-01-03 21:00 UTC))`, and asserts the result is `2026-01-04 21:00 UTC` — then asserts that omitting `reference_time` (using `now()`) produces a different date, proving the grid-anchoring is what drives the result.

**Finding 2 — design question, flagging for human decision.** You're right that the two paths can now intersect for fixed-time games: a player who confirms but submits no orders would satisfy `all_confirmed` in `filter_due_phases` while simultaneously hitting the `order_count=0` branch in `_check_and_apply_nmr_extensions`, causing the phase to be extended despite everyone confirming. Before this PR this could never happen because fixed-time games were excluded from early resolution entirely.

The question is what "confirm with no orders" should mean for NMR purposes in a fixed-time game. Two plausible policies:
- **Confirmed = done**: switch `_check_and_apply_nmr_extensions` to use `orders_confirmed=False` for fixed-time games (same as duration), treating a confirmation as "I'm done, no orders intended".
- **Orders = activity**: keep `order_count=0` as the NMR signal and add a frontend guard (disable the confirm button or show a warning when orderable provinces exist but no orders are submitted).

I'd lean toward the first (match duration-game semantics — confirming is the signal), but this touches the NMR extension policy and I want your call before changing it. Happy to implement whichever direction you choose, or to open a follow-up issue if you'd prefer to keep this PR focused.

---

JorenC:
Confirmed = done. If a user confirms but adds no orders, it's a signal that they are not intending to make any moves.

johnpooch:
The issue descriptions specifies that there should be user facing copy describing what pressing confirm orders does. 

Can we confirm whether this exists? What do the notifications currently say when the fixed time game resolves early

johnpooch:
I think the problem with testing confirm as done is that we auto confirm cd players I think. Can we confirm whether this is the case? Will this break how CD members are handled?

johnpooch (commented):
**Copy — now added.** Two changes pushed:

1. **`PhaseGuidance.tsx`**: Removed `game.deadlineMode !== "fixed_time"` from `confirmationApplies`. Fixed-time games now show the same "confirmed / not confirmed" status in the guidance bar as duration games.

2. **`build_notification_body`**: When a fixed-time player has all orders ready but hasn't confirmed, the deadline-approaching notification now reads: *"All orders ready. Confirm to advance the game early — the next deadline stays on schedule. X remaining."*

**CD auto-confirmation — no breakage.** CD players are auto-confirmed when the new phase is created (line 792: `orders_confirmed=member.civil_disorder`). Our NMR change (`orders_confirmed=False` instead of `order_count=0`) actually *improves* CD handling: the old check would have consumed a CD player's extension when they had no orders (because `order_count=0`), even though they had no agency. Now CD players are already confirmed and correctly excluded from the NMR check.

---

JorenC:
@johnpooch this looks good to me.

service/phase/models.py
```diff
@@ -200,16 +200,9 @@ def sweep_due_phases(self):
...
-            ).annotate(order_count=Count('orders')).filter(
-                order_count=0,
-            ).select_related('member')
-        else:
-            unconfirmed = phase.phase_states.filter(
-                has_possible_orders=True, orders_confirmed=False
-            ).select_related('member')
+        unconfirmed = phase.phase_states.filter(
+            has_possible_orders=True, orders_confirmed=False
+        ).select_related('member')
```
johnpooch, line 206: The NMR-extension criterion for fixed-time games silently changes from "has no orders" to "not confirmed". Before this PR, submitting any order was enough to avoid an extension. After, a player must confirm.

For in-flight fixed-time games this is a one-time surprise: players who currently have partial orders but have never seen the confirm button (because it was hidden) will immediately qualify as "unconfirmed" and may have their deadline extended on the next NMR sweep. The PR description doesn't mention this behavioural change — worth capturing explicitly so reviewers and the release note can account for it.

---

service/phase/tests.py
```diff
@@ -4525,15 +4548,18 @@ def test_fixed_time_with_any_order_skips_extension(
         game, italy, germany, phase = deadline_warning_game_factory(
             DeadlineMode.FIXED_TIME, now - timedelta(minutes=1)
         )
+        game.movement_frequency = PhaseFrequency.EVERY_2_DAYS
+        game.fixed_deadline_time = time(12, 0)
+        game.fixed_deadline_timezone = "UTC"
+        game.save()
```
johnpooch: These four lines are dead weight. Since Italy's phase state has `orders_confirmed=True`, `unconfirmed` is empty and `_check_and_apply_nmr_extensions` returns before it ever touches the game's deadline fields — so `movement_frequency`, `fixed_deadline_time`, `fixed_deadline_timezone`, and the `.save()` are never read. Drop them to keep the test focused on the one thing it's asserting.

---

johnpooch (commented):
Overall the approach is sound — the three backend changes work together correctly, the frontend change is the right single-line removal, and `calculate_next_fixed_deadline` already had `reference_time` support so the deadline-grid preservation works without any hidden changes.

Two findings to address before merging:

1. **NMR extension behaviour change** (inline): the criterion change from "has any order" → "has confirmed" affects in-flight fixed-time games in a way that isn't described anywhere. Doesn't block the feature, but should be called out explicitly.

2. **Dead test setup** (inline): the `movement_frequency` / `fixed_deadline_time` / `fixed_deadline_timezone` lines in `test_fixed_time_confirmed_skips_extension` are never exercised.

---

johnpooch:
The thing I'm still not sure about is:

When a player goes into civil disorder, we ensure that they don't slow down phases. Effectively we assume that they have already confirm so the phase resolves if all players except for the CD player have confirmed. 

I can't remember how this works. Do we explicitly set "has confirmed" to true on CD players, or do we exclude cd players from the relevant queries

johnpooch:
CD players are handled via explicit `orders_confirmed=True` at phase state creation — not by excluding them from queries.

At `phase/models.py:792`:
```python
orders_confirmed=member.civil_disorder,
```

So every CD player starts each new phase already confirmed. The `all_confirmed` subquery (`filter_due_phases`, line 80) looks for `orders_confirmed=False`, so CD players never appear in it and don't block early resolution. The same applies to `_check_and_apply_nmr_extensions` — it also filters on `orders_confirmed=False`, so CD players are never candidates for an NMR extension.

The change in this PR doesn't affect CD player behaviour — they were auto-confirmed before and remain auto-confirmed after.

---

PR #870, opened 2026-06-24
Fix toast action button native appearance on Chrome/Linux

### LINKED ISSUE #869: Toast action button does not match success notification style
## Goal

Fix the action button inside the "A new phase has started" success toast so it is visually consistent with the notification's style and clearly pressable.

## Current behaviour

The "Jump to newest" action button renders as unstyled plain text inside the green success toast, with no border, background, or other affordance to indicate it is interactive.

## Expected behaviour

The action button should adapt to the success notification's colour scheme — for example, a dark green border and appropriate contrast — so it reads as a distinct, pressable element within the toast.

## Where

`packages/web/src/components/ui/sonner.tsx` — the `Toaster` component only defines `--normal-*` CSS variables; no `--success-*` action-button variables are set. The toast is triggered in `packages/web/src/components/PhaseSelect.tsx:63`.

johnpooch:
Fixes the action button rendering with native GTK browser styling on Chrome/Linux instead of Sonner's intended dark button.

Tailwind v4's preflight no longer strips `appearance: button` from form controls (unlike v3), so on Chrome/Linux the GTK button style renders over Sonner's CSS-specified `background` and `border`. Adding `appearance: none` to `[data-sonner-toast] [data-button]` removes the native rendering and lets Sonner's own styles take full effect.

Closes #869

## After

after

### DIFF
--- a/packages/web/src/index.css
+++ b/packages/web/src/index.css
@@ -152,27 +152,11 @@ body {
   }
 }
 
-[data-rich-colors='true'][data-sonner-toast][data-type='success'] [data-button] {
-  background: transparent;
-  border: 1px solid var(--success-text);
-  color: var(--success-text);
-}
-
-[data-rich-colors='true'][data-sonner-toast][data-type='error'] [data-button] {
-  background: transparent;
-  border: 1px solid var(--error-text);
-  color: var(--error-text);
-}
-
-[data-rich-colors='true'][data-sonner-toast][data-type='warning'] [data-button] {
-  background: transparent;
-  border: 1px solid var(--warning-text);
-  color: var(--warning-text);
-}
-
-[data-rich-colors='true'][data-sonner-toast][data-type='info'] [data-button] {
-  background: transparent;
-  border: 1px solid var(--info-text);
-  color: var(--info-text);
+/* Tailwind v4 preflight no longer strips native button appearance, so on
+   Chrome/Linux the GTK button style overrides Sonner's background/border.
+   Force appearance: none so Sonner's own CSS takes full effect. */
+[data-sonner-toast] [data-button] {
+  -webkit-appearance: none;
+  appearance: none;
 }
 
--- a/packages/web/src/index.css
+++ b/packages/web/src/index.css
@@ -151,3 +151,28 @@ body {
     @apply bg-background text-foreground;
   }
 }
+
+[data-rich-colors='true'][data-sonner-toast][data-type='success'] [data-button] {
+  background: transparent;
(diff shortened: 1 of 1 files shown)


JorenC:
Actually, the bug might be different. The black button works well and is good as canonical design, and should be implemented as such.

The bug is not that the black button is incorrect - the bug is that the black button does not display correctly. If you look at this (rendered on Chrome/Linux) you can see the button follows a native styling, rather than the correct black styling.

Screenshot_20260624-122004

johnpooch:
Thanks for clarifying — I had the wrong mental model of the bug. I've pushed a corrected fix.

The root cause: Tailwind v4's preflight no longer sets `appearance: none` on button elements (a change from v3), so on Chrome/Linux the browser applies its native GTK button appearance over Sonner's CSS-specified `background`/`border`. The fix is a single CSS rule in `index.css`:

```css
[data-sonner-toast] [data-button] {
  -webkit-appearance: none;
  appearance: none;
}
```

This removes the native rendering and lets Sonner's own dark-button styles take effect. Screenshot of the result (taken on this Linux environment):

after

---

johnpooch (commented):
Review (self-review, cannot self-approve):

Clean fix. The selector is correctly scoped to `[data-sonner-toast] [data-button]` so it doesn't affect any other buttons in the app, and `appearance: none` is the right tool — it removes the GTK override without touching Sonner's own styling logic.

The approach is also better than the alternative suggested in the issue (adding `--success-*` CSS variables): that would only have affected success toasts, whereas this fixes the native-appearance issue across all toast types on Chrome/Linux.

The comment in the CSS is justified — the Tailwind v4 preflight behaviour change affecting GTK button rendering is a non-obvious platform-specific quirk that warrants explanation. This is the documented exception to the no-comments rule.

No issues found. Ready to merge.

---

johnpooch (commented):
**Review: no findings.**

- `appearance: none` is the right fix — removes GTK native rendering so Sonner's CSS takes full effect; `-webkit-appearance: none` covers older WebKit.
- Selector `[data-sonner-toast] [data-button]` is tightly scoped with no unintended side effects.
- Comment is warranted: cross-OS, Tailwind-v4-specific workaround that would surprise a reader without it.
- Also fixes the missing end-of-file newline.

Ready to merge.

---

JorenC:
This still has a problem - however, it works on my phone, and it also gives the error on my linux/firefox. 

I think this might be a too edge case than I thought. Let's close this and leave-it-as-is?

PR #876, opened 2026-06-24
Show minReliability on Game Info screen

### LINKED ISSUE #873: Game Info screen does not show the minimum reliability requirement
## Goal

Surface a game's minimum reliability requirement on the Game Info screen.

## Current behaviour

Games can be created with a minimum reliability threshold (`minReliability`), but the Game Info screen does not display it, even though the value is available on the game retrieve response.

## Expected behaviour

The Game Info screen shows the game's minimum reliability requirement alongside the other settings (Visibility, Press type, NMR extensions, etc.).

## Where

`packages/web/src/components/GameInfoContent.tsx` renders the settings rows; `minReliability` is already present on the game retrieve type but isn't rendered there.

johnpooch:
Closes #873

Surfaces the game's minimum reliability requirement on the Game Info screen, alongside the other settings (Visibility, Press type, NMR extensions).

Also fixes the pending-game action when a non-member cannot join due to the game's reliability requirement: instead of showing the Leave button (which would be incorrect for a non-member), a disabled "Join game" button is shown with the message "Your reliability is too low to join this game". The existing Leave + Share action is now gated on `canLeave` rather than `!canJoin`.

## Screenshots

**Desktop (1280×800)**
Game Info screen — Player reliability row

**Mobile (390×844)**
Game Info screen — mobile

### DIFF
--- a/packages/web/package-lock.json
+++ b/packages/web/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "diplicity-react",
-  "version": "1.5.2",
+  "version": "1.5.3",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "diplicity-react",
-      "version": "1.5.2",
+      "version": "1.5.3",
       "dependencies": {
         "@capacitor-firebase/messaging": "^8.1.0",
         "@capacitor/android": "^8.3.4",
--- a/packages/web/src/components/GameInfoContent.tsx
+++ b/packages/web/src/components/GameInfoContent.tsx
@@ -1,5 +1,5 @@
 import React from "react";
-import { BookOpen, Calendar, Clock, Users, Lock, Unlock, User, Map, Trophy, Pause, Shield, MessageSquare, MessageSquareOff } from "lucide-react";
+import { BookOpen, Calendar, Clock, Users, Lock, Unlock, User, Map, Trophy, Pause, Shield, ShieldCheck, MessageSquare, MessageSquareOff } from "lucide-react";
 
 import { Skeleton } from "@/components/ui/skeleton";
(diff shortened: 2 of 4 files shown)


johnpooch (commented):
## Review

The implementation is correct and well-structured. Two observations — one a design question, one a genuine improvement the PR makes over the old test.

---

**`GameInfoContent.tsx` — "Player reliability" is always rendered**

```tsx
<MetadataRow
  icon={<ShieldCheck className="size-4" />}
  label="Player reliability"
  value={
    MIN_RELIABILITY_OPTIONS.find(o => o.value === game.minReliability)?.label ??
    game.minReliability
  }
/>
```

This is unconditional — every game info page will show "Player reliability: Open" by default. `nmrExtensionsAllowed` uses a conditional (`{game.nmrExtensionsAllowed > 0 && ...}`), but Visibility and Press type are also always shown, so the inconsistency is pre-existing. Always showing the row is defensible (it makes the game's entry requirements explicit at a glance) and consistent with surfacing Visibility unconditionally. Noting it as a design observation rather than a request to change.

---

**`GameInfo.test.tsx` — `pendingGameCanLeave` fixture fix**

```ts
// Before
const pendingGameCanLeave = mockPendingGames.find((g) => !g.canJoin)!;
// After
const pendingGameCanLeave = mockPendingGames.find((g) => !g.canJoin && g.canLeave)!;
```

Good catch. The old selector would have matched any game where `canJoin` was false — including the new reliability-blocked scenario where `canLeave` is also false. Adding `&& g.canLeave` makes the test honest about what it's covering.

---

No correctness issues. The `canJoin ? … : canLeave ? … : minReliability !== "open" ? … : null` chain handles all cases correctly. LGTM.

---

JorenC:
1) there are no screenshots.
2) all games always showing reliability (default:open) is intended behaviour.

PR #886, opened 2026-06-24
Scope My Games board prefetch to the current phase

johnpooch:
## What this does

Fixes a performance regression on the `GET /games/` endpoint (the My Games screen) introduced by #857.

#857 added `units`/`supply_centers` to the games-list response so the map preview reflects live board state, prefetching them via `units__…`/`supply_centers__…` on the **phases** prefetch. Because that prefetch loads *all* phases of every game, it hydrated the board for **every historical phase** even though only the current phase is ever serialized. On games with long histories this pulls tens of thousands of unit/SC rows (plus their nation-flag and province joins) per request.

The impact is visible in production Honeycomb: `GET /games/` **p50 roughly tripled (~135ms → ~390ms)** and p95/p99 roughly doubled, with a clean step-change in the first hour after #857 merged, sustained since.

## The fix

Scope the units/supply-centre prefetches to the **current phase** (latest ordinal) via a correlated subquery, so hydration is **O(games)** instead of **O(games × phases)**. The lightweight all-phases prefetch (needed for `current_phase` derivation, `phase_confirmed`, `order_status`) is unchanged, as is the serialized response shape — so **no codegen or frontend changes**.

Games-list query count drops from **15 → 9**.

## Tests

- **`test_list_games_hydrates_units_for_current_phase_only`** — regression test asserting only the current phase's board is hydrated regardless of history depth. It fails against the previous prefetch (`48 != 6` for a 2-game × 8-phase fixture) and passes with this fix. The waste was internal hydration (the JSON response was identical before/after), so this test evaluates the view's queryset and asserts the prefetch footprint, with `django_assert_num_queries(0)` confirming no N+1.
- **`test_list_games_current_phase_includes_board`** — new functional coverage (#857 shipped without it) asserting the list response still carries the current phase's units/supply centres, and only the current phase's (a FLEET in the active phase, not the ARMY in the prior completed phase).
- Updated the four existing query-count assertions 15 → 9.
- Full game app suite green (307 passed).

## Follow-ups (not in this PR)

- **Lean preview serializer**: the map preview only consumes `{province_id, nation_name, type, dislodged}`, but the full `UnitSerializer`/`SupplyCenterSerializer` (nation-flag URL, province parent, named coasts) is still loaded and serialized. A preview-specific serializer would let us drop those joins and trim query count further. Deferred because it touches the API contract (codegen + frontend types).
- **Denormalise board state at phase resolution** if list latency is still a hotspot at scale.

No visual changes (backend only).

---
_Generated by [Claude Code](https://claude.ai/code/session_015vzjSckZXj3CUwbMbXjMKV)_

### DIFF
--- a/service/game/models.py
+++ b/service/game/models.py
@@ -89,16 +89,29 @@ class GameQuerySet(models.QuerySet):
                 order_count=Count("orders")
             ),
         )
+        current_phase_id = (
+            Phase.objects.filter(game=OuterRef("phase__game_id"))
+            .order_by("-ordinal")
+            .values("id")[:1]
+        )
+        current_units_prefetch = Prefetch(
+            "units",
+            queryset=Unit.objects.filter(phase=Subquery(current_phase_id))
+            .select_related("nation__flag", "province__parent")
+            .prefetch_related("province__named_coasts"),
+        )
+        current_supply_centers_prefetch = Prefetch(
+            "supply_centers",
+            queryset=SupplyCenter.objects.filter(phase=Subquery(current_phase_id))
+            .select_related("nation__flag", "province__parent")
+            .prefetch_related("province__named_coasts"),
+        )
         phases_prefetch = Prefetch(
             "phases",
             queryset=Phase.objects.prefetch_related(
                 phase_states_prefetch,
-                "units__nation__flag",
-                "units__province__parent",
-                "units__province__named_coasts",
-                "supply_centers__nation__flag",
-                "supply_centers__province__parent",
-                "supply_centers__province__named_coasts",
+                current_units_prefetch,
+                current_supply_centers_prefetch,
             ),
         )
 
--- a/service/game/tests.py
+++ b/service/game/tests.py
@@ -437,6 +437,59 @@ class TestGameListView:
                 assert field in game
 
     @pytest.mark.django_db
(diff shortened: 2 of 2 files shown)


johnpooch (commented):
### PR Review

Scopes the units/supply-centre prefetch on `GET /games/` to the current (max-ordinal) phase via a correlated subquery, instead of hydrating the board for every historical phase. The change is correct and will improve performance: only `current_phase` is ever serialized (`get_current_phase` in `serializers.py`), so the narrower prefetch leaves the response shape identical while cutting hydration from O(games × phases) to O(games); query count drops 15 → 9 thanks to folding the nation/province joins into `select_related`. The subquery's `OuterRef("phase__game_id")` and `-ordinal` ordering match the `current_phase` derivation (`phases[-1]` under `Meta.ordering = ["ordinal"]`), and the empty-history case degrades safely to no units. Verified the field paths, serializer usage, and ordering against the codebase.

One minor, non-blocking note: there's no composite `(game, ordinal)` index on `Phase`, so the correlated subquery does a per-game in-memory sort — negligible given the small phase count per game, but worth keeping in mind if list latency resurfaces at scale.

Well covered by the new regression test (`...current_phase_only`) and functional test. No issues worth blocking.

---

PR #887, opened 2026-06-24
Fix map preview not filling card height for wide-aspect-ratio variants

### LINKED ISSUE #882: Game card map preview leaves empty space below short maps
## Goal

Make the game card map preview fill the full height of its panel for variants whose map is shorter (wider aspect ratio) than the classical map.

## Current behaviour

PR #854 fixed cards growing taller for taller maps (e.g. Hundred) by setting a fixed `md:h-44` on the map panel and relying on `preserveAspectRatio="xMidYMid slice"` cover mode. For a variant whose map is shorter/wider than classical (e.g. the Canton/Asia map), the preview does not fill the panel height and leaves empty space below the map. This is visible on the Home / My Games game cards.

Image

## Expected behaviour

The map preview fills the full height of the panel for every variant, with no empty space below the map, just as the classical map does.

## Where

`packages/web/src/components/GameCard.tsx` (the map panel button height) and `packages/web/src/components/MapPreview.tsx` (the cover / `preserveAspectRatio` rendering).

johnpooch:
Closes #882

## What changed

`MapPreview` with `cover=true` was leaving empty space below the map for variants whose SVG has a wider aspect ratio than the game card container (e.g. Canton: 2081×1500 ≈ 1.39:1 vs container 192×176 ≈ 1.09:1).

**Root cause**: SVG `height="100%"` presentation attributes don't reliably override an SVG's intrinsic aspect ratio in all browsers. When the browser couldn't resolve the height, it fell back to the intrinsic ratio — rendering Canton at ~138px tall inside a 176px card, leaving a ~38px gap at the bottom.

**Fix**: In cover mode, the inner div (SVG wrapper) now uses `absolute inset-0` instead of `w-full h-full`. Height from absolute positioning constraints is always resolved concretely (never falls back to intrinsic ratio). The outer div gets `relative overflow-hidden` to scope the positioning. The SVG's `preserveAspectRatio="xMidYMid slice"` continues to handle the content scaling.

## Screenshots

**Variants list (all variant maps fill their containers):**

variants list

**Card gallery (desktop):**

card gallery desktop

**Card gallery (mobile):**

card gallery mobile

---
_Generated by [Claude Code](https://claude.ai/code/session_01XaPMW1SFyPC98CHobFYDKJ)_

### DIFF
--- a/packages/web/src/components/MapPreview.tsx
+++ b/packages/web/src/components/MapPreview.tsx
@@ -70,9 +70,16 @@ const MapPreview: React.FC<MapPreviewProps> = ({
   }, [dsvg, variant, phase, cover]);
 
   return (
-    <div ref={containerRef} className={className} style={style}>
+    <div
+      ref={containerRef}
+      className={cover ? `relative overflow-hidden ${className ?? ""}` : className}
+      style={style}
+    >
       {svg ? (
-        <div className="w-full h-full" dangerouslySetInnerHTML={{ __html: svg }} />
+        <div
+          className={cover ? "absolute inset-0" : "w-full h-full"}
+          dangerouslySetInnerHTML={{ __html: svg }}
+        />
       ) : (
         <Skeleton className="w-full h-full" />
       )}
--- a/packages/web/src/components/MapPreview.tsx
+++ b/packages/web/src/components/MapPreview.tsx
@@ -81,7 +81,7 @@ const MapPreview: React.FC<MapPreviewProps> = ({
           dangerouslySetInnerHTML={{ __html: svg }}
         />
       ) : (
-        <Skeleton className="w-full h-full" />
+        <Skeleton className={cover ? "absolute inset-0" : "w-full h-full"} />
       )}
     </div>
   );


johnpooch (commented):
Clean, minimal fix — the root-cause analysis is correct and the change is well-scoped.

**One nit:** in cover mode the `<Skeleton>` still uses `w-full h-full` while the loaded SVG div uses `absolute inset-0`. The outer container always has concrete dimensions (every call site passes `className="w-full h-full"` or equivalent), so this is functionally fine — but the loading and loaded states use inconsistent positioning strategies inside the same branch. Aligning the skeleton to `absolute inset-0` in cover mode would make the two states symmetric:

```tsx
) : (
  <Skeleton className={cover ? "absolute inset-0" : "w-full h-full"} />
)}
```

Not a blocker — the current code renders correctly.

---

johnpooch:
Good catch — applied in cd5e4db.

---

johnpooch:
image

Still appearing. I think I misdiagnosed the issue. It's actually a problem because the content of the card is greater than the height of the classical map. 

I think the better fix is to just reduce the height of the card content to fit within the expected height. 

Closing this PR and the related issue.

PR #889, opened 2026-06-25
Add Leaflet map renderer behind MapView (Step 2 of map re-architecture)

johnpooch:
## What this does

Step 2 of the measurement-first map re-architecture: a **new map renderer behind the already-frozen `MapView` interface** (#883), using Leaflet `CRS.Simple` and the proposal's three-layer model. This is opt-in only — **the default stays the existing SVG map**; flipping the default is Phase 2.

The core invariant from the [proposal](https://github.com/johnpooch/diplicity-react/discussions/788) is "rasterise the static base once, move only a camera + light overlay per frame". The new module implements exactly that:

| Layer | Source | Rebuilt when |
|---|---|---|
| **Base (visual)** | full-detail variant SVG **rasterised once** to an `L.imageOverlay` at native viewBox resolution (the accepted "soft at 4× zoom" trade) | phase / ownership change |
| **Hit-test** | **decimated** province polygons (Ramer–Douglas–Peucker, ~5px) built once per variant; drives click → `{x,y}`, hover, `selected`/`highlighted`/`renderableProvinces` | once per variant |
| **Overlay** | units + order arrows, reusing the existing renderer's output as a light `L.svgOverlay` | per interaction / phase |

`orderGeometry.ts` and the order/unit paint code are reused unchanged, so the overlay is visually identical to the SVG map.

## How to try it

Add `?map=leaflet` to any game route (also `?map=canvas`, or `localStorage["map:impl"]="leaflet"`). The new module is **code-split** (`GameMapCanvas` chunk, ~159 kB incl. Leaflet) and only loads when selected — the default bundle is unchanged.

## Telemetry parity

`recordInitialRender` / `recordGesture` now carry a **`map.implementation`** attribute (`"svg"` | `"leaflet"`); both implementations set it, so Honeycomb can split the Block 2 distributions when comparing.

## Performance (Block 1 harness, `--map leaflet`)

`map-bench.mjs` gained a `--map` flag that targets the Leaflet DOM. Same machine, headless Chromium under CDP CPU throttle, classical `active-movement`. **p95 frame time (ms), dropped-frame %:**

| viewport | cpu | SVG (p95 / dropped) | Leaflet (p95 / dropped) |
|---|---|---|---|
| 1280×800 | 1× | 16.8 / 2.6% | **16.7 / 0%** |
| 1280×800 | 4× | 66.7 / 14.8% | **16.7 / 0.2%** |
| 1280×800 | 6× | 108.4 / 16.5% | **16.8 / 3.8%** |
| 390×844 | 1× | 16.8 / 0.2% | **16.7 / 0%** |
| 390×844 | 4× | 41.7 / 13.8% | **16.8 / 2.3%** |
| 390×844 | 6× | 58.4 / 11.9% | **16.8 / 3.8%** |

The new map holds p95 ≤ 16.8ms at every tier — meeting the proposal's `p95 < 16ms` budget where the SVG map blows through it 4–7×. The trade is a one-time base rasterisation (~0.4–0.7s unthrottled, longer under heavy throttle), as designed. Committed as `packages/map/bench/results-leaflet.json`.

## Screenshots

**New map (Leaflet) — Spring 1901 movement, with order arrows:**
leaflet movement

**Existing SVG map, same fixture (for parity):**
svg movement

**Build phase (ownership fills) and retreat phase (dislodged unit + retreat flag):**
leaflet build
leaflet retreat

**Mobile (390×844):**

## Tests

New unit tests cover the pure pieces: RDP decimation, SVG path flattening, province-ring building, base/overlay SVG splitting, the dev-switch resolver, and the `map.implementation` telemetry tag. `npm run build`, `npm run lint`, and the new vitest files all pass. Click wiring was verified end-to-end under mocks (a click resolves the exact province id and emits `{x,y}`, identical to the SVG map; the order menu itself is inert under mocks because `GET options` is empty, as documented).

## Out of scope (Phase 2 / later)

Flipping the default, removing the legacy SVG map, and migrating the other consumers (`MapPreview`, `ExpandableMapPreview`, `TutorialMap`).

---
_Generated by [Claude Code](https://claude.ai/code/session_01Mg5uiZ85jPvmRcPEnyMQfz)_


johnpooch:
@JorenC this is very promising. 

Try looking at del Cid on staging in two separate tabs. 

In one tab add ?map=canvas to the end of the url. 

There are a shit load of bugs, but the performance improvement on zoom and pan is dramatic. 

If we can get to the bottom of all of the bugs, this will put us in a good place re performance. 

We also now have telemetry which means we can see things like frame rate and dropped frames across
All devices so that we can see problems before users even report them. It gives us an object comparison against the previous implementation across all device types. 

But one thing I should mention. We really need to make sure that maps that come in are performant in the first place. IMO classical is a pretty map. Other maps shouldn't need to be much more complex than it right? I see no reason that del cid should perform so terribly compared with classical. It's totally unusable even on a high powered iPhone.

JorenC:
This is amazing! Not opened it yet, but I know this is the right direction.
With the whack a mole I've seen the night and day it can be at certain maps.

About performance; I see it differently - the opposite. All maps will
perform worse than classical!

Classical is the default map. I spent around 40 hours on that optimising
the hell out of it, because every KB extra saved is a KB extra lived by
everyone all the time. I manually cut every province border in pieces...

But I can't optimise every map like that. Already now I'm automating a lot
of optimisations, like embedding only used glyphs instead of outlining
text, simplifying the AI detected path.

But more importantly, if we allow community variants, we will eventually
get sub optimized maps. Example - my AI algorithm cuts borders so every
line is only passed once. Counterink just copied all layers, gave them a
stroke, meaning that borders are rendered three times, each of a different
layer. We can't expect every variant maker to understand that specifics of
optimizing their map.

Finally, a 20 player map will be much bigger and more detailed than
classical, so expected to run worse.


Also, if you look at Figma, it can handle dynamic rendering of all this in
the browser easily without performance hit. Like you said, it's 2026, we
should be able to display a 2 MB svg.... In addition, while Martin/zond did
use a rasterize on interaction hack (the map would blur and unblur), it
could handle all of these maps - even more heavily suboptimised -
relatively well. Some of these maps were 6-8 MB at the time. I optimized
them when I imported them into the new style.


- that being said, i think we need to find the middle ground.  Optimizing
for size - like I did - might actually not optimise for performance.

I think that we should:
- Try our effing best to render as performant as possible. Every
millisecond the renderer improves is improvement everywhere. This is
similar to optimizing classical.
- identify slow devices and use hacks/settings. I've seen the night/day
difference on certain rendering techniques per device - e.g. the
will-change: transform etc. this is really where i believe we need to make
an exception to the "no toggles" rule - yes, rasterizing makes the map look
bad, but the performance gain is worth it. Like running a 3D game with all
settings on minimal - you care that you can play it, not that it looks like
the ads.

Then, open these settings (and ideally already auto-detect and apply). A
"graphic" Vs "performance" mode is common on Xbox etc.

That does not say we should just not care about quality, but try to apply
the right hacks (e.g. blur on interaction) per device. And use maybe 2-3
approaches.

- identify what hits performance in the maps and optimise for that. My maps
have a shadow. In old dip and Figma this runs fine - in the RPPZ this
shadow gets dynamically recalculated every frame - hit. With Claude I
identified that and used webp to rasterize the blur on the background
layer. No real visual impact and both a size and performance improvement.
Knowing that we can
- automate certain optimisations (e.g. in the dSVG creator)
- ban certain patterns ('do not use gaussian blur on your layers, as it
hits performance') and guide map makers. This could be gradients etc.
- optimize for that




I'm curious why you mention del Cid. The first version of the maps I made
had super detailed vector lines, and it ran like shit. I completely redid
the map, and it runs fine on my end (in bed now, haven't tried your new
style). So, I'm wondering why you take that one. On my end, cold war and
spice islands really struggle...



I might sound like "it's a you problem" - that is not my intention, we need
to look at it from both sides, but we first need to know what elements of
the SVG impact performance before we talk optimization, and we need to
understand that MY optimization can't be applied to all variant makers, as
they are not experts with both our detailed knowledge. And, I can't spend
20 hours creating a map anymore... It needs to run more automated.

But I think we're making great, great progress. This telemetry is so great.
It will help us find these answers. Amazing work!!!!

johnpooch:
> I'm curious why you mention del Cid

I mention this variant because when I run this variant on my iPhone 17 in Safari, it's absolutely unplayable. 

---

About enabling different settings for different device types, I strongly believe that this is not the right direction. It will get us into trouble. 

---

> while Martin/zond did use a rasterize on interaction hack

By my memory, this hack was a pretty ugly user experience. I'd like to avoid it if possible. 

---

> About performance; I see it differently - the opposite. All maps will perform worse than classical!

Okay, it's fine if classical happens to be the most performant map because of extensive optimization, but what I'm getting at is that we need to have some benchmark in place that helps us understand what a good example of quality versus performance looks like. 

All I'm saying is that we need to consider approaching this from both directions. Some maps are massively less performant than others while not being obviously more detailed or beautiful. 

My main suggestion would be to clearly surface the important metrics on the variant after it is uploaded, so that we can eyeball variants that are not well optimized. At least then we understand why rendering might be a little bit sluggish on a certain variant. 

---

> but we first need to know what elements of the SVG impact performance before we talk optimization

Yes, I 100% agree with this. We don't know if it's the map size or the path detail or the number of layers or shadows, etc. 

---

But yes, I think we're generally aligned here. We'll start by trying to squeeze the most out of the rendering code possible. If we still see performance issues with certain variants, we will take a data-led approach to optimizing them.

johnpooch:
Claude:

There are a large number of rendering bugs on the canvas version of the map. The original SVG version of the map renders correctly. 

I'll start by listing the bugs that I see on desktop. 

In the original version of the map, there is a floating action button at the bottom right of the map which changes the map from being clamped such that the camera cannot go beyond the boundaries of the map, to be wider such that the full map fills the viewport. This button is not visible on the new implementation. 

In the original version of the map, I can zoom out to the point at which the map is at the minimum zoom, which still fits within the viewport. In the new version, I cannot zoom out beyond the initial zoom level, which is too zoomed in. 

In the original version of the map, when I click a province that has an orderable unit, a floating menu appears where I can select the order type. In the new map, I can't see the floating menu. I wonder if the map is being rendered above it. 

In the original version of the map, the Province SVG is highlighted when I hover on it or select it, etc. In the new version of the map, the highlighted hover path is a simplified shape. I assume that this is actually the touch detection shape rather than the visible shape. This looks wrong. 

On mobile, all of the issues mentioned above are present as well. The bottom menu, where the order type options appear, renders weirdly such that I can only see one option. Again, I think this is a case of the map rendering on top of the bottom menu.

johnpooch:
Claude: Fixed the rendering/interaction bugs on the Leaflet map. The rasterise-once architecture and the SVG-by-default behaviour are unchanged — these are purely the bug fixes from your list.

**1 & 2 — fullscreen FAB + zoom-out range.** Ported the fill/contain concept from the SVG map. There's now a bottom-right FAB toggling between *fill* (board fills the viewport, clamped to the edges) and *contain* (zoom out until the whole board is visible). The "can't zoom out" trap was `getBoundsZoom` clamping its result to the map's current zoom range — the default minimum of `0` blocked the *negative* zoom needed to fit a board larger than its container, so the contain fit silently came back as 0. The zoom floor is now dropped before querying, and the board re-fits on resize while you're still zoomed all the way out.

**4 — hover/selection shape.** Hover/selection/highlight now draw the **exact dSVG province path** via a dedicated highlight overlay, instead of styling the decimated hit-test polygons (which is what made it look chunky). The decimated polygons stay invisible and are used only for click/hover detection. Fills/strokes match the SVG map (selected white 0.8, hover white 0.6, highlight stripes, white stroke 5).

**3 & 5 — order menu / mobile sheet hidden behind the map.** Leaflet assigns z-indexes up to ~700 to its panes which were leaking into the root stacking context and painting over the portalled `FloatingMenu` / bottom sheet. The Leaflet container is now wrapped in an isolated stacking context (`isolation: isolate`), so the app menus sit above it. Verified with `elementFromPoint`: with the fix the menu item is the topmost element; remove the isolation and a Leaflet `<path>` covers it again. The mobile "only one option visible" is the same issue — the full sheet now renders above the map.

Also checked the fleet-colour concern at deep zoom — units/orders come straight from the shared renderer output and match the SVG map; no colour bug found.

> Note: under MSW mocks `GET /game/:id/options/` is empty, so the order menu can't be triggered there. The menu screenshots below were produced with a temporary local options stub (not committed); the stacking fix itself is also verified structurally as described above.

### Screenshots

**Fill vs contain (FAB) — bugs 1 & 2**
| Fill (default) | Contain (whole board) |
|---|---|
| fill | contain |

**Exact-shape highlight — bug 4**
hover

**Order menu above the map — bugs 3 & 5**
| Desktop floating menu | Mobile bottom sheet |
|---|---|
| desktop menu | mobile menu |


---

PR #891, opened 2026-06-25
Add colorblind mode to user profile and map rendering

### LINKED ISSUE #890: Add colorblind mode to user profile and map rendering
## Goal

Allow users to select a colorblind mode that replaces nation colors with an optimized palette throughout the UI.

## Context

Nation colors appear in province fills, unit fills, supply-center fills, `NationBadge`, and `NationFlag`. The palette substitution should happen on the backend rather than in the frontend: the API returns already-adjusted colors based on the authenticated user's `colorblind_mode`, so the frontend remains a dumb renderer with no palette logic. This also makes the substitution straightforward to test at the serializer level.

The setting persists on the backend (not in localStorage) so it applies on every device the user logs into.

## Approach

1. Add a nullable `colorblind_mode` field to `UserProfile` with choices `deuteranopia`, `protanopia`, `tritanopia` (`null` means disabled). Add the migration.
2. Expose the field via `GET /user/` (read) and `PATCH /user/update/` (write) using the existing `UserProfileSerializer`.
3. When serializing nation colors, apply the user's colorblind palette substitution server-side so the frontend receives pre-adjusted hex values. Run codegen to regenerate the TypeScript client.
4. Add a colorblind mode selector (off / deuteranopia / protanopia / tritanopia) to the account settings screen (`packages/web/src/screens/Home/Account.tsx`), persisting the choice via the existing `useUserUpdatePartialUpdate` mutation.

Closes #878, #879, #880.

johnpooch:
Closes #890

## What this does

Adds a colorblind mode setting to user profiles with three modes: deuteranopia, protanopia, and tritanopia (null = off).

Nation colors are adjusted **server-side** when serializing the variant list — the frontend receives pre-adjusted hex values and remains a dumb renderer. Color substitutions use the Okabe-Ito palette.

A colorblind mode selector (Off / Deuteranopia / Protanopia / Tritanopia) is added to the Account settings screen under Appearance, persisted via `PATCH /user/update/`.

## Changes

- `service/common/colorblind.py` — palette mapping and `apply_colorblind_palette()` helper
- `service/common/constants.py` — `ColorblindMode` constants class
- `service/user_profile/models.py` + migration — nullable `colorblind_mode` field on `UserProfile`
- `service/user_profile/serializers.py` — `colorblindMode` exposed via existing user profile endpoints
- `service/nation/serializers.py` — `NationSerializer.color` applies palette from `request.user.profile.colorblind_mode`
- `packages/web/src/screens/Home/Account.tsx` — colorblind mode selector in Appearance section
- Codegen: `openapi-schema.yaml` and `endpoints.ts` regenerated with `ColorblindModeEnum`

## Screenshots

Account screen (desktop):

account screen

Account screen (mobile):

account screen mobile

---
_Generated by [Claude Code](https://claude.ai/code/session_013ECw5aLMc6Vwe93eookzja)_


johnpooch (commented):
Review of PR #891 — several issues to address before merging.

---

## High severity

### `service/nation/serializers.py` — bare `except Exception: pass`

`get_color` silently discards every exception with `except Exception: pass`. Only `AttributeError` (or `RelatedObjectDoesNotExist`) can arise from a missing profile — programming errors or DB errors would be swallowed and hidden. Replace with:

```python
try:
    mode = self.context["request"].user.profile.colorblind_mode
except AttributeError:
    mode = None
```

Or use `getattr(self.context["request"].user, "profile", None)` and guard on `None`.

---

### `packages/web/src/api/generated/openapi-schema.yaml` — duplicate schema file

A full copy of the OpenAPI YAML (3 500+ lines) has been added to `packages/web/src/api/generated/`. Per CLAUDE.md, `packages/web/src/api/generated/` is for orval-generated TypeScript only. The canonical schema lives at `service/openapi-schema.yaml`. Having a second copy here will diverge silently. Remove this file; if orval needs a YAML input path, point it at `service/openapi-schema.yaml`.

---

### `service/variant/tests.py` — mutating session-scoped `primary_user` fixture

`TestNationColorColorblindAdjustment` sets `primary_user.profile.colorblind_mode = "deuteranopia"` and calls `.save()` in multiple test methods. CLAUDE.md is explicit: *session-scoped users/clients persist across the whole session — never mutate or delete them.* If these tests run before others that depend on `colorblind_mode` being `None`, those tests break non-deterministically.

Use `user_factory` to create a disposable user for tests that mutate profile state, or add teardown that resets the field to `None`.

---

### `service/variant/tests.py` — `viewname` likely undefined

`TestNationColorColorblindAdjustment` calls `reverse(viewname)` but the diff doesn't show `viewname` defined in the class. If it isn't a class attribute defined elsewhere in the file, this is a `NameError` at runtime. Verify and fix.

---

## Medium severity

### `service/nation/serializers.py` — N+1 query: `request.user.profile` fetched per nation

`get_color` accesses `request.user.profile` once per `Nation` in the queryset. For a 7-nation variant that's 7 redundant DB hits for the same row. Resolve the mode once — e.g. in `get_serializer_context` on the view (so it's a dict lookup in the method) or read it once in an overridden `to_representation`.

---

### `service/nation/serializers.py` — `get_flag_url` missing `@extend_schema_field`

The PR correctly adds `@extend_schema_field` to the new `get_color` method field but `get_flag_url` in the same serializer still lacks the annotation. CLAUDE.md requires all `SerializerMethodField`s to be annotated. Add `@extend_schema_field(serializers.URLField(allow_null=True))` to `get_flag_url` in this same change.

---

### `service/user_profile/serializers.py` — `ChoiceField` should take full `CHOICES` tuple

```python
# current (deviates from codebase pattern)
serializers.ChoiceField(choices=[c[0] for c in ColorblindMode.CHOICES], ...)

# consistent with every other ChoiceField in the project
serializers.ChoiceField(choices=ColorblindMode.CHOICES, ...)
```

---

### `packages/web/src/screens/Home/Account.tsx` — `mutateAsync` called without try/catch

The `onValueChange` handler calls `updateProfileMutation.mutateAsync(...)` with no `try/catch`. CLAUDE.md requires all mutation handlers to have try/catch with an error toast:

```typescript
try {
  await updateProfileMutation.mutateAsync({ data: { colorblindMode: value } });
} catch {
  toast.error("Failed to update colorblind mode");
}
```

---

### `service/user_profile/tests.py` — possible wrong response key (snake_case vs camelCase)

The new tests assert on `response.data["colorblind_mode"]` (snake_case), but the API applies camelCase conversion middleware — the JSON response key is `colorblindMode`. If `response.data` is the post-conversion parsed response, the key is absent and the assertions pass vacuously. Verify the correct key and fix all four new test methods.

---

## Low severity

### `packages/web/src/screens/Home/Account.tsx` — potentially redundant `queryClient.invalidateQueries`

The colorblind handler manually calls `queryClient.invalidateQueries(...)` after `mutateAsync` succeeds; the existing email-notifications toggle on the same screen doesn't. If `useUserUpdatePartialUpdate` already handles cache invalidation via `onSuccess`, this is redundant. Check the generated hook and remove if unnecessary.

---

### `service/common/colorblind.py` — hex case sensitivity

`apply_colorblind_palette` does `palette.get(color, color)`. If `Nation.color` stores lowercase hex (`#f44336`) but palette keys are uppercase (`#F44336`), the lookup silently returns the original color with no error. Normalise before lookup (e.g. `color.upper()`) or add a test verifying the case matches fixture data.

---

JorenC:
Right now it's not working. When I set my colour blind mode to off, or to two different types, opening my map (this is a sandbox) but the colours do not change. They should change to the right colour palette.

Screenshot_20260625-110309
Screenshot_20260625-110259
Screenshot_20260625-110249

Each screenshot was taken with a different option selected. One was off.

johnpooch:
Thanks for testing this, @JorenC — you've caught a real bug (two actually).

**Root causes:**
1. The backend ETag for `GET /variants/` didn't include the user's colorblind mode, so the server returned `304 Not Modified` even after the mode changed — the browser never received the updated colors.
2. The frontend didn't invalidate the variants query cache after saving the colorblind mode preference, so React Query kept serving the stale response from the initial page load.

**Fix pushed** (commit 252c78a): the ETag now incorporates the colorblind mode, and the frontend invalidates the variants query alongside the user profile query when the setting changes. Both issues needed to be fixed together — the cache invalidation alone wouldn't help if the server kept returning 304s.

---

PR #893, opened 2026-06-25
Mobile: tapping Chat while in a channel returns to chat overview

### LINKED ISSUE #892: Mobile: tapping Chat while in a channel returns to the chat overview
## Goal

On mobile, tapping the Chat nav item while already viewing a chat channel should return to the chat overview (channel list), instead of staying in the current channel.

## Context

The bottom nav Chat item preserves the `channelId` search param so that tapping Chat from Map or Orders resumes the last open channel — that behaviour should be kept. The desktop sidebar already drops `channelId` to land on the channel list. We want the same "tap the active tab to go back to the overview" behaviour on mobile, without breaking resume-from-elsewhere.

## Approach

In `GameDetailLayout.tsx`, the bottom nav Chat item should drop the `channelId` param (as `sidebarNavItems` already does) when the current location is already inside a chat channel; otherwise keep it so resuming the last channel from Map/Orders still works.

johnpooch:
Closes #892

When the user is already inside a chat channel on mobile, tapping the Chat bottom nav item now navigates to the channel list instead of staying in the current channel. Resume-from-Map/Orders (preserving `channelId` in the search params) is unchanged.

The fix is in `GameDetailLayout.tsx`: when `location.pathname` is inside a chat channel path (`/chat/channel/...`), the bottom nav Chat item drops `channelId` from its target URL, mirroring what `sidebarNavItems` already does for the desktop sidebar.

chat channel on mobile

---
_Generated by [Claude Code](https://claude.ai/code/session_01XrUczFgHqTb3hq9PsfNsus)_

### DIFF
--- a/packages/web/src/components/GameDetailLayout.tsx
+++ b/packages/web/src/components/GameDetailLayout.tsx
@@ -52,6 +52,8 @@ const GameDetailLayout: React.FC<GameDetailLayoutProps> = ({
       ? navigationItems.filter(item => item.label !== "Chat")
       : navigationItems;
     const searchParamsStr = searchParams.toString();
+    const chatBasePath = `/game/${gameId}/phase/${phaseId}/chat`;
+    const isInChatChannel = location.pathname.startsWith(chatBasePath + "/");
     return items.map(item => {
       const basePath = item.path
         .replace(":gameId", gameId)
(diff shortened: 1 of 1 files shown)


JorenC:
Confirmed correct. Staging can be torn down, @johnpooch you can pull this without confirming behaviour - just quickly scan the code (minimal fix)

johnpooch (commented):
Logic is correct and the implementation is clean. A few observations:

- `isInChatChannel` correctly uses `startsWith(chatBasePath + "/")` — the trailing slash ensures the channel list root (`/chat`) is not treated as "in a channel", so the special-case only fires when navigating away from a specific channel.
- Hoisting `chatBasePath` above `.map()` is a minor improvement over the original (computed once rather than per-iteration).
- The intentional asymmetry with `sidebarNavItems` — which unconditionally strips `channelId` regardless of current location — is correct per the issue spec and consistent with the desktop UX.
- Screenshot in the PR description is appreciated.

LGTM.

---

PR #895, opened 2026-06-25
Fix colorblind mode map recolour without page reload

### LINKED ISSUE #894: Colorblind mode change doesn't recolour the map without a page reload
## Goal

Make the map recolour immediately when colorblind mode changes (and whenever any variant data changes), on both web and native, without requiring a full page reload.

## Current behaviour

Changing colorblind mode in Account settings does not recolour the map. The user has to do a full browser reload before the new colours appear. On native (Capacitor) there is no reload gesture, so it looks permanently broken there.

This persists *even after* the recently merged fix `252c78a` ("Fix colorblind mode not taking effect after change"), which folded colorblind mode into the server ETag and added `invalidateQueries(getVariantsListQueryKey())` to the Account handler. Those changes are correct but not sufficient — a second cache layer still serves stale data.

### Root cause (evidenced)

1. Changing the mode in `packages/web/src/screens/Home/Account.tsx` (~lines 213–221) PATCHes the profile, then calls `queryClient.invalidateQueries({ queryKey: getVariantsListQueryKey() })`. React Query dutifully refetches `GET /variants/` via the axios instance (`packages/web/src/api/axiosInstance.ts`).
2. The **browser/WebView HTTP cache intercepts that refetch**. The server sends `Cache-Control: private, must-revalidate, max-age=60` from `VariantListCreateView.list` in `service/variant/views.py`. Within that 60-second window the browser treats the cached response as *fresh* and serves the old-colour JSON **without any network request** — so the server's (now correct) ETag logic never even runs.
3. React Query stores the stale colours. `packages/web/src/components/GameMap.tsx:74` (`useVariantsList()`) → `packages/web/src/components/InteractiveMap/toRenderState.ts:47-48` (`nationColors[nation.name] = nation.color`) recolours with the same values, so nothing visibly changes.
4. A full page reload uses the browser's "reload" cache mode, which forces revalidation → the server sees the changed mode in the ETag → returns `200` with fresh colours. That is exactly why "you have to refresh".

`max-age=60` is also redundant with the frontend: React Query already sets `staleTime: 60 * 60 * 1000` on the variants query (`packages/web/src/App.tsx:26-28`). React Query — not the HTTP cache — is the layer that prevents over-fetching on navigation. The HTTP `max-age=60` window saves nothing React Query hasn't already suppressed, while actively hijacking the one moment we deliberately want fresh data (the explicit `invalidateQueries`).

## Expected behaviour

After changing colorblind mode, the map recolours immediately with no reload, identically on web and native. The same holds for any future variant edit (e.g. a draft owner changing nation colours) — the change is reflected on the next React Query refetch.

## Where

- `service/variant/views.py` — `VariantListCreateView.list`, the `Cache-Control` header.
- Supporting context: `packages/web/src/App.tsx` (React Query `staleTime`), `packages/web/src/screens/Home/Account.tsx` (the invalidation), `packages/web/src/components/GameMap.tsx` + `packages/web/src/components/InteractiveMap/toRenderState.ts` (the colour render path), `service/variant/views.py` `_variants_list_etag` (the ETag).

## Approach

One-line backend change: in `VariantListCreateView.list`, change the `Cache-Control` header from `private, must-revalidate, max-age=60` to `private, no-cache`.

- `no-cache` means the browser/WebView may store the response but **must revalidate with the server before every use**. The existing ETag path (`If-None-Match` → `304`) still produces cheap `304 Not Modified` responses when nothing changed, so the efficiency is preserved; a changed mode immediately returns `200` with fresh colours.
- Deliberately **not** `no-store` — that would disable the `304` optimisation and always re-send the full body.
- No frontend change is required; this fixes web and native together.

### Exact diff

`service/variant/views.py`, in `VariantListCreateView.list`:

```python
        response["ETag"] = etag
-        response["Cache-Control"] = "private, must-revalidate, max-age=60"
+        response["Cache-Control"] = "private, no-cache"
        return response
```

### Tests

Add to `service/variant/tests.py` (alongside the existing `test_etag_changes_when_colorblind_mode_changes`):

```python
@pytest.mark.django_db
def test_list_variants_cache_control_forces_revalidation(authenticated_client):
    response = authenticated_client.get(reverse(viewname))

    assert response["Cache-Control"] == "private, no-cache"


@pytest.mark.django_db
def test_list_variants_returns_304_when_etag_matches(authenticated_client):
    response1 = authenticated_client.get(reverse(viewname))
    etag = response1["ETag"]

    response2 = authenticated_client.get(reverse(viewname), HTTP_IF_NONE_MATCH=etag)

    assert response2.status_code == status.HTTP_304_NOT_MODIFIED
```

The first pins the no-stale-window contract; the second proves the `304` revalidation path still works. The existing query-count tests at `service/variant/tests.py:119,155` (18 prefetch + 2 ETag aggregates) remain valid — the change does not alter query counts.

### Cost

Effectively zero in practice, because React Query's 1-hour `staleTime` already governs how often `/variants/` is fetched. The `max-age=60` window only ever shadowed requests React Query had already suppressed. Concretely:

- **Server:** a few extra conditional GETs per session. When the ETag matches (the common case), `list` returns `304` *before* running `super().list()`, so the cost is just the two indexed `Max(updated_at)` aggregates in `_variants_list_etag` plus a headers-only response — no 18-query prefetch, no serialization. The full list path runs only when data actually changed, which is unavoidable and is the point of the fix.
- **Bandwidth:** `304`s are headers-only. The heavy 2 MB map SVG is served by a *separate* endpoint (`VariantSvgView`, `Cache-Control: public, max-age=31536000, immutable`) and is **not** touched by this change.
- **User-perceived latency:** none. React Query renders from its in-memory cache instantly and revalidates in the background; with `staleTime` at 1 hour, genuine refetches are rare.
- **Native (Capacitor):** same analysis, and it is the only path that gains the correctness fix since there is no reload gesture.

johnpooch:
Changes `Cache-Control` on `GET /variants/` from `private, must-revalidate, max-age=60` to `private, no-cache`. The 60-second `max-age` window was intercepting React Query's explicit `invalidateQueries` call after a colorblind mode change — the browser/WebView served the cached (old-colour) response without hitting the server, so the ETag logic never ran and the map stayed the same colour.

`no-cache` forces revalidation on every request while preserving the ETag/304 path: when nothing has changed the server returns `304 Not Modified` (headers only, no body); when the mode changed it returns `200` with fresh colours. The efficiency trade-off is effectively zero because React Query's `staleTime: 60 * 60 * 1000` already governs how often `/variants/` is fetched — the `max-age=60` window only ever shadowed requests React Query had already suppressed.

Two tests added: one pins the new `Cache-Control` contract, one proves the `304` revalidation path still works.

Closes #894

---
_Generated by [Claude Code](https://claude.ai/code/session_01WEpHQ6vpqT4Dpep176XM5q)_

### DIFF
--- a/service/variant/tests.py
+++ b/service/variant/tests.py
@@ -79,6 +79,23 @@ def test_etag_changes_when_colorblind_mode_changes(user_factory, authenticated_c
 
 
 @pytest.mark.django_db
+def test_list_variants_cache_control_forces_revalidation(authenticated_client):
+    response = authenticated_client.get(reverse(viewname))
+
+    assert response["Cache-Control"] == "private, no-cache"
+
+
+@pytest.mark.django_db
+def test_list_variants_returns_304_when_etag_matches(authenticated_client):
+    response1 = authenticated_client.get(reverse(viewname))
+    etag = response1["ETag"]
+
+    response2 = authenticated_client.get(reverse(viewname), HTTP_IF_NONE_MATCH=etag)
+
+    assert response2.status_code == status.HTTP_304_NOT_MODIFIED
+
+
+@pytest.mark.django_db
 def test_list_variants_includes_svg_url(authenticated_client):
     response = authenticated_client.get(reverse(viewname))
 
--- a/service/variant/tests/test_variants_list_performance.py
+++ b/service/variant/tests/test_variants_list_performance.py
@@ -82,8 +82,7 @@ def test_list_variants_returns_etag_and_cache_control(authenticated_client):
     response = authenticated_client.get(reverse("variant-list"))
     assert response.status_code == status.HTTP_200_OK
     assert response["ETag"].startswith('"') and response["ETag"].endswith('"')
-    assert "max-age=60" in response["Cache-Control"]
-    assert "must-revalidate" in response["Cache-Control"]
+    assert response["Cache-Control"] == "private, no-cache"
 
 
 @pytest.mark.django_db
--- a/service/variant/views.py
+++ b/service/variant/views.py
@@ -76,7 +76,7 @@ class VariantListCreateView(generics.ListCreateAPIView):
         else:
             response = super().list(request, *args, **kwargs)
         response["ETag"] = etag
-        response["Cache-Control"] = "private, must-revalidate, max-age=60"
+        response["Cache-Control"] = "private, no-cache"
         return response
 
 


johnpooch (commented):
**PR review (automated — triggered by `workflow: needs-pr-review`)**

Clean, minimal fix. No blocking issues found.

- **Backend change**: `private, no-cache` is correct — forces revalidation while preserving the ETag/304 path. Efficiency trade-off is effectively zero since React Query's `staleTime: 1h` already governs fetch frequency; `max-age=60` only ever shadowed requests React Query had already suppressed.
- **Tests**: Both new tests check observable HTTP outcomes (not internal state), consistent with project conventions. `test_list_variants_cache_control_forces_revalidation` pins the new contract; `test_list_variants_returns_304_when_etag_matches` proves the 304 revalidation path still works.
- **Performance test update**: Correct — assertion updated to match the new header value.
- **Scope**: No scope creep. Exactly the diff described in the issue.

One minor observation (not blocking): the new `test_list_variants_cache_control_forces_revalidation` in `tests.py` and the updated `test_list_variants_returns_etag_and_cache_control` in `tests/test_variants_list_performance.py` both assert `response["Cache-Control"] == "private, no-cache"`. Slight duplication across files, but acceptable given different file contexts.

---

JorenC:
the preview is not working - it gives a 404

johnpooch:
This is a backend-only change (one `Cache-Control` header in `service/variant/views.py`) — Netlify skipped the frontend deploy preview because no files in `packages/web/` changed, hence the 404.

To test the fix: use the staging backend at https://diplicity-react-staging-pr-895.up.railway.app with email/password login, change colorblind mode in Account settings, and verify the map recolours without a page reload.

---

johnpooch:
> the preview is not working - it gives a 404
> 
> 

There can only be one staging environment. The 404 was likely caused by us deploying another pr to staging. Going to redeploy to staging now.

PR #897, opened 2026-06-25
Persist implicit Hold orders for dislodged units with no explicit order

### LINKED ISSUE #896: Dislodged unit with no explicit order shows no failed hold
## Goal

When a unit is dislodged during a movement phase without having submitted an explicit order, the completed phase view should show a failed hold order for that unit.

## Current behaviour

Units that don't submit an order are given an implicit hold during adjudication, but that hold is never persisted to the database. If the implicit hold fails (e.g. the unit is dislodged), no `Order` or `OrderResolution` record is created, so the completed phase view shows nothing for that unit — there is no indication the hold was attempted and failed.

## Expected behaviour

A dislodged unit with no explicit order should appear in the completed phase view with a failed hold order, the same as a unit that explicitly submitted a hold and was dislodged.

## Where

`service/phase/models.py` — `create_from_adjudication_data`. The `Order` model already has an `is_implicit` boolean field intended for this case.

## Approach

In `create_from_adjudication_data`, before persisting resolutions, create `Order` records (with `is_implicit=True`) for any province that has a failed resolution but no existing order row. This means only dislodged/bounced implicit holds are persisted — successful quiet holds remain invisible.

johnpooch:
## Summary

Closes #896

When a unit submits no explicit order, the adjudicator gives it an implicit hold during resolution. If that hold fails (the unit is dislodged or bounced), the adjudicator emits a failed resolution for the province — but because no `Order` row exists, `create_from_adjudication_data` silently dropped the result, leaving the completed phase view showing nothing for that unit.

The fix is in `PhaseManager.create_from_adjudication_data` (`service/phase/models.py`): before processing existing orders' resolutions, scan `adjudication_data["resolutions"]` for any failed result (`result != "OK"`) that has no corresponding `Order` row. For each such resolution, create an implicit `Order` with `is_implicit=True` and `order_type=Hold` on the appropriate `PhaseState`, then create its `OrderResolution` alongside the rest. Successful quiet holds are not persisted — only failed ones.

One subtlety: `all_orders` is now saved to a local list up front so that newly created implicit orders don't get picked up by a later iteration of the same property (which would re-query the DB when the phase isn't loaded via `with_adjudication_data`).

---
_Generated by [Claude Code](https://claude.ai/code/session_01CXm6wUQD7KAUgTkq95JSkp)_


JorenC:
Do you have a screenshot to confirm the implementation?

johnpooch:
This is a backend-only change — there's no frontend output to screenshot. The fix is in `create_from_adjudication_data` in `service/phase/models.py`: when the adjudicator returns a failed resolution for a province that has no explicit order (e.g. a dislodged unit that the player left without instructions), we now create an implicit Hold order with `is_implicit=True` and attach the failure resolution to it. This makes the resolution visible in the API response alongside all other orders.

The correctness is verified by the two new tests added in `service/phase/tests.py`:
- `test_create_from_adjudication_data_creates_implicit_hold_for_failed_resolution_without_explicit_order` — confirms the implicit Hold order and its `ErrBounce` resolution are created
- `test_create_from_adjudication_data_does_not_create_implicit_hold_for_successful_resolution_without_explicit_order` — confirms no spurious implicit orders appear when all resolutions are OK

---

service/phase/models.py
```diff
@@ -636,15 +636,51 @@ def create_from_adjudication_data(self, previous_phase, adjudication_data):
...
+                                        Order(
+                                            phase_state=phase_state,
+                                            source=source_province,
+                                            order_type=OrderType.HOLD,
+                                            is_implicit=True,
+                                        )
+                                    )
+                                    implicit_resolution_pairs.append(resolution_data)
+
+                    if implicit_orders_to_create:
```
johnpooch, line 676: The `if implicit_orders_to_create:` guard appears twice — once here to bulk-create the orders, and again after the existing-orders loop to build their resolutions. `created_implicit_orders` is only assigned in the first block and only used in the second, with unrelated code in between. A linter will flag it as possibly undefined, and the intent is harder to follow.

Move `resolutions_to_create = []` before this block and merge both guards into one:

```python
resolutions_to_create = []

if implicit_orders_to_create:
    created_implicit_orders = Order.objects.bulk_create(implicit_orders_to_create)
    for implicit_order, res in zip(created_implicit_orders, implicit_resolution_pairs):
        by_province = province_lookup.get(res["by"]) if res["by"] else None
        resolutions_to_create.append(
            OrderResolution(
                order=implicit_order,
                status=res["result"],
                by=by_province,
            )
        )

for order in existing_orders:
    ...
```

The ordering in `resolutions_to_create` doesn't matter for `bulk_create`, so this is safe.

---

service/phase/models.py
```diff
@@ -636,15 +636,51 @@ def create_from_adjudication_data(self, previous_phase, adjudication_data):
...
-                    order_count = len(previous_phase.all_orders)
+                    existing_orders = list(previous_phase.all_orders)
+                    order_count = len(existing_orders)
 
                     # Delete existing resolutions in bulk
-                    order_ids = [order.id for order in previous_phase.all_orders]
+                    order_ids = [order.id for order in existing_orders]
                     OrderResolution.objects.filter(order_id__in=order_ids).delete()
 
+                    # Build lookups needed to create implicit Hold orders
```
johnpooch: Per the backend style guide: "Do not add docstrings or comments." These two new section-header comments (`# Build lookups needed to create implicit Hold orders` and `# Create implicit Hold orders for failed resolutions with no existing order row`) explain what the code does, which the identifiers already communicate. Please remove them.

---

johnpooch (commented):
Two findings:

1. **Structural**: The `if implicit_orders_to_create:` guard is split across two locations with unrelated code in between, and `created_implicit_orders` is only assigned in the first block but used in the second. Merging both guards into one block (see inline comment) eliminates the "possibly undefined" variable smell and makes the intent clearer.

2. **Style**: Two new section-header comments were added that explain *what* the code does. Per the backend style guide, comments should not be added. Please remove them.

---

JorenC:
Hold orders that are implicit  should have the black octagon to be 30% transparent, while the error cross should be 100% opaque.

Otherwise - if not implicit - it should be 100% opaque.

This is a front end change. Make the screenshot to show the difference.

johnpooch:
Done. The frontend changes are in the latest commit.

**What changed:**
- `is_implicit` is now exposed from `OrderSerializer` (and the OpenAPI schema + generated TypeScript types are updated)
- `holdMarkup` renders the octagon at `opacity="0.3"` when `isImplicit` is true; the failure cross is a sibling element inside the `<g>` so it remains at full opacity
- `isImplicit` is threaded through `OrderState` → `toRenderState` → `mapRenderer`

**Before** (explicit hold, full opacity octagon):
explicit hold

**After** (implicit hold, 30% opacity octagon — Norway, top-centre of map):
implicit hold

Russia's army in Norway (no order submitted) shows the faded octagon with a full-opacity red cross. England's fleet in Norwegian Sea (explicit hold, succeeded) shows the full-opacity octagon for comparison.

---

johnpooch (commented):
Ran the full `TestCreateFromAdjudicationData` and `TestCreateFromAdjudicationDataPerformance` suites — 10/10 pass, query count still 14. No blocking findings.

The core logic is correct:

- Snapshotting `existing_orders` as a list before creating implicit orders is the right fix — prevents the newly created rows from being picked up in the subsequent `for order in existing_orders` loop.
- The two parallel lists (`implicit_orders_to_create` / `implicit_resolution_pairs`) zipped against `bulk_create`'s return value is sound on PostgreSQL (insertion order is preserved).
- New lookups (`units.all()`, `phase_states.all()`) are covered by the `with_adjudication_data` prefetches (`units__province`, `units__nation__flag`, `phase_states__member__nation__flag`), so no extra queries in production.

**One nit:** `test_create_from_adjudication_data_creates_implicit_hold_for_failed_resolution_without_explicit_order` uses `ErrBounce` (failed hold/move, unit stays in place) but the issue title describes a *dislodged* unit, which produces `ErrForcedDisband` (see `mock_adjudication_data_with_dislodged_unit`). Both result codes exercise the same code path so the fix is verified, but the test doesn't directly reproduce the reported scenario. Worth adding a second case with `ErrForcedDisband` to close that gap — not a blocker.

---

JorenC:
@johnpooch I think this is okay. The screenshots look a bit weird (incorrect adjucation etc) but I don't think that's the critical part, and the order is rendering exactly as it needs:
- explicit hold order: black octagon.
- no hold order: no octagon, just stay
- no hold order/dislodged: transparent (implicit) hold order, with cross.

Changes are mainly in back end, front end is dumb - they use the implicit flag to ensure the order is rendered transparent.

For me, this is ready to pull.

We'll get to Backstabbr levels of map readability soon!

PR #901, opened 2026-06-26
Transfer game management when managing player becomes unavailable

### LINKED ISSUE #900: Transfer game management when the managing player exits
## Goal

When the managing player of a player-run game (no dedicated game master) becomes unavailable through civil disorder, elimination, or being kicked, automatically transfer game management to the most reliable remaining eligible member.

## Context

In games where `game_master` is NULL, `can_manage()` (`service/game/models.py`) grants management rights to the creator via `created_by`. If the creator enters civil disorder, gets eliminated, or is kicked, no one else has management rights and the game becomes unmanageable.

## Approach

Select the new manager by picking the non-eliminated, non-kicked, non-CD member with the highest reliability score from `get_player_stats()`. Trigger the transfer atomically alongside the exit event and notify all members. If no eligible member exists, skip the transfer.

Before implementing, decide how to record who currently holds management in a player-run game. Mutating `created_by` would corrupt historical data; a separate field (e.g. `admin_member`) is likely needed to decouple "who created the game" from "who currently manages it".

johnpooch:
Adds automatic management transfer for player-run games when the current manager becomes unavailable.

A new nullable `managing_member` FK on `Game` decouples "who created the game" from "who currently manages it". When the manager enters civil disorder, is eliminated, or is kicked, `transfer_management_if_needed()` selects the most reliable eligible member (ranked by reliability tier, then combined NMR+CD rate) and assigns them as the new `managing_member`. A `management_transfer` notification is sent to all game members.

`can_manage()` and `IsGameManager` check `managing_member` first (after game-master), so existing management flows for GM games and sandbox games are unaffected.

Closes #900

---
_Generated by [Claude Code](https://claude.ai/code/session_01HUeHyQg4zFw66AfpCTM7Jj)_


JorenC:
## PR Review — Transfer game management when managing player becomes unavailable

Reviewed statically against `main` (no local Postgres/venv in this session, so the suite wasn't executed). Overall this is a well-structured, well-tested change: the `managing_member` FK cleanly decouples "creator" from "current manager", `can_manage()` and `IsGameManager` are updated consistently, and the test file covers CD, elimination, kick, GM/sandbox exclusion, the "no eligible replacement" cases, transfer chaining, and API-level permission assertions. A few things worth addressing before merge.

### 🔴 N+1 on the games-list endpoint

`can_manage()` reads `self.managing_member.user_id` (`game/models.py:565`), and `get_can_manage` is a `SerializerMethodField` on **both** `GameListSerializer` (`game/serializers.py:79`) and `GameRetrieveSerializer` (`:228`). But `managing_member` is not in the `select_related` of any prefetch strategy:

- `with_list_data` → `select_related("variant", "victory", "game_master__profile")` (`game/models.py:120`)
- `with_retrieve_data` → same set (`:149`)

So every game in the list that has a `managing_member` set triggers an extra query when `can_manage` is evaluated — a genuine N+1 that grows as transfers happen in production. `IsGameManager` has the same access (`common/permissions.py:179`) but it's single-game per request, so only the list path is N+1.

Fix is one line — add `"managing_member"` to the `select_related(...)` in `with_list_data` / `with_retrieve_data` (only `.user_id`, a column on `Member`, is needed, so no `__user` join is required). Per the repo convention this also warrants a query-count test on the list endpoint ("Add performance tests to ensure N+1 query issues are avoided").

### 🟡 `get_player_stats` is run once per eligible member during adjudication

`reliability_key` calls `get_player_stats(member.user)` inside `min(eligible, key=...)` (`game/models.py:596-601`), and `get_player_stats` itself issues ~5 queries (`user_profile/utils.py`). For a 7-player game that's up to ~30 queries per transfer, executed synchronously inside the phase-resolution transaction (`phase/models.py:457`, `:562`). Transfers are rare (only when the *current manager* exits), so this is probably acceptable, but it's worth a comment in the PR — and if adjudication latency matters, the selection could be narrowed before computing stats. Minor related point: `new_manager.nation` (`:605`) isn't `select_related` on the `eligible` query (`:590` only fetches `user`), so it adds one more query when building the notification body.

### 🟡 Unrelated migration / scope creep

`variant/migrations/0025_...` only alters `adjudication_modifiers`/`dominance_rules` to `default=list`. The model already has those defaults (`variant/models.py:114,117`), added back in commit `5e86347a` ("Add adjudicationModifiers field on Variant") — this PR is just sweeping up a `makemigrations` that should have shipped with that earlier change. It's harmless, but it's unrelated to management transfer and contradicts the "no scope creep" guideline. Consider splitting it into its own commit/PR, or at least call it out in the description.

### 🟢 Edge case worth a note (not a blocker)

In `member/views.py:perform_destroy`, `instance.delete()` runs *before* `transfer_management_if_needed(kicked_user, ...)`. Because the FK is `SET_NULL`, deleting a member who is the current `managing_member` nulls `managing_member` in the DB first; the subsequent transfer call would then see `created_by_id` as the "current manager" and bail out instead of electing a new manager. In practice this is unreachable today (only the manager can kick, and they won't kick themselves), so the `reason="was kicked"` transfer path is effectively defensive/dead. Fine to leave, but a test documenting the "managing member is removed" expectation (or a comment) would prevent a future surprise if self-kick/GM-kick is ever added.

### Nits

- Tie-breaking in `min(eligible, key=...)` falls back to DB row order (the `eligible` query has no `order_by`), so the chosen manager among equally-reliable members is non-deterministic. The tests correctly assert membership in a set rather than equality, so this is fine — just flagging that the selection isn't stable.
- The transfer notification goes to all members via `notification_user_ids()` with no exclusion, so the eliminated/CD'd ex-manager also receives "management has been transferred … because the previous manager was eliminated." Likely intended, just confirming it's deliberate.

### What's good

- Correct sequencing: `bulk_update` of `civil_disorder`/`eliminated` happens *before* the transfer call in both paths (`phase/models.py:453→457`, `:556→562`), so the `eligible` filter (`eliminated=False, kicked=False, civil_disorder=False`) reads the persisted state — no stale-eligibility bug.
- `transfer_management_if_needed` correctly uses `transaction.on_commit` for the notification, and the elimination path runs inside the resolve transaction (`phase/models.py:602`) so the notification defers properly.
- Reliability-tier scoring degrades gracefully when `reliability_tier` is `None` (`.get(..., 0)`).
- Test coverage is thorough and asserts on observable outcomes (API responses, permission results), matching the repo's testing philosophy.

---
*Static review via Claude Code; the backend test suite was not executed in this session.*

johnpooch:
Thanks for the thorough review. Addressing the actionable items:

**🔴 N+1 fix** — adding `"managing_member"` to `select_related` in both `with_list_data` and `with_retrieve_data`, plus a query-count test on the game-list endpoint. Incoming commit.

**🟡 `nation` N+1** — adding `select_related("user", "nation")` to the eligible member query in `transfer_management_if_needed`. Good catch.

**🟡 Unrelated migration** — agreed on the scope. Migration 0025 was auto-generated during the session and committed separately. I'll leave it here rather than add a revert commit, but happy to split it if you prefer.

**🟢 Self-kick edge case** — confirmed dead code today (manager can't kick themselves, no GM-kick path). Left as-is.

**Nits** — non-deterministic tie-breaking is deliberate (tests assert set membership, not identity). Ex-manager receiving the transfer notification is intentional.

---

service/common/permissions.py
```diff
@@ -175,6 +175,8 @@ def has_permission(self, request, view):
         if game.game_master_id is not None:
             self.message = "Only the game master can perform this action."
             return game.game_master_id == request.user.id
+        if game.managing_member_id is not None:
+            return game.managing_member.user_id == request.user.id
```
johnpooch: Extra DB query here. `game.managing_member_id` is available as an integer FK column without a query, but `game.managing_member.user_id` fetches the `Member` row. The existing `game_master_id == request.user.id` check avoids this by comparing the FK integer directly.

The fix is to add `select_related("managing_member")` when the game is fetched in `has_permission`:

```python
game = get_object_or_404(Game.objects.select_related("managing_member"), id=game_id)
```

Then `game.managing_member.user_id` is served from the cache.

---

service/game/models.py
```diff
@@ -552,13 +561,63 @@ def can_manage(self, user):
         with tracer.start_as_current_span("game.models.can_manage"):
             if self.game_master_id is not None:
                 return self.game_master_id == user.id
+            if self.managing_member_id is not None:
+                return self.managing_member.user_id == user.id
             if self.created_by_id is None or self.created_by_id != user.id:
```
johnpooch, line 616: Same extra query issue as in `permissions.py`: `self.managing_member.user_id` fetches the related `Member` row each time `can_manage` is called when `managing_member_id` is set. `managing_member` should be included in `select_related` on any queryset that fetches games for views where `can_manage` is called (e.g. `with_retrieve_data()`, `with_list_data()`).

---

service/game/models.py
```diff
@@ -552,13 +561,63 @@ def can_manage(self, user):
...
+            if self.managing_member_id is not None
+            else self.created_by_id
+        )
+        if current_manager_id != exiting_user.id:
+            return
+        eligible = list(
+            self.members.filter(
+                eliminated=False, kicked=False, civil_disorder=False
+            ).exclude(user_id=exiting_user.id).select_related("user", "nation")
+            .filter(user__isnull=False)
```
johnpooch: `nation` is missing from `select_related`. Line 606 accesses `new_manager.nation.name`, which will fire an extra query per transfer (since `nation` isn't prefetched here). Change to:

```python
.select_related("user", "nation")
```

---

service/game/models.py
```diff
@@ -552,13 +561,63 @@ def can_manage(self, user):
...
+                return self.managing_member.user_id == user.id
             if self.created_by_id is None or self.created_by_id != user.id:
                 return False
             return any(
                 member.user_id is not None and member.user_id == user.id
                 for member in self.members.all()
             )
 
+    def transfer_management_if_needed(self, exiting_user, reason=None):
+        if exiting_user is None:
```
johnpooch: `transfer_management_if_needed` issues a DB query (`self.members.filter(...)`), updates the game (`self.save()`), and schedules a notification — this is complex modification logic that belongs on a Manager method per the project convention. The model's role is fields + derived properties; mutation flows live in `GameManager` (see `create_from_template`, `clone_from_phase` as examples). Suggest moving this to `Game.objects.transfer_management_if_needed(game, exiting_user, reason)` or similar.

---

service/game/tests/test_management_transfer.py
```diff
@@ -0,0 +1,495 @@
...
+from game.models import Game
+from member.models import Member
+from user_profile.models import UserProfile
+
+retrieve_viewname = "game-retrieve"
+pause_viewname = "game-pause"
+kick_viewname = "game-kick"
+
+User = get_user_model()
+
```
johnpooch: Per project convention, `*_factory` fixtures belong in `conftest.py`, not in the test file itself. `player_run_active_game_factory`, `player_run_pending_game_factory`, `game_master_game_factory`, and `sandbox_game_factory` should be moved to either `service/conftest.py` (if they could be reused across apps) or `service/game/tests/conftest.py` (if game-local). The test file should only contain test classes and functions.

---

service/variant/migrations/0025_alter_variant_adjudication_modifiers_and_more.py
johnpooch, line 1: This migration is unrelated to management transfer — it alters `adjudication_modifiers` and `dominance_rules` on the `Variant` model. It shouldn't be included in this PR. Please either remove it (if it was accidentally included via rebase) or split it out into a separate PR.

---

johnpooch (commented):
The feature logic is sound — `managing_member` cleanly decouples "who created the game" from "who manages it", the guard conditions are correct, and the test coverage is thorough. A few issues need addressing before merge:

**Must fix**
- Unrelated migration (`variant/migrations/0025_...`) — remove or split to a separate PR
- Missing `select_related("nation")` on the eligible member query (extra query per transfer)
- Extra DB query in `IsGameManager.has_permission` when `managing_member` is set — needs `select_related("managing_member")` on the game fetch

**Should fix**
- `can_manage()` same extra query via `self.managing_member.user_id` — add `managing_member` to the relevant QuerySet `select_related` calls
- Test fixtures (`player_run_active_game_factory`, etc.) defined in the test file rather than `conftest.py` — per convention they belong in `service/game/tests/conftest.py`

**Suggestion**
- `transfer_management_if_needed` does DB queries, a save, and notification scheduling — per CLAUDE.md conventions this complexity belongs in a Manager method rather than on the model instance

---

service/phase/models.py
```diff
@@ -452,6 +452,10 @@ def _check_civil_disorder(self, phase):
             m.civil_disorder = True
         Member.objects.bulk_update(newly_cd_members, ["civil_disorder"])
 
+        for m in newly_cd_members:
+            if m.user_id is not None:
+                phase.game.transfer_management_if_needed(m.user, reason="entered civil disorder")
+
```
johnpooch, line 458: `m.user` triggers an extra DB query per CD member here. The `newly_cd_members` list is built from `phase.phase_states.select_related("member")`, which loads the `Member` row but not `member.user`. `_check_eliminations` avoids this correctly with `select_related("user", "nation")` on the candidates query.

The fix is to add `select_related("user")` to the phase-states query used to populate `current_nmr_members`, or to access `m.user_id` and pass a `User` object looked up beforehand. The simplest fix is to extend the existing `select_related("member")` to `select_related("member__user")` on the `phase.phase_states` query inside `_check_civil_disorder`.

---

service/game/models.py
```diff
@@ -305,6 +307,54 @@ def create_sandbox(self, user, *, name, variant, **kwargs):
...
+        if game.sandbox:
+            return
+        current_manager_id = (
+            game.managing_member.user_id
+            if game.managing_member_id is not None
+            else game.created_by_id
+        )
+        if current_manager_id != exiting_user.id:
+            return
+        eligible = list(
```
johnpooch, line 324: `Member.name` (the fallback when `new_manager.nation` is `None`) does `self.user.profile.name`, which requires `user.profile`. Only `user` and `nation` are in the `select_related` here, so hitting the fallback triggers an extra DB query.

Change `select_related("user", "nation")` to `select_related("user__profile", "nation")` so the profile is covered. This also benefits `reliability_key` on line 331, which calls `get_player_stats(member.user)` — worth verifying whether that function accesses `user.profile` too.

---

johnpooch (commented):
Two inline findings, both related to missing `select_related` coverage introduced by this PR.

**Overall:** The design is sound — the `managing_member` FK cleanly separates "who created the game" from "who manages it", the `transfer_management_if_needed` guard conditions are correct (sandbox, GM games, and non-manager exits all short-circuit cleanly), and the test coverage is thorough. The two issues are small `select_related` fixes.

---

JorenC:
What is the notification if a player becomes the new admin/game master?

johnpooch:
All game members (including the new manager) receive a single `management_transfer` push notification:

- **Title**: the game name
- **Body**: `"Game management has been transferred to {nation}{reason}."` — e.g. `"Game management has been transferred to England because the previous manager entered civil disorder."`

The reason suffix is omitted when no reason is provided. No separate targeted notification is sent to the new manager specifically.

---

service/game/tests/conftest.py
johnpooch, line 1: Per CLAUDE.md: "Shared test fixtures live in the **root `service/conftest.py`** — this is the single place to look for (and add) fixtures. Before adding a fixture to a per-app conftest, check whether it belongs in the root conftest instead."

These four factories (`player_run_active_game_factory`, `player_run_pending_game_factory`, `game_master_game_factory`, `management_transfer_sandbox_game_factory`) follow the `*_factory` naming convention and create game/member scenarios that are just as reusable as `game_factory` and `member_factory` in the root conftest. They should live there instead.

Per-app conftest files are reserved for fixtures that patch app-internal behaviour (e.g. `login/conftest.py` patches Google/Apple auth) — not for scenario factories that other test files could reasonably need.

---

service/member/views.py
```diff
@@ -56,9 +56,11 @@ def get_object(self):
 
     def perform_destroy(self, instance):
         game = instance.game
+        kicked_user = instance.user
         user_id = instance.user_id
         with transaction.atomic():
             instance.delete()
+            game.transfer_management_if_needed(kicked_user, reason="was kicked")
```
johnpooch, line 62: This call to `transfer_management_if_needed` is unreachable dead code. The managing member can never be the one being kicked:

1. `MemberKickView` requires `IsGameManager` — only the current manager can kick.
2. `get_object` blocks self-kicks (`member.user == self.request.user`).
3. Therefore the manager (`kicked_user`) can never equal the manager — the check `current_manager_id != exiting_user.id` in `transfer_management_if_needed` will always be True, and the function returns immediately without doing anything.

In a GM game the short-circuit at `game.game_master_id is not None` fires first, but the outcome is the same.

The actual functional triggers are the CD and elimination hooks in `phase/models.py`. This line should be removed to avoid misleading future readers into thinking kicks can transfer management.

---

service/game/models.py
```diff
@@ -305,6 +307,54 @@ def create_sandbox(self, user, *, name, variant, **kwargs):
...
 
+    def transfer_management_if_needed(self, game, exiting_user, reason=None):
+        if exiting_user is None:
+            return
+        if game.game_master_id is not None:
+            return
+        if game.sandbox:
+            return
+        current_manager_id = (
+            game.managing_member.user_id
```
johnpooch, line 318: When this method is called from the phase resolution path (`phase.game.transfer_management_if_needed(...)`) or the kick path (`instance.game`), the `game` object is not guaranteed to have `managing_member` select_related. If `game.managing_member_id is not None`, accessing `game.managing_member.user_id` hits the DB.

The API-endpoint path is fine (covered by the `resolve_game` fix in `common/views.py` and the `with_list_data`/`with_retrieve_data` QuerySet updates). But the phase adjudication and kick callers don't go through those paths.

Consider fetching `managing_member` lazily inside the method if it isn't already cached:
```python
if game.managing_member_id is not None:
    if not hasattr(game, '_managing_member_cache') or game._managing_member_cache is None:
        game.managing_member  # load via FK
    current_manager_id = game.managing_member.user_id
```
Or, simpler: add `select_related("game__managing_member")` to the `Member` queryset in `get_object()` of `MemberKickView`, and ensure `phase.game` is fetched with `managing_member` where the phase adjudication path loads it.

---

johnpooch (commented):
The core design is solid — `managing_member` as a separate FK cleanly decouples "who created" from "who currently manages", and the CD/elimination hooks in `phase/models.py` are the right integration points. Three issues to address before merge:

1. **Dead code in `MemberKickView`** (medium): `game.transfer_management_if_needed(kicked_user, ...)` can never transfer management because the manager can't be kicked (only the manager can kick, and self-kicks are blocked). The line should be removed. See inline comment.

2. **Test fixtures in wrong conftest** (convention): The four factories in `service/game/tests/conftest.py` belong in root `service/conftest.py` per the project's documented convention. See inline comment.

3. **Potential extra DB query in phase/kick paths** (minor): When `transfer_management_if_needed` is called from the phase resolution or kick paths, `game.managing_member` may not be select_related, causing an extra query when `managing_member_id` is not None. The API-endpoint path is correctly covered by the `resolve_game` and QuerySet fixes; the phase path is not. See inline comment.

---

service/common/permissions.py
johnpooch: Missing `self.message` here. The game-master branch above it sets a message before returning, but this new branch returns silently. When `managing_member_id` is set and the requesting user is not the managing member, the client gets the generic DRF "permission denied" response instead of something actionable.

```python
if game.managing_member_id is not None:
    self.message = "Only the managing member can perform this action."
    return game.managing_member.user_id == request.user.id
```

---

service/game/models.py
```diff
@@ -305,6 +307,54 @@ def create_sandbox(self, user, *, name, variant, **kwargs):
...
+            game.members.filter(
+                eliminated=False, kicked=False, civil_disorder=False
+            ).exclude(user_id=exiting_user.id).select_related("user__profile", "nation")
+            .filter(user__isnull=False)
+        )
+        if not eligible:
+            return
+
+        def reliability_key(member):
+            stats = get_player_stats(member.user)
```
johnpooch, line 336: `get_player_stats` issues ~5 DB queries per call (multiple `count()` calls against `Member` and `PhaseState`, confirmed in `user_profile/utils.py`). Called once per eligible member via `min(eligible, key=reliability_key)`, this is an N+1 for every transfer — N being the number of surviving members in the game.

The `select_related("user__profile", "nation")` on the `eligible` queryset is fetching profile data but `get_player_stats` doesn't use the profile at all; it runs fresh queries against `Member` and `PhaseState`. The prefetch is unused.

Simplest fix: resolve the `min` outside the function so the scoring is clear, and accept that this is an infrequent admin-path operation. But if this runs during adjudication (which could affect many games simultaneously), a single-query approach would be worth the complexity.

At minimum, document the trade-off with a comment explaining why N queries here is acceptable, or replace the per-member calls with a batch query (e.g. annotate the eligible queryset with NMR/CD counts directly).

---

service/game/tests/test_management_transfer.py
johnpooch, line 1: `TestCanManageWithManagingMember` calls `game.can_manage()` directly. Per the project conventions, tests should assert on observable API outcomes rather than internal model methods. `TestCanManageApiResponse` already covers exactly the same behaviour through the API (`GET /game/<id>/` → `can_manage` field). The unit test class is redundant and can be removed.

---

johnpooch (commented):
Three issues found, one blocking.

**Blocking — missing kick trigger**

The issue and PR description both say management transfers when the manager is *kicked*, but `MemberKickView.perform_destroy` (`service/member/views.py:57–69`) never calls `transfer_management_if_needed`. The hooks are only wired into `_check_civil_disorder` and `_check_eliminations` in `phase/models.py`. A manager who is kicked from a pending game would leave the game unmanageable.

The fix is straightforward — call `game.transfer_management_if_needed(instance.user)` inside `MemberKickView.perform_destroy` before or after the `instance.delete()`, and add a corresponding test (e.g. *when the managing member is kicked, management transfers to the next eligible member*).

**Inline comments below cover:**
- Missing `self.message` in `IsGameManager` when the user is not the managing member (minor)
- N+1 DB queries in `reliability_key` — `get_player_stats` runs ~5 queries per eligible member (performance)
- `TestCanManageWithManagingMember` bypasses the API — redundant with `TestCanManageApiResponse` (minor)

---

service/common/permissions.py
johnpooch, line 173: I think this permissions class is smelly and points at a problem in the design. 

In order to determine whether a player is the game manager, we have to ask the following questions:
1. Is there a Game Master?
2. Is the User the Game Master?
3. Is there a Managing Member ID?
4. Is the User's ID the Managing Member's ID?
5. Is the user a member of the game at all?
6. Does the user have a creator, and is the creator's ID the same as the user's?

Instead I think we should just ask one question:
- Is the user the game manager?

johnpooch: Also note that the message on this class is wrong. 

The revised class might read something like this. 
```
class IsGameManager(BasePermission):
    message = "Only the game manager can perform this action."

    def has_permission(self, request, view):
        game = resolve_game(request, view.kwargs.get("game_id"))
        return game.admin_user == request.user

File: service/game/models.py [hunk 11]
```diff
@@ -305,6 +307,56 @@ def create_sandbox(self, user, *, name, variant, **kwargs):
 
         return game
 
+    def transfer_management_if_needed(self, game, exiting_user, reason=None):
```
johnpooch, line 310: This function is convoluted.
Why would an existing user ever be none? 
Why do we rely on `game.game_master_id` here?
Why would this function ever get called if the game is a sandbox game? 

We are baking in complicated rules about assigning to the next most reliable player. Why not just randomly assign to any eligible player? 

Fundamentally, this function is almost 40 lines long, and it really doesn't need to be.

service/game/models.py
```diff
@@ -370,6 +422,13 @@ class Game(BaseModel):
         blank=True,
         related_name="game_master_games",
     )
+    managing_member = models.ForeignKey(
+        "member.Member",
+        on_delete=models.SET_NULL,
+        null=True,
+        blank=True,
+        related_name="managing_games",
+    )
```
johnpooch, line 431: I don't like the name of this field. We don't need to bake the object type into the name. `manager` Would be fun. Personally, I prefer `admin`.

I think the big problem with how this PR is implemented is that this field is sometimes left null. This field should always be populated.

service/phase/models.py
```diff
@@ -43,7 +43,7 @@ def with_detail_data(self):
         )
 
     def with_adjudication_data(self):
-        return self.select_related("variant", "game").prefetch_related(
+        return self.select_related("variant", "game__managing_member").prefetch_related(
```
johnpooch, line 46: It's not clear to me why we need all of these `select_related` calls throughout this PR.

service/phase/models.py
```diff
@@ -553,6 +557,10 @@ def _check_eliminations(self, previous_phase, new_phase):
 
         game = previous_phase.game
 
+        for member in newly_eliminated:
+            if member.user_id is not None:
+                game.transfer_management_if_needed(member.user, reason="was eliminated")
```
johnpooch, line 562: Do we really need to transfer ownership if the user is eliminated?

service/phase/models.py
```diff
@@ -452,6 +452,10 @@ def _check_civil_disorder(self, phase):
             m.civil_disorder = True
         Member.objects.bulk_update(newly_cd_members, ["civil_disorder"])
 
+        for m in newly_cd_members:
+            if m.user_id is not None:
+                phase.game.transfer_management_if_needed(m.user, reason="entered civil disorder")
```
johnpooch, line 457: Part of me thinks that a signal could be a good way to represent the logic which triggers this check.

johnpooch (commented):
I don't think this PR is especially clean or well scoped. I think it should be abandoned in favor of a revised PR that is more narrow in scope.

I think we should choose the simplest possible modeling option. Things are getting complicated fast, with the game creator, game master, and game manager all being represented differently.

PR #908, opened 2026-06-27
Add LLM bot players: identity, seating, and self-play (walking skeleton)

johnpooch:
## What this PR does

The end-to-end walking skeleton for bot players from [discussion #907](https://github.com/johnpooch/diplicity-react/discussions/907#discussioncomment-17455083): a shared bot can be seated in a game and then plays it to completion, submitting legal orders every phase (movement, retreat, and adjustment) with no NMRs or manual intervention. Order selection is deterministic ("first legal option") — **no LLM reasoning yet**.

This combines the two slices that were reviewed as a stack (former PRs 1 and 2).

### Identity + seating
- New `bot` Django app. A data migration seeds a single shared bot user (`bot@diplicity.com`, "Diplicity Bot") so it exists in dev, CI, and production automatically.
- `include_bot_opponent` on game creation seats the bot as a `Member` at creation and joins it to the public channel.
- Allow-list gating: `BOT_OPPONENT_ALLOWLIST` setting (comma-separated emails, empty by default). The create serializer rejects the flag (400) from off-list users; `UserProfileSerializer` exposes a computed `can_create_bot_games` so the frontend hides the "Include bot opponent" checkbox unless the current user is allowed.

### Trigger + play (thin signals, fat tasks)
- `bot/signals.py` only extracts data and defers tasks; the phase engine never imports `bot`. Guards live in decorators (`bot/decorators.py`):
  - `plan` fires when a phase **becomes ACTIVE** (`@on_phase_activated`, backed by a pre_save status cache, so the opening phase — a PENDING→ACTIVE edit during game start — is caught and later edits are ignored) for games seated with the bot (`@with_bot_members`).
  - `finalize` fires once **all human players** with possible orders have confirmed (`@on_orders_confirmed` + `@when_humans_confirmed`).
- `bot/tasks.py` holds the logic as a sequence of in-process DRF calls (`APIClient`, force-authenticated as the bot), so the bot exercises the same endpoints a human client does rather than touching the phase/game models:
  - `plan(user_id, game_id)` — GET order options → POST the first legal order per source. Does not confirm.
  - `finalize(user_id, game_id)` — checks the already-confirmed state first, then resubmits (idempotent upsert) and confirms. Guards on the confirmed state before the confirm call, since that endpoint toggles and isn't idempotent under retry.
- Adjustment order limit: the create loop stops at the first rejected order rather than attempting one per source, so the bot never pushes past the allowed number of builds/disbands (the create endpoint is the authority on the cap). Follow-up #911 will expose the cap so the bot can check it up front instead.

### Out of scope (later PRs)
LLM order reasoning, negotiation/chat, personas, memory, structured commitments, separate budgets, observability.

## Checklist

- [x] This PR does one thing — the bot walking skeleton
- [x] Tests cover the change

## Tests

- `game/tests/test_bot_opponent.py` — seating, allow-list reject, default-off, the `can_create_bot_games` profile flag.
- `bot/tests.py` — order marshalling; `plan` (creates orders, does not confirm); `finalize` (submits + confirms, and does not double-toggle when already confirmed); the adjustment stop-on-reject limit; and triggers (opening-phase activation defers `plan`, editing an active phase does not re-fire, non-bot games defer nothing, human confirm defers `finalize`).
- `integration/test_bot.py` — end-to-end: bot `plan` → human submits + confirms → bot `finalize` → resolve → next phase active, and the bot can play the following phase.
- A pre-existing order-sensitive notification assertion (set-derived `user_ids`) is made order-insensitive; two query-count assertions gain one query from the bot-membership / pre_save status lookups.
- Full backend suite passes (1658) and the frontend TypeScript build, lint, and CreateGame tests are green.

## Screenshots

The "Include bot opponent" checkbox on the create-game General step (allow-listed user):

Create Game — bot opponent checkbox, desktop

Create Game — bot opponent checkbox, mobile


packages/web/src/api/generated/endpoints.ts
```diff
@@ -360,6 +360,7 @@ export interface GameCreate {
   nmrExtensionsAllowed?: number;
   pressType?: PressTypeEnum;
   minReliability?: MinReliabilityEnum;
+  includeBotOpponent?: boolean;
```
johnpooch, line 363: @timwyse this file is an automatically generated TypeScript file which defines the contract between the front end and the back end.

In this PR, we are adding the `includeBotOpponent` argument to the `GameCreate` request payload. And we are adding the `canCreateBotGames` Argument to the user profile response data.

packages/web/src/mocks/fixtures/users.ts
```diff
@@ -12,6 +12,7 @@ export const currentUserProfile: UserProfile = {
   email: "mock.player@example.com",
   emailNotificationsEnabled: true,
   colorblindMode: null,
+  canCreateBotGames: false,
```
johnpooch, line 15: @timwyse these mock fixtures are used to render the front end independently from the backend. In practice, they allow Claude to render the front end easily and take screenshots so that the pull request has screenshots in its description. 

Generally speaking, we don't have to ask Claude to update these mock files explicitly. It always remembers to do it itself.

packages/web/src/screens/Home/CreateGame.tsx
```diff
@@ -110,6 +111,7 @@ const gameSchema = z.object({
     "reliable_and_new",
     "reliable_only",
   ] as const),
+  includeBotOpponent: z.boolean(),
```
johnpooch, line 114: @timwyse this just adds the `includeBotOpponent` field to the form data on the `CreateGame` screen. It defaults to false.

packages/web/src/screens/Home/CreateGame.tsx
```diff
@@ -408,6 +410,7 @@ const CreateGameForm: React.FC<CreateGameFormProps> = ({
   initialPrivate,
   initialMode,
 }) => {
+  const { data: userProfile } = useUserRetrieveSuspense();
```
johnpooch, line 413: @timwyse we fetch the user's profile so that we can check whether the user is allowed to create bot games.

packages/web/src/screens/Home/CreateGame.tsx
```diff
@@ -435,6 +438,7 @@ const CreateGameForm: React.FC<CreateGameFormProps> = ({
       retreatFrequency: null,
       nmrExtensionsAllowed: "0",
       minReliability: "open",
+      includeBotOpponent: false,
```
johnpooch, line 441: @timwyse setting the default value to false.

packages/web/src/screens/Home/CreateGame.tsx
```diff
@@ -617,6 +621,30 @@ const CreateGameForm: React.FC<CreateGameFormProps> = ({
               )}
             />
 
+            {userProfile.canCreateBotGames && (
```
johnpooch, line 624: @timwyse we optionally render the form field if the user can create bot games.

service/bot/migrations/0001_create_bot_user.py
```diff
@@ -0,0 +1,29 @@
+from django.db import migrations
```
johnpooch, line 1: @timwyse This is a data migration. It ensures that the bot user exists both in our local development environments and in production. It is effectively just a bit of code that updates the data when the service is deployed.

service/bot/apps.py
```diff
@@ -0,0 +1,6 @@
+from django.apps import AppConfig
```
johnpooch, line 1: @timwyse Django boilerplate file

service/bot/utils.py
```diff
@@ -0,0 +1,15 @@
...
+
+def get_bot_user():
+    User = get_user_model()
+    return User.objects.get(email=BOT_USER_EMAIL)
+
+
+def user_can_use_bot_opponent(user):
+    if user is None or not user.email:
+        return False
+    return user.email.lower() in settings.BOT_OPPONENT_ALLOWLIST
```
johnpooch, line 17: @timwyse `BOT_OPPONENT_ALLOWLIST` This is a list of users that are allowed to create bot games. It will include my email, your email, and Joren's email. I will configure the setting in Railway, which is where our service is deployed.

service/game/tests/test_bot_opponent.py
```diff
@@ -0,0 +1,95 @@
+import pytest
```
johnpooch, line 1: @timwyse these are unit tests that Claude added to cover the new behavior. Note that Claude almost always adds tests as required.

service/game/serializers.py
```diff
@@ -443,6 +449,13 @@ def validate_variant_id(self, value):
             raise serializers.ValidationError("Variant with this ID does not exist.")
         return value
 
+    def validate_include_bot_opponent(self, value):
+        if value and not user_can_use_bot_opponent(self.context["request"].user):
+            raise serializers.ValidationError(
+                "You are not permitted to add a bot opponent."
```
johnpooch, line 455: @timwyse in practice, this should never happen because the field is not rendered for users without permission, but this is good defensive programming.

service/project/settings.py
```diff
@@ -251,6 +252,12 @@
 
 FRONTEND_URL = os.getenv("FRONTEND_URL", "https://diplicity.com")
 
+BOT_OPPONENT_ALLOWLIST = [
```
johnpooch, line 255: @timwyse this converts the value that I set on Railway into a list of emails.

service/openapi-schema.yaml
```diff
@@ -2005,6 +2005,10 @@ components:
           allOf:
           - $ref: '#/components/schemas/MinReliabilityEnum'
           default: open
+        includeBotOpponent:
```
johnpooch, line 2008: @timwyse This is an auto-generated open API schema that is used to create the TypeScript API contract above.

johnpooch (commented):
@timwyse I've left a bunch of comments here as little signposts to help you build up a map of the service. But I imagine this would be very disorienting. 
Over time, it would make more sense. 

But to be honest, once this plubming work is in place, most of the AI-specific stuff will be spread across a small number of files in the `/bots/` subfolder. So don't worry about it if you can't really follow what's going on.

service/bot/constants.py
```diff
@@ -0,0 +1,3 @@
+BOT_USER_EMAIL = "bot@diplicity.com"
```
johnpooch, line 1: @timwyse Hard coding the single bot user's details here for now. This is a temporary thing while we only have one bot user.

service/bot/decorators.py
```diff
@@ -0,0 +1,98 @@
+import functools
```
johnpooch, line 1: @timwyse this is just a technical thing, but I'm using function decorators to represent checks that we apply to the signal functions to determine whether to trigger a task. 

In plain English, Django's signals are functions which you can define, which are invoked when something happens elsewhere in the system, e.g., a phase is saved.

Tasks are background tasks. These are bits of functionality that you want to run separately from the main thread. For example, when a new phase is created, we want the bots to do some work and create orders, etc. We do not want to block the main thread which creates the phase on this slow and fragile work. So instead, we create the phase and queue a task, which then lets a background worker pick up that work. 

Basically:
- Signal: Listen for some event and call a function when it happens
- Task: Background work that should not be handled on the main thread (typically slow work, or work that depends on a separate service, e.g. calling Anthropic)

service/bot/tasks.py
```diff
@@ -0,0 +1,77 @@
...
+from procrastinate.contrib.django import app
+from rest_framework.test import APIClient
+
+from bot.utils import first_legal_selections
+
+logger = logging.getLogger(__name__)
+
+
+@app.task(name="bot.plan", retry=3)
+def plan(user_id, game_id):
```
johnpooch, line 14: @timwyse this file is the meat of the bot functionality so far.

We have two functions: plan and finalize. 

Plan runs at the beginning of the new phase. 

Finalize runs when all human users have confirmed their orders. 

In the future, we will need to make finalize also run shortly before the phase deadline. 

Some important things to note about how this file is written:
- The bot uses "API client" to interact with the rest of the service. This is an intentional decision to make sure that our bot only has access to the information that a normal user 
- There is a decent bit of logging here, which will be helpful during early experiments. If something isn't working, we'll be able to find out where it failed. 

Eventually, these tasks will call the LLM before creating the orders, etc.

service/bot/utils.py
```diff
@@ -0,0 +1,45 @@
...
+
+
+def option_to_selected(option):
```
johnpooch, line 20: @timwyse the two functions here are basically just data marshalling functions to extract the "quote options" and create an order using the first legal move. 

Basically, marshaling options into orders.

service/integration/test_bot.py
```diff
@@ -0,0 +1,97 @@
+from unittest.mock import patch
```
johnpooch, line 1: @timwyse The `service/integration` App is where we define integration tests. These tests are useful for testing the logical flow of the service. They define tests with multiple steps:
- User creates a game
- User adds a bot
- Game starts
- User submits orders and confirms
- Assert bot submits orders and confirms

etc.

PR #914, opened 2026-06-27
Expose adjustment-phase order limit in the API

### LINKED ISSUE #911: Expose adjustment-phase order limit in the API
## Goal

Expose the per-player adjustment-phase order limit as a value in the API, so the frontend and the bot can check it up front instead of discovering it reactively.

## Context

`PhaseState.max_allowed_adjustment_orders()` (`service/phase/models.py:1005`) already computes the limit (`abs(supply_centers − units)`) but it is exposed in no serializer. Today it is only enforced reactively — `orderable_provinces` shrinks once the cap is hit (`phase/models.py:1028-1034`) and the order-create endpoint rejects extras (`order/serializers.py:101`) — so the order wizard shows every home centre as buildable until the allowance is used up (no "N builds remaining" indicator), and the bot has to submit-until-rejected because it reads the order-options endpoint, which never shrinks.

## Approach

Add a `max_orders` `SerializerMethodField` to `PhaseStateSerializer` (`service/phase/serializers.py`), annotated `@extend_schema_field(serializers.IntegerField(allow_null=True))`, returning `max_allowed_adjustment_orders()` for adjustment phases and `null` otherwise (the method returns `inf` for other phase types, which is not JSON-clean, and the concept only applies to adjustment). Regenerate the API client and update the `PhaseState` mock fixtures, surface "N builds/disbands remaining" in the order wizard and gate it on the value, and switch the bot's adjustment handling from stop-on-reject to an upfront cap (read `maxOrders` from `phase-state-list` and slice the selections).

johnpooch:
Closes #911

## What

Adds `max_orders` to `PhaseStateSerializer` so the API returns the backend-computed adjustment-phase order cap per player. The field is an integer (or `null` for non-adjustment phases).

`PhaseGuidance` on the frontend now reads `userPhaseState.maxOrders` from the API response instead of recomputing the cap locally. The fallback (`?? Math.abs(delta)`) is retained for safety but should never be reached in practice.

## Changes

- **Backend**: `PhaseStateSerializer` gains a `max_orders` `SerializerMethodField` annotated with `@extend_schema_field`. For adjustment phases it calls the existing `max_allowed_adjustment_orders()` model method; for all other phases it returns `None`.
- **API schema / generated client**: `maxOrders: number | null` added to both `PhaseState` and `PatchedPhaseState`.
- **Mock fixtures**: `makePhaseState` defaults `maxOrders` to `null`; the `activeGameBuild` adjustment-phase fixture sets accurate per-nation values (`1` for England, `0` for all others). Inline objects in `legacy.ts` updated to include the new required field.
- **Tests**: Two new API-level tests verify that `max_orders` is `null` for a movement phase and equals the correct integer for an adjustment phase with a known supply-center/unit delta.

---
_Generated by [Claude Code](https://claude.ai/code/session_01AS5kSgghkVHzrsoVdNyu1k)_


johnpooch (commented):
## Review

The overall structure is correct: the new `max_orders` field is properly annotated with `@extend_schema_field`, the OpenAPI schema and generated client are regenerated, mock fixtures are updated, and two API-level tests are added. Two issues worth addressing before merging.

---

### `max_allowed_adjustment_orders()` is called twice for adjustment phases

`service/phase/serializers.py` — `get_max_orders`

```python
def get_max_orders(self, obj):
    if obj.phase.type != PhaseType.ADJUSTMENT:
        return None
    return obj.max_allowed_adjustment_orders()
```

`max_allowed_adjustment_orders()` issues 2 DB queries: `supply_centers.filter(nation=...).count()` and `units.filter(nation=...).count()`. For adjustment phases the `orderable_provinces` field already calls this same method (via `get_orderable_provinces` → `max_allowed_adjustment_orders()`), so the new field adds 2 redundant round-trips per request.

The simplest fix is to memoize at the model level:

```python
# phase/models.py
@cached_property
def max_allowed_adjustment_orders(self):
    if self.phase.type != PhaseType.ADJUSTMENT:
        return float("inf")
    nation = self.member.nation
    supply_centers_count = self.phase.supply_centers.filter(nation=nation).count()
    units_count = self.phase.units.filter(nation=nation).count()
    return abs(supply_centers_count - units_count)
```

Call sites in `get_orderable_provinces` and `get_max_orders` then use `obj.max_allowed_adjustment_orders` (property access) and the result is cached on first read. Note: converting a method to a `cached_property` also removes the parentheses at all call sites — `get_orderable_provinces` uses `self.max_allowed_adjustment_orders()` and `order/serializers.py:101` also references it — both would need the `()` removed.

---

### No query count test for `phase-state-list`

`service/phase/tests.py` — new tests `test_max_orders_in_api_response_non_adjustment` / `test_max_orders_in_api_response_adjustment`

The two new tests verify the response value but not query count. Other endpoints have query count tests (e.g. `test_list_games_query_count_with_phase_states`, `test_order_create_query_count_with_many_phase_states`, `test_list_phases_query_count`). A test like the following would prevent future regressions:

```python
@pytest.mark.django_db
def test_phase_state_list_query_count(
    self,
    django_assert_num_queries,
    authenticated_client,
    active_game_with_phase_options,
):
    url = reverse("phase-state-list", args=[active_game_with_phase_options.id])
    with django_assert_num_queries(N):
        authenticated_client.get(url)
```

(Replace `N` with the baseline count after the memoization fix above is applied.)

---

PR #915, opened 2026-06-27
Expose adjustment-phase order limit in the API

### LINKED ISSUE #911: Expose adjustment-phase order limit in the API

johnpooch:
Closes #911

## What this does

Exposes `PhaseState.max_allowed_adjustment_orders()` as `max_orders` in the phase-state API endpoint. Returns `null` for non-adjustment phases and the computed limit (`abs(supply_centers - units)`) for adjustment phases.

### Backend (`service/phase/serializers.py`)

Added a `max_orders` `SerializerMethodField` to `PhaseStateSerializer` with `@extend_schema_field(serializers.IntegerField(allow_null=True))`. Returns `None` for non-adjustment phases and calls `obj.max_allowed_adjustment_orders()` for adjustment phases.

### OpenAPI + client codegen

Regenerated `service/openapi-schema.yaml` and `packages/web/src/api/generated/endpoints.ts`. The `PhaseState` type now includes `maxOrders: number | null`.

### Frontend (`PhaseGuidance.tsx`)

Changed the Adjustment case to use `userPhaseState.maxOrders ?? Math.abs(delta)` so the guidance text reflects the API-authoritative limit when available.

### Bot (`service/bot/tasks.py`)

Both `plan` and `finalize` tasks now call a shared `_get_max_orders` helper that fetches the phase-state list and reads `max_orders`. The selection generator is capped upfront via `itertools.islice` so the bot never attempts to submit more orders than allowed, without relying on stop-on-reject.

### Mock fixtures

Added `maxOrders: null` to `makePhaseState` in `builders.ts` and to all 7 inline `PhaseState` objects in `legacy.ts`.

## No visual changes

The guidance text for adjustment phases is unchanged in practice for the existing fixtures (which have `maxOrders: null`, falling back to the delta calculation). The field is now available for future UI use.

---
_Generated by [Claude Code](https://claude.ai/code/session_01Df5J6728dQDFWGdHhXhNgj)_


johnpooch:
Closing as already-landed. The `max_orders` feature this PR introduces is already in `main` — the same change was re-landed independently as commit `8c7c429` ("Expose adjustment-phase order limit in the API (#911)"), which is now an ancestor of `main`.

Verified that every part of this PR is present in `main`:
- `max_orders` field in `phase/serializers.py` (byte-identical)
- `PhaseGuidance.tsx` → `maxOrders ?? Math.abs(delta)`
- `maxOrders: null` in mock builders/fixtures
- `maxOrders` in generated `endpoints.ts` + `openapi-schema.yaml`
- Bot order-limit capping (`get_max_orders`), now in refactored form
- Phase tests for `max_orders`

Merging `main` into this branch and resolving the conflicts leaves no relevant changes — only stale codegen noise (re-adding `apiTestSentryRetrieve`/`FCMDevice`, env-dependent enum casing) and duplicate `maxOrders: null` keys/tests. So there's nothing left for this PR to contribute.

---

PR #917, opened 2026-06-27
Cap bot adjustment orders up front using the API order limit

johnpooch:
## What this PR does

The bot follow-up enabled by #911 / #914: now that the adjustment order limit is exposed on the phase-state endpoint as `max_orders`, the bot checks it **up front** instead of submitting one order per source and stopping at the first rejection.

In both `plan` and `finalize`, after fetching the order options the bot fetches its own phase state from the `phase-state-list` endpoint and, when `max_orders` is set (adjustment phases), slices the selections to that many before creating them. For non-adjustment phases `max_orders` is `null`, so every legal order is submitted. The create loop now logs a failure rather than treating a rejection as "limit reached", since the count is known before submitting.

This replaces the stop-on-reject behaviour from the bot skeleton, which was a deliberate stopgap pending the exposed count.

## Checklist

- [x] This PR does one thing — use the API order limit to cap bot orders
- [x] Tests cover the change
- [x] Screenshots embedded in the PR description for any visual changes — **N/A, no visual changes** (backend-only)

## Tests

`bot/tests.py` `TestAdjustmentOrderLimit` is reworked to verify the cap: with `max_orders: 1` and three legal build options the bot creates exactly one order; with `max_orders: null` it creates all three. The existing in-process `plan`/`finalize` tests still pass (movement phases report `max_orders: null`, so nothing is capped). Full backend suite passes (1663).

---
_Generated by [Claude Code](https://claude.ai/code/session_01WHr91Pe52JxFpEP5edLh6q)_

### DIFF
--- a/service/bot/tasks.py
+++ b/service/bot/tasks.py
@@ -23,16 +23,21 @@ def plan(user_id, game_id):
     if options_response.status_code != 200:
         logger.error(f"[{label}] options request failed: {options_response.status_code}")
         return
+    selections = first_legal_selections(options_response.data["orders"])
+
+    phase_states_response = client.get(reverse("phase-state-list", args=[game_id]))
+    if phase_states_response.status_code != 200:
+        logger.error(f"[{label}] phase states request failed: {phase_states_response.status_code}")
+        return
+    max_orders = phase_states_response.data[0]["max_orders"] if phase_states_response.data else None
+    if max_orders is not None:
+        selections = selections[:max_orders]
 
     create_url = reverse("order-create", args=[game_id])
-    for selected in first_legal_selections(options_response.data["orders"]):
+    for selected in selections:
         response = client.post(create_url, {"selected": selected}, format="json")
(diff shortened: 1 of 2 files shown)


service/bot/tasks.py
```diff
@@ -23,16 +23,21 @@ def plan(user_id, game_id):
     if options_response.status_code != 200:
         logger.error(f"[{label}] options request failed: {options_response.status_code}")
         return
+    selections = first_legal_selections(options_response.data["orders"])
+
+    phase_states_response = client.get(reverse("phase-state-list", args=[game_id]))
+    if phase_states_response.status_code != 200:
+        logger.error(f"[{label}] phase states request failed: {phase_states_response.status_code}")
+        return
+    max_orders = phase_states_response.data[0]["max_orders"] if phase_states_response.data else None
```
johnpooch, line 32: Why would `phase_states_response.data` be undefined here?

PR #924, opened 2026-06-29
Refactor bot LLM layer into actions, client, and DTOs

johnpooch:
## What this PR does

Restructures the `bot` app's LLM code into a small object-oriented layer so the moving parts are isolated and the prompts/tool schemas are easy to find and iterate on. `bot/llm.py` is split apart; behaviour is unchanged.

New structure:

- **`bot/actions/`** — an `Action` ABC plus `SelectOrderAction` and `ReplyAction`. Each action receives its data at construction and owns its tool schema, prompt template, response parsing, and fallback. The prompt for each action lives at the top of its file, next to its tool and parse logic.
- **`bot/llm_client.py`** — `LLMClient` encapsulates the Anthropic plumbing and the repeated policy bits (no `ANTHROPIC_API_KEY` → fallback; LLM call errors → fallback). `run(action)` dispatches any `Action` without knowing which one it is.
- **`bot/api_client.py`** — `BotApiClient` wraps the DRF `APIClient`, returns DTOs, and raises `BotApiError` on unexpected status so tasks collapse the repeated `if status != 200` checks into one `try`/`except`.
- **`bot/dto.py`** — `OrderOption`, `OrderOptionCollection`, and `ChatMessage` wrap the raw API dicts and own their marshaling (`from_api`) and access logic. The order-option helpers (`option_to_selected`, `first_legal_selections`) moved here from `utils.py`.

`tasks.py` is now thin orchestration: gather data via the API client, build an action, hand it to the LLM client. `utils.py` keeps only the bot-identity/config helpers (`get_bot_user`, `user_can_use_bot_opponent`, `bot_request_host`).

Marshaling lives on the DTOs (`from_api`) rather than being injected into the API client, and prompts live inside each action file — both per discussion before implementing.

## Checklist

- [x] For PRs of any significant complexity: I ran `/review-pr` against this PR in Claude Code and addressed (or responded to) its findings — note: the `/review-pr` skill shells out to `gh`, which isn't wired up in this environment, so I self-reviewed the diff manually and removed one piece of dead code (`OrderOptionCollection.__len__`) as a result.
- [x] Tests cover the change
- [x] Screenshots embedded in the PR description for any visual changes (see CLAUDE.md) — N/A, backend-only refactor with no visual changes

Tests: `bot/tests.py` (34) and `integration/test_bot.py` (2) updated for the new structure and passing; `game/tests/test_bot_opponent.py` (8) and `manage.py check` also green.


service/bot/actions/reply.py
```diff
@@ -0,0 +1,57 @@
...
+        "properties": {
+            "should_reply": {"type": "boolean"},
+            "message": {"type": "string"},
+        },
+        "required": ["should_reply", "message"],
+        "additionalProperties": False,
+    },
+}
+
+PROMPT_PREAMBLE = (
```
johnpooch: I'm curious about this. Should this not be the system prompt and have the system role?

service/bot/actions/reply.py
```diff
@@ -0,0 +1,57 @@
...
+class ReplyAction(Action):
+    name = "reply"
+    tool = TOOL
+
+    def __init__(self, messages):
+        self.messages = messages
+
+    def build_prompt(self):
+        lines = [f"{message.speaker}: {message.body}" for message in self.messages]
+        return PROMPT_PREAMBLE + "\n\n" + "\n".join(lines)
```
johnpooch: I feel like the function should be called build_messages. 

It should return a list of messages ready for the LLM client. 

I think that the chat messages in this particular action should have the assistant role when they are sent by the bot itself. I have the user role and it's sent by any other user.

service/bot/actions/select_order.py
```diff
@@ -0,0 +1,77 @@
...
+                    "additionalProperties": False,
+                },
+            }
+        },
+        "required": ["choices"],
+        "additionalProperties": False,
+    },
+}
+
+PROMPT_PREAMBLE = (
```
johnpooch: I feel like this should also be a system prompt, right?

service/bot/llm_client.py
```diff
@@ -0,0 +1,38 @@
...
+
+from anthropic import Anthropic
+from django.conf import settings
+
+logger = logging.getLogger(__name__)
+
+
+class LLMClient:
+    def run(self, action):
+        if not settings.ANTHROPIC_API_KEY:
```
johnpooch: Should the Anthropic key not get passed in at class initialization? 

Should we throw an error if there is no Anthropic API key? 

Then the calling code can handle the exception gracefully.

service/bot/actions/reply.py
```diff
@@ -0,0 +1,57 @@
...
+
+    def parse(self, tool_input):
```
johnpooch, line 46: Removed `fallback()` from the actions. I didn't lift it into the LLM client, though, because the two fallbacks aren't uniform: `select_order` falls back to first-legal selections (which needs the `OrderOptionCollection`), while `reply` just stays silent. A generic client can't produce a first-legal selection without knowing about order options.

Instead `LLMClient.run()` now raises `LLMError` (on a missing key or a failed request), and the **task** owns the fallback — `plan`/`finalize` catch it and use `options.first_legal_selections()`, `reply` catches it and stays silent. That keeps the boilerplate method off the actions and matches the "calling code handles the exception" approach from your `llm_client.py` comment.

---

PR #945, opened 2026-07-01
Add BotProfile model and identify bots via attribute check

### LINKED ISSUE #944: Add BotProfile model and identify bots by attribute
## Goal

Add a `BotProfile` model that carries persona details (disposition and voice) for bot users, and switch bot identification from the magic email `bot@diplicity.com` to an attribute check (the user has a bot profile).

## Context

Supporting multiple bots per game (e.g. six bots in a classical game) requires a pool of distinct persistent bot users rather than today's single shared bot user, which is identified everywhere by the hardcoded email in `service/bot/constants.py`. Introducing `BotProfile` is the foundational modelling step: it gives each bot a persona (per the design doc at `service/bot/docs/TECHNICAL_DESIGN.md`) and replaces the brittle email-string match used throughout `service/bot/decorators.py` and `service/bot/utils.py`.

## Approach

Add the `BotProfile` model with persona fields (disposition, voice), related to the bot `User`. Replace the `user__email == BOT_USER_EMAIL` / `user__email=BOT_USER_EMAIL` checks in `service/bot/decorators.py` and `service/bot/utils.py` (`get_bot_user`) with a bot-profile attribute check, and back-fill a `BotProfile` for the existing bot user in a migration. Scope is limited to the model, persona fields, and the identification change — adding more bot users to the pool, the add-bot endpoint, and the staging-screen "Add Player" UI (Option B) are follow-up issues.

johnpooch:
## What this PR does

Adds a `BotProfile` model (`disposition`, `voice`) as the foundational persona model for bot users, and switches bot identification off the hardcoded `bot@diplicity.com` email onto a `bot_profile` attribute check. `service/bot/decorators.py` and `service/bot/utils.py` (`get_bot_user`) now check `user__bot_profile__isnull=False` / `hasattr(user, "bot_profile")` instead of matching the magic email string. A migration (`bot/0002_create_bot_profile`) creates the model and backfills a `BotProfile` for the existing bot user.

The now-unused `BOT_USER_EMAIL` constant was removed from `bot/constants.py` since nothing references it outside the frozen `0001` migration (which already hardcodes its own copy, per the existing pattern).

Scope is limited to the model and the identification switch, per the issue — adding more bot users to the pool, the add-bot endpoint, and staging-screen UI are follow-up issues.

Closes #944

## Checklist

- [x] Tests cover the change
- [x] No visual changes — backend-only model/migration change.

---
_Generated by [Claude Code](https://claude.ai/code/session_01WSeoTVEDe2mCnQDwx13N4J)_

### DIFF
--- a/service/bot/decorators.py
+++ b/service/bot/decorators.py
@@ -57,8 +57,7 @@ def on_public_chat_message(func):
             return
         if instance.channel.private:
             return
-        message_sender = instance.sender
-        if message_sender.user is not None and hasattr(message_sender.user, "bot_profile"):
+        if Member.objects.filter(pk=instance.sender_id, user__bot_profile__isnull=False).exists():
             return
         return func(sender=sender, instance=instance, created=created, **kwargs)
 
--- a/service/bot/decorators.py
+++ b/service/bot/decorators.py
@@ -4,15 +4,13 @@ from common.constants import PhaseStatus
 from member.models import Member
 from phase.models import Phase
 
-from bot.constants import BOT_USER_EMAIL
-
 _previous_phase_status = {}
 
 
 def _bot_user_ids_for_phase(phase_id):
     return set(
         Member.objects.filter(
-            game__phases=phase_id, user__email=BOT_USER_EMAIL
+            game__phases=phase_id, user__bot_profile__isnull=False
         ).values_list("user_id", flat=True)
     )
 
@@ -42,7 +40,7 @@ def with_bot_members(func):
     def wrapper(sender, instance, **kwargs):
         bot_members = list(
             Member.objects.filter(
-                game_id=instance.game_id, user__email=BOT_USER_EMAIL
+                game_id=instance.game_id, user__bot_profile__isnull=False
             ).select_related("user")
         )
         if not bot_members:
@@ -60,7 +58,7 @@ def on_public_chat_message(func):
         if instance.channel.private:
             return
         message_sender = instance.sender
-        if message_sender.user is not None and message_sender.user.email == BOT_USER_EMAIL:
+        if message_sender.user is not None and hasattr(message_sender.user, "bot_profile"):
             return
         return func(sender=sender, instance=instance, created=created, **kwargs)
 
@@ -71,7 +69,9 @@ def with_bot_channel_members(func):
     @functools.wraps(func)
     def wrapper(sender, instance, **kwargs):
         bot_members = list(
-            instance.channel.members.filter(user__email=BOT_USER_EMAIL).select_related("user")
+            instance.channel.members.filter(
+                user__bot_profile__isnull=False
+            ).select_related("user")
         )
         if not bot_members:
             return
--- a/service/bot/migrations/0002_create_bot_profile.py
+++ b/service/bot/migrations/0002_create_bot_profile.py
@@ -1,0 +1,55 @@
+import django.db.models.deletion
+from django.conf import settings
+from django.db import migrations, models
+
+BOT_USER_EMAIL = "bot@diplicity.com"
+BOT_DISPOSITION = "Balanced and pragmatic; willing to ally or betray as the board demands."
+BOT_VOICE = "Plain and direct."
+
+
+def create_bot_profile(apps, schema_editor):
+    User = apps.get_model("auth", "User")
+    BotProfile = apps.get_model("bot", "BotProfile")
+
+    user = User.objects.filter(email=BOT_USER_EMAIL).first()
+    if user is None:
+        return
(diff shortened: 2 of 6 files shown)


johnpooch (commented):
### PR Review

Self-review via `/review-pr` (gh CLI unavailable in this environment; performed manually using the GitHub MCP tools and the same rubric).

Adds `BotProfile` (disposition, voice) and switches bot identification from the `bot@diplicity.com` email match to a `bot_profile` attribute check in `decorators.py`/`utils.py`, with a migration backfilling the profile for the existing bot user. Traced all four identification call sites (`_bot_user_ids_for_phase`, `with_bot_members`, `on_public_chat_message`, `with_bot_channel_members`) plus `get_bot_user()` — all correctly switched to the new attribute, and `hasattr(user, "bot_profile")` works as intended because Django's reverse-`OneToOneField` `RelatedObjectDoesNotExist` inherits from `AttributeError`. Confirmed `BOT_USER_EMAIL` has no remaining references outside the frozen `0001` migration (which already hardcodes its own copy) before removing it from `constants.py`. Ran the full `bot/tests.py` suite (45 passed) and added two tests proving identification no longer depends on the magic email. No frontend or other backend app references bot identification, so scope is fully contained.

---

service/bot/migrations/0002_create_bot_profile.py
```diff
@@ -0,0 +1,55 @@
...
+        return
+    BotProfile.objects.get_or_create(
+        user=user,
+        defaults={"disposition": BOT_DISPOSITION, "voice": BOT_VOICE},
+    )
+
+
+class Migration(migrations.Migration):
+
+    initial = True
```
johnpooch, line 25: 💡 **suggestion**: `initial = True` here doesn't match how Django would actually generate this migration — the autodetector only sets `initial` when the app has no prior migrations (`app_label not in self.existing_apps`), and `bot` already has `0001_create_bot_user`. No other non-`0001` migration in the codebase sets this flag; it looks like a manual addition and could confuse `--fake-initial` detection down the line. Consider removing it.

---

service/bot/decorators.py
```diff
@@ -60,7 +58,7 @@ def wrapper(sender, instance, created, **kwargs):
         if instance.channel.private:
             return
         message_sender = instance.sender
-        if message_sender.user is not None and message_sender.user.email == BOT_USER_EMAIL:
+        if message_sender.user is not None and hasattr(message_sender.user, "bot_profile"):
```
johnpooch: 💡 **suggestion**: `hasattr(message_sender.user, "bot_profile")` triggers an extra DB query per public chat message, since the reverse OneToOne isn't prefetched here (the old `.email` check was free — `email` was already loaded on `user`). This fires on every public message send; consider `Member.objects.filter(pk=message_sender.pk, user__bot_profile__isnull=False).exists()` to avoid the added round trip.

---

johnpooch (commented):
### PR Review

Adds a `BotProfile` model and switches bot identification from the hardcoded email to a `bot_profile` attribute check across `decorators.py`, `utils.py`, and the queryset filters. The core change is correct — the model matches the migration (`makemigrations --check` shows no drift), the migration runs cleanly, and the old `.email` checks are fully replaced with no dangling references to the removed `BOT_USER_EMAIL` constant. Two minor points below; neither blocks merging.

**2** inline comment(s) (0 blocking, 2 suggestion).

---

service/bot/migrations/0002_create_bot_profile.py
johnpooch: Verified this empirically rather than taking it on faith: `initial` is set by `app_label not in self.existing_apps` where `existing_apps` comes from models already present in migration state (`autodetector.py:60`), not "has any prior migration." `bot/0001_create_bot_user.py` only does a `RunPython` — it defines no models for the `bot` app — so `bot` genuinely has zero models before `0002`, and Django's autodetector sets `initial = True` on its own.

I confirmed by deleting the migration and regenerating it from scratch with `makemigrations bot --name create_bot_profile`: Django reproduces `initial = True` unprompted. `makemigrations bot --check --dry-run` also reports no drift with the flag in place. Removing it would make the committed migration diverge from what Django actually generates, so I've left it as-is.

---

service/bot/decorators.py
johnpooch: Fixed in 19c6e31 — replaced with `Member.objects.filter(pk=instance.sender_id, user__bot_profile__isnull=False).exists()`, using `instance.sender_id` so it doesn't even need to fetch the `Member`/`User` rows first. Down to a single query. Tests still pass (45/45).

---

PR #947, opened 2026-07-02
Add logged-out web landing showing open games

johnpooch:
## What this PR does

When a logged-out user opens the app in a **web browser**, they now land on a new **Open Games** screen (joinable pending games) instead of the login wall, with a Bluesky-style login panel replacing the nav. Clicking into a game opens the read-only **Game Info** / **Player Info** screens, and the join CTA becomes **"Sign in to join"** (deep-linking back to the game after login). **Native apps are unchanged** — iOS/Android still land on the login screen.

This is the frontend half of the logged-out experience; the backend read endpoints (variant list, phase retrieve) were opened to anonymous access in #941 (merged).

### Changes

- **`Router.tsx`** — the logged-out branch now splits on `isNativePlatform()`: native keeps today's tree (Login at `/`); web gets a `LoggedOutLayout` wrapping Open Games (index), Game Info, and Player Info, with Login moved to `/login`. The authenticated root gains redirect routes for the auth paths (`/login → /`) so the post-login URL resolves, and the web catch-all stores a deep link then routes to `/login`.
- **`LoggedOutLayout.tsx`** (new) — logo, a "Welcome to Diplicity" heading, and side-by-side **Create account** / **Sign in** controls on desktop; a persistent bottom login bar on mobile.
- **`OpenGames.tsx`** (new) — dedicated screen querying `status=pending` (no auth-only `can_join`/`eligible_only` filters, no filter UI), reusing `GameCard`.
- **`GameInfo.tsx`** — adds a "Sign in to join" branch (stores a deep link, routes to `/login`) when logged out.
- **`UserAvatar.tsx`** — renders nothing when there is no access token, so shared screens (Game Info / Player Info) don't fire the authenticated user request and trip their error boundary. Guarded via `tokenStorage.isLoggedIn()` (the same primitive `useAuth().loggedIn` derives from).

The `Login` screen is unmodified.

## Screenshots

Logged-out **Open Games** landing (desktop):

open games desktop

Logged-out **Open Games** (mobile, with the persistent login bar):

open games mobile

Logged-out **Game Info** with the "Sign in to join" CTA:

game info desktop

## Checklist

- [x] Tests cover the change

### Tests

- New `OpenGames.test.tsx` — asserts `status: "pending"` query, absence of `can_join`/`eligible_only`, card rendering, unknown-variant filtering, and the empty state.
- New `LoggedOutLayout.test.tsx` — Create account / Sign in navigation.
- Extended `GameInfo.test.tsx` — the "Sign in to join" branch and deep-link storage when logged out.
- Full frontend suite green (481 passed), `tsc -b --noEmit` clean, lint clean.

---
_Generated by [Claude Code](https://claude.ai/code/session_01SmzcTGbL5y2XyVZ2odfnQs)_

### DIFF
--- a/packages/web/src/Router.tsx
+++ b/packages/web/src/Router.tsx
@@ -17,10 +17,12 @@ import { GameDetail, Home, Tutorial, Variants } from "./screens";
 const CardGallery = React.lazy(() => import("./screens/CardGallery"));
 import { ErrorFallbackUI } from "./components/ErrorBoundary";
 import { HomeLayout } from "./components/HomeLayout";
+import { LoggedOutLayout } from "./components/LoggedOutLayout";
 import { GameDetailLayout } from "./components/GameDetailLayout";
 import { GamePhaseRedirect } from "./components/GamePhaseRedirect";
 import { useIsMobile } from "./hooks/use-mobile";
 import { getVariantsListQueryOptions } from "./api/generated/endpoints";
+import { isNativePlatform } from "./utils/platform";
 import * as Sentry from "@sentry/react";
 import { deepLinkStorage, useDeepLink } from "./deepLink";
 
@@ -51,6 +53,14 @@ const HomeLayoutWrapper: React.FC = () => {
     <HomeLayout>
       <Outlet />
     </HomeLayout>
+  );
+};
+
+const LoggedOutLayoutWrapper: React.FC = () => {
+  return (
+    <LoggedOutLayout>
+      <Outlet />
+    </LoggedOutLayout>
   );
 };
 
@@ -308,10 +318,22 @@ const Router: React.FC<RouterProps> = ({ loggedIn, queryClient }) => {
                     },
                   ],
                 },
+                ...[
+                  "login",
+                  "register",
+                  "forgot-password",
+                  "check-email",
+                  "verify-email",
+                  "reset-password",
+                ].map(path => ({
+                  path,
+                  loader: () => redirect("/"),
+                })),
               ],
             },
           ])
-        : createBrowserRouter([
+        : isNativePlatform()
+        ? createBrowserRouter([
             {
               path: "/",
               element: <Login />,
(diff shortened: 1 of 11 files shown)


johnpooch:
@JorenC could you take a look at the staging environment of this? I think this is a good change. I want people who come to the website to quickly see the website in action rather than a marketing screen initially. 

When we enable "spectator" mode on the game detail screen and move pending games to the game detail screen, it will be even more engaging.

JorenC:
I did not get notified of your comment, so only see it now. Apologies. I think your idea is good, but the implementation needs a lot of changes. 

1. The core function of the website is to make sure new users understand what this is - and see the value proposition. Right now, you are just 'dumping' the games for them. While the idea I still understand, this is good for _one kind of user_ - the one that already knows diplomacy, knows diplicity is a web implementation like others, and wants to join. For them, it's a good, easy funnel into a game.

However, users:
- That do now know diplomacy
- That know dip but do not know anything about web versions online
- That are here for/through the Android/iOS version
- That know dip, and know other platforms, but are looking to see why Diplicity might be the one for them to choose (out of many)
Are all under- or not served.

2. While the desktop implementation is okay-ish, the mobile implementation is not. I just come on a home screen with title 'Open games' (Where am I? What is this? What game? Title of the website?). There is no context whatsoever. Even the 'sign up' and 'log in' don't even know what I'm signing up for.
This should _at least_ have an app bar/top bar with the name, what it is and then the login/sign up. Then, _underneath_ that all you could make an open games section. 

But in both cases, even that 'Open Games' does not give context - open games? What? If anything, that subtitle needs more context "Games that are currently looking for players" or something. This is pre-context, so it needs more explanation.

------------------------------

Overall, while I do see the value you want to give, the problem is I disagree with your (complete) approach. I do not think we should become **completely open** because this is a niche site that not many people know *and it needs more context*. 

Even X, which I think you referenced, has a full-page blocker:  (continued below)
image
image

I think that while I see your approach, making the website open as-if-the-player-already signed up is not the right approach. 

Would it not be better if we add a - specifically designed - section to funnel people in? Why do we need to show the 'open games' interface?

Example (which I prefer):
- We can show a third column on desktop with a scrollable window/'iFrame') that the user can see all staging games _while keeping context_. This both showcases the context, Android AND ongoing games. 
image

For desktop, we could use a carousel of games in the scroll - for larger devices even in the viewport, that would have a 'see all' that would load a new page (back to home with login/sign up direct buttons on the top) with all open games in there:
image


This way, we keep context for new users, we can have a space (which we will need to redo later anyway) with more value proposition (e.g. number of variants, variant uploaders etc), AND we show the new games. But we don't throw away the baby with the bathwater in trying to open it. 

This has both - AND the inviting, open games list where I can see the full games, AND the context of what the hell I'm looking at. 

I know right now the 'read the rules' is not the BEST - but already much better than it was - but this should become a showcase of the value proposition of Diplicity in the long run. Ideally, after the interactive tutorial is done. 

WDYT?

johnpooch:
Hey @JorenC thanks for the push back and for sharing your thoughts. I think you're right about a lot of this. I want to share where I'm coming from. I think there is a middle ground that is worth exploring. 

Just for context, I took inspiration from Blue Sky for this. There are other websites like Reddit that have a logged-out home page experience. 

> While the desktop implementation is okay-ish, the mobile implementation is not. I just come on a home screen with title 'Open games' (Where am I? What is this? What game? Title of the website?). 

This is fair. 

---

I think what I'm trying to achieve here is something that allows people in the diplomacy community to get interested in Diplicity without necessarily having to sign up. I think that the game detail screen could be the answer to this problem. Maybe we could allow logged-out viewing of specific games so that when people share links to their games, people can see them. 

---

I'm not sure I totally like your mock-ups above. It feels a bit noisy and jarring. 

I think we need to create a GitHub discussion item to track this. I'm going to close this PR for now.

JorenC:
image
Uploading here for future discussion. I see your BSky analogy, but:
- Note that bsky is also having a clear 'homepage' you need to click away before you see the open elements:
image

- Note that bsky is a fundamental different service. You consume the content by just reading it - that would be the same if for Dip you would basically come _mainly_ for the game overviews. You don't - you come for a deep inclusion into one game. 
I think that what we need is:
- Introduction: Where am I?
- Clear CTA: What should I do (login/register)
- More info: What is this? - learn more, learn about the game
- Showcase games: Show that there are open games. Allow users to see them, explore, and join (<- this is your Bsky analogy).
- Value propositions: What is different/better about diplicity than other same sites? - e.g. variants, modes, features, community size (amount of players/games currently ongoing). 

_I think that at the same time, we should make a logical flow from 'seeing a game' - e.g. share your game link - to log in, as I made earlier in a very old PR for this. Basically, without signing in I should be able to see any game, and from there (if open) easily join into the game_. This combination with allowing people to explore (open and/or ongoing games) would allow for users to explore games - to learn what it is about by SEEING it while at the same time preventing dropping them in cold.

johnpooch:
I feel like in the spirit of breaking this down into smaller problems, it feels like the next increment would be enable logged-out viewers to spectate a game. Would you agree?

PR #952, opened 2026-07-02
Write LLMCall rows for bot LLM calls

timwyse:
## What this PR does

Starts populating the `LLMCall` telemetry model (added in #936). Persists one row per bot LLM call at the **plan**, **commit**, and **reply** stages, capturing token usage and status for cost/inspection analysis.

- `LLMClient.complete` now returns an `LLMResult` (response text, resolved model, and token counts — input, output, cache read/write) read from the Anthropic response, instead of a bare string.
- A recording helper in `bot/tasks.py` writes exactly one `LLMCall` row per call on **both** success and error, linking the phase the context was built from (`current_phase_id`) and the bot's member.
- Telemetry writes are wrapped so a recording failure logs and moves on — it never breaks order submission or replies.

`NEGOTIATE` remains defined-but-unused: there is no separate negotiate task yet (`reply` is the in-phase negotiation), matching the design doc.

## Checklist

- [x] Tests cover the change — `complete` usage capture; plan/reply record success rows with the right phase/member/tokens; plan records an error row when the LLM is unavailable
- [x] Screenshots embedded in the PR description for any visual changes (see CLAUDE.md) — N/A, backend-only

https://claude.ai/code/session_01X5syXinzpp3VshVAFPzSqM

---
_Generated by [Claude Code](https://claude.ai/code/session_01X5syXinzpp3VshVAFPzSqM)_

### DIFF
--- a/service/bot/llm_client.py
+++ b/service/bot/llm_client.py
@@ -1,5 +1,5 @@
 import logging
-from dataclasses import dataclass
+import time
 
 from anthropic import Anthropic
 from django.conf import settings
@@ -11,24 +11,17 @@ class LLMError(Exception):
     pass
 
 
-@dataclass
-class LLMResult:
-    text: str
-    model: str
-    input_tokens: int
-    output_tokens: int
-    cache_read_tokens: int
-    cache_write_tokens: int
-
-
 class LLMClient:
-    def __init__(self, api_key):
+    def __init__(self, api_key, recorder):
         if not api_key:
             raise LLMError("ANTHROPIC_API_KEY is not set")
         self._client = Anthropic(api_key=api_key)
+        self._recorder = recorder
 
     def complete(self, *, system, messages):
         logger.info(f"[bot.llm] completing via {settings.BOT_LLM_MODEL}")
+        user_content = "\n\n".join(message["content"] for message in messages)
+        start = time.monotonic()
         try:
             message = self._client.messages.create(
                 model=settings.BOT_LLM_MODEL,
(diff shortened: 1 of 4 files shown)


service/bot/llm_client.py
```diff
@@ -10,6 +11,16 @@ class LLMError(Exception):
...
+    text: str
+    model: str
+    input_tokens: int
+    output_tokens: int
+    cache_read_tokens: int
+    cache_write_tokens: int
+
+
 class LLMClient:
     def __init__(self, api_key):
```
johnpooch: I think that this change is spreading out throughout the tasks. I feel like it should be something that's captured right at the point where the LLM call is made, and it should not leak into the tasks file. 

I think the LLMClient should receive an object on initialization that handles all this behaviour.

Something like:
```
def __init__(self, api_key, recorder):
```

But Recorder could have a different name. I dunno, maybe `result_sink` or something?

service/bot/llm_client.py
johnpooch, line 32: After this line, you would call:
```
self.recorder.record_error(**options)
```

service/bot/llm_client.py
```diff
@@ -31,4 +42,13 @@ def complete(self, *, system, messages):
         text = "".join(block.text for block in message.content if block.type == "text")
         if not text:
             raise LLMError("no text content in response")
-        return text
+
+        usage = message.usage
```
johnpooch, line 53: I'm not sure why we assigned this variable here.

service/bot/llm_client.py
```diff
@@ -31,4 +42,13 @@ def complete(self, *, system, messages):
...
+        return LLMResult(
```
johnpooch: I think we should return text like before, but also call `self.recoder.record_success`

service/bot/tasks.py
```diff
@@ -12,12 +15,95 @@
...
 )
 from bot.llm_client import LLMClient, LLMError
+from bot.models import LLMCall
 from bot.prompts import load_prompt
 
 logger = logging.getLogger(__name__)
 
 
-def _submit_orders_from_context(api, data, game_id, label):
+def _record_llm_call(
```
johnpooch: I think all of the recording logic can be removed from this file once we use the recorder object approach I described above. 

We should be able to statically define a recorder implementation in a recorder.py file.

johnpooch: I think that the only changes to this file should be the fact that the recorder is passed to the LLM client on initialization.

johnpooch (commented):
I'm not sure about my naming suggestions at all, but the fundamental direction of what I am proposing will massively simplify this PR. 

I don't like the way this PR spreads its tentacles through all of the tasks.

service/bot/llm_client.py
timwyse: Done — reworked along these lines. `LLMClient.__init__(api_key, recorder)` now takes a recorder and calls `record_error`/`record_success` around the API call; `complete()` returns text as before. All recording logic is out of `tasks.py` — it just builds the recorder (bound to game/user/phase/stage) and passes it in. The implementation lives in a new `bot/recorder.py`.

On the name: I went with `recorder` / `LLMCallRecorder` since it matched your wording most closely, but it's trivially renameable to `result_sink` if you prefer that — happy to switch.

One thing worth flagging: the recorder still needs the domain context (phase, member, stage), so `tasks.py` constructs it per call. That's the only task-side change beyond passing it to the client.

---

PR #954, opened 2026-07-02
Add tactical annotations and nation identity to bot context (Gap 03)

johnpooch:
## What this PR does

Closes the game-state serialization gap (`service/bot/docs/GAP_03_GAME_STATE_SERIALIZATION.md`): the bot's context was a bare unit/SC list with no map geometry, so a small model had to re-derive distances and threats itself. This exposes province adjacencies over the variant API and uses them to render per-unit **tactical annotations** — adjacent provinces, nearest enemy units, and nearest uncontrolled supply centers (with distances) — plus names the nation the bot is playing.

**Backend**

- `VariantProvinceSerializer` now serializes `type`, `supply_center`, and the lossless `adjacencies` `[{to, pass}]` shape (via `SerializerMethodField` + `inline_serializer`, since `pass` is a Python keyword).
- `ApiClient.get_variant` + `fetch_context` carry the variant into `ContextData`; a missing `variant_id` degrades gracefully to no annotations rather than failing.

**Bot context**

- New `service/bot/context/map.py` builds a typed (army/fleet) province graph from the `pass` metadata — collapsing named coasts into their parent — and runs BFS for nearest enemy units and nearest uncontrolled supply centers. The graph is internal computation only; the model sees plain prose, never graph structure.
- `ContextBuilder` gains `with_tactical_annotations()` (shared block, all units, deterministic sort by province id) and `with_identity()` (`You are playing <nation>.`, private block). Phase-type framing is added to the shared game-state block: a dislodged-units line for retreats and per-nation build/disband deltas for adjustments.
- Per the Gap 10 correction, legal orders and the per-bot order count move to the **private** block — both are per-bot data (`order-options` only returns the acting user's options), not shared. Every shared section sorts deterministically so the block is byte-identical across bots, which the Gap 10 caching work depends on.

**Codegen**

- Regenerated `service/openapi-schema.yaml` and the TypeScript client for the new `VariantProvince` fields; updated the inline mock provinces (`src/mocks/legacy.ts`, `buildOptimisticOrder.test.ts`) for the now-required `adjacencies` field.

## Checklist

- [x] Tests cover the change
- [x] Screenshots embedded in the PR description for any visual changes (see CLAUDE.md) — N/A, no visual changes

### Tests

- `variant/tests/test_variant_crud.py`: variant endpoint returns `type`, `supply_center`, and `adjacencies` (incl. `stp` ↔ `mos` and the `stp/nc` named-coast row).
- `bot/tests.py`: new `TestMapGraph` (typed edges, coast collapsing, BFS distances per unit type, tie-breaks, n-capping, empty-adjacency degradation) and expanded `TestContextBuilder` (annotation blocks, identity line, orders/phase-state now private, adjustment/retreat framing, shared-block byte-identity).
- `integration/test_bot.py`: existing bot-plays-phase tests still pass; new `test_bot_plans_when_variant_has_no_adjacencies` asserts graceful degradation.
- Full affected backend suites (bot, variant, province, phase, integration) green — 478 passed. Frontend `tsc -b --noEmit`, `eslint`, and the affected vitest file all clean.

https://claude.ai/code/session_01KuTX6diksRZVkPgFoWQ7aD

---
_Generated by [Claude Code](https://claude.ai/code/session_01KuTX6diksRZVkPgFoWQ7aD)_

### DIFF
--- a/service/bot/context/fetch.py
+++ b/service/bot/context/fetch.py
@@ -1,6 +1,6 @@
 import logging
 
-from bot.api_client import ApiClient
+from bot.api_client import ApiClient, ApiClientError
 from bot.types import ContextData
 
 logger = logging.getLogger(__name__)
@@ -11,7 +11,12 @@ def fetch_context(api: ApiClient, game_id) -> ContextData:
     current_phase_id = game.get("current_phase_id")
     phase = api.get_phase(game_id, current_phase_id) if current_phase_id else {}
     variant_id = game.get("variant_id")
-    variant = api.get_variant(variant_id) if variant_id else {}
+    variant = {}
+    if variant_id:
+        try:
+            variant = api.get_variant(variant_id)
+        except ApiClientError as e:
+            logger.warning(f"[bot.context] variant fetch failed ({e}); continuing without tactical annotations")
     data: ContextData = {
         "orders": api.get_order_options(game_id),
         "phase_states": api.get_phase_states(game_id),
--- a/service/bot/context/fetch.py
+++ b/service/bot/context/fetch.py
@@ -10,16 +10,19 @@ def fetch_context(api: ApiClient, game_id) -> ContextData:
     game = api.get_game(game_id)
     current_phase_id = game.get("current_phase_id")
     phase = api.get_phase(game_id, current_phase_id) if current_phase_id else {}
+    variant_id = game.get("variant_id")
+    variant = api.get_variant(variant_id) if variant_id else {}
     data: ContextData = {
         "orders": api.get_order_options(game_id),
         "phase_states": api.get_phase_states(game_id),
         "game": game,
         "phase": phase,
         "channels": api.get_channels(game_id),
+        "variant": variant,
     }
     logger.info(
         f"[bot.context] fetched context for game {game_id}: "
         f"{len(data['orders'])} order option(s), {len(data['channels'])} channel(s), "
-        f"{len(phase.get('units', []))} unit(s)"
+        f"{len(phase.get('units', []))} unit(s), {len(variant.get('provinces', []))} province(s)"
     )
     return data
--- a/service/bot/context/fetch.py
+++ b/service/bot/context/fetch.py
@@ -1,6 +1,6 @@
 import logging
 
-from bot.api_client import ApiClient, ApiClientError
+from bot.api_client import ApiClient
 from bot.types import ContextData
 
 logger = logging.getLogger(__name__)
@@ -11,12 +11,7 @@ def fetch_context(api: ApiClient, game_id) -> ContextData:
     current_phase_id = game.get("current_phase_id")
     phase = api.get_phase(game_id, current_phase_id) if current_phase_id else {}
     variant_id = game.get("variant_id")
-    variant = {}
-    if variant_id:
-        try:
-            variant = api.get_variant(variant_id)
-        except ApiClientError as e:
-            logger.warning(f"[bot.context] variant fetch failed ({e}); continuing without tactical annotations")
+    variant = api.get_variant(variant_id) if variant_id else {}
     data: ContextData = {
         "orders": api.get_order_options(game_id),
         "phase_states": api.get_phase_states(game_id),
--- a/service/variant/serializers.py
+++ b/service/variant/serializers.py
@@ -3,6 +3,7 @@ from typing import Optional
 
 from django.db import IntegrityError, transaction
 from django.urls import reverse
+from drf_spectacular.utils import extend_schema_field, inline_serializer
 from lxml import etree
 from rest_framework import serializers
 from nation.serializers import NationSerializer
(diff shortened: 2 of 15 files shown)


service/bot/context/fetch.py
```diff
@@ -10,16 +10,24 @@ def fetch_context(api: ApiClient, game_id) -> ContextData:
     game = api.get_game(game_id)
     current_phase_id = game.get("current_phase_id")
     phase = api.get_phase(game_id, current_phase_id) if current_phase_id else {}
+    variant_id = game.get("variant_id")
+    variant = {}
+    if variant_id:
+        try:
+            variant = api.get_variant(variant_id)
+        except ApiClientError as e:
```
johnpooch: Why would this fail? Why are we treating this as an optional endpoint, whereas the other ones are required?

service/variant/serializers.py
```diff
@@ -28,7 +29,20 @@ class VictoryConditionsSerializer(serializers.Serializer):
 class VariantProvinceSerializer(serializers.Serializer):
     id = serializers.CharField(source="province_id")
     name = serializers.CharField()
+    type = serializers.CharField()
+    supply_center = serializers.BooleanField()
     parent_id = serializers.CharField(source="parent.province_id", allow_null=True)
+    adjacencies = serializers.SerializerMethodField()
```
johnpooch, line 35: Does this not significantly increase the size of the variant's payload? Could this have an impact on performance in production?

service/bot/context/fetch.py
johnpooch: Good call — reverted in 9324f76 to treat the variant as required, like the other fetches.

My reasoning for making it optional was weak: `get_variant` hits the same in-process, authenticated `reverse()` path as `get_game`/`get_phase`, and the variant always exists (the game references it by FK), so it can't realistically fail in isolation while the essential fetches succeed. If the API were that broken, aborting the turn is the correct behaviour — you don't want a bot submitting orders against a half-broken backend. It's now `variant = api.get_variant(variant_id) if variant_id else {}`, mirroring the `phase` guard on the line directly above.

The graceful degradation the gap doc actually asked for is for missing adjacency *data* (a variant whose provinces have empty `adjacencies`) — that path renders no annotations and is unaffected, still covered by `test_bot_plans_when_variant_has_no_adjacencies`.

---

service/variant/serializers.py
johnpooch: It does grow the payload, but the impact is small. I measured against the seeded DB (6 variants, 548 provinces):

- **List endpoint** (what the frontend loads for GameMap/GameInfo/CreateGame): ~81 KB → ~182 KB raw, but adjacency data is extremely repetitive (`{"to":"...","pass":"army"}`), so gzipped it's only **~8.7 KB → ~16.6 KB** (+~8 KB on the wire).
- **Detail endpoint** (what the bot fetches): ~10 KB → ~29 KB.
- **Zero extra DB queries** — `adjacencies` is a `JSONField` already loaded with the province row (`with_related_data()`), so no N+1; serialize count stays at 18 for the whole list.
- The list endpoint is **ETag-cached** (`_variants_list_etag`, keyed on `updated_at`) and variants are static/seeded, so the body is only transferred when a variant actually changes — most requests get a 304.

So the cost is a few KB gzipped on a static, cached, no-N+1 endpoint. I did consider scoping `adjacencies` to the detail endpoint only (no frontend code reads `.adjacencies`; only the bot needs it, via detail) — happy to do that if you'd prefer — but given the measured impact we agreed to keep it on both for simplicity.

---

johnpooch (commented):
### PR Review

The PR exposes province adjacencies over the variant API and renders per-unit tactical annotations plus nation identity into the bot context. I verified it end-to-end and found no issues worth flagging.

Correctness checks that passed:
- **snake_case reads are correct** — `map.py`/`builder.py` read `supply_center`, `parent_id`, `to`/`pass`, `member.nation`, `variant_id` off `response.data`, which is pre-render, so the `CamelCaseJSONRenderer` doesn't apply. Consistent with the existing `max_orders`/`current_phase_id` reads.
- **Serializer fields exist** — `GameRetrieveSerializer.variant_id` and `MemberSerializer.nation` (source `nation.name`) are present, so `with_identity()` resolves the real nation rather than silently no-op'ing.
- **`phase_states[0]` is the bot's own** — `PhaseStateListView.get_queryset` filters `member__user=request.user`, so identity and `max_orders` are per-bot.
- **Phase framing strings match** — `PhaseType.RETREAT`/`ADJUSTMENT` equal `"Retreat"`/`"Adjustment"`, the exact values the serializer emits.
- **No prompt regression from the shared→private move** — both order-selection and reply flows join `build_shared()` + `build_private()`, so relocating legal orders / max_orders / identity to the private block keeps them in the LLM prompt.
- **Graceful degradation** — missing `variant_id` and empty adjacencies both short-circuit to no annotations; covered by `test_bot_plans_when_variant_has_no_adjacencies` and `test_with_tactical_annotations_without_variant_is_noop`.

Evidence: `bot/tests.py` 62 passed locally; frontend `tsc -b --noEmit` and `eslint` on the changed files clean. The named-coast collapsing and the adjustment `SC − units` delta are documented, intentional heuristics (with `max_orders` as the authoritative private figure), not correctness defects.

---

PR #958, opened 2026-07-02
Attribute bot chat transcripts, inject persona, harden prompts

johnpooch:
## What this PR does

This is **PR A** of the Gap 09 negotiation-loop plan — a self-contained step that improves today's public-only bot chat behaviour without any trigger changes, laying the groundwork for private and bot-to-bot negotiation in later PRs.

Currently a bot sees every non-self message as an anonymous `user:` voice, so in a multi-member channel it cannot tell who said what, and the seeded `BotProfile.disposition`/`voice` personas are injected into no prompt. This PR:

- **Attributes transcripts by nation.** `_message_role` labels non-self messages with the sender's nation name (`nation.name`, already in the payload via `ChannelMemberSerializer.nation`), falling back to `user` when absent; own messages stay `assistant`. The channel header is marked `(public)`/`(private)` so the prompt applies public-press rules to the right transcript.
- **Adds `ChatSenderDict.nation`** to the context types.
- **Injects the persona block.** New `_persona_block(user_id)` helper reads the bot's `BotProfile` and appends `Disposition`/`Voice` to the **system** prompt for order selection (`plan`, `finalize`) and chat replies, so both orders and press stay in character.
- **Hardens prompts.** `reply_system.txt` is rewritten for multi-party chat (nation-labelled senders, judiciousness criteria — only speak when directly addressed, materially affected, or making a concrete proposal — and an injection guardrail). The same guardrail sentence is added to `select_orders_system.txt`.

## Checklist

- [x] Tests cover the change
- [x] Screenshots embedded in the PR description for any visual changes (see CLAUDE.md) — n/a, no visual changes

Builder tests updated for nation labels and privacy markers; new tests assert the persona is injected into the reply and plan system prompts. Full `bot/tests.py` suite passes (76 passed). The two failures in `integration/test_bot.py` are pre-existing in this environment (a 403 on the options endpoint) and reproduce on a clean tree.

https://claude.ai/code/session_018VgeK1D5PGSyLyi2g3mEvs

---
_Generated by [Claude Code](https://claude.ai/code/session_018VgeK1D5PGSyLyi2g3mEvs)_


service/bot/context/builder.py
```diff
@@ -7,7 +7,10 @@
 
 
 def _message_role(message: ChatMessageDict) -> str:
-    return "assistant" if message["sender"]["is_current_user"] else "user"
+    if message["sender"]["is_current_user"]:
+        return "assistant"
+    nation = (message["sender"].get("nation") or {}).get("name")
+    return nation or "user"
```
johnpooch: I feel like this is a weird misuse of the message role field when constructing an LLM message. We're conflating assistant and user in a weird way. Should the user role only be added to the newly created message? Then again, we don't really track what the newly created message is. I think we need a better strategy for this. Maybe all messages that aren't created by the current user are marked as user, and the nation is applied to the message separately in the line.

johnpooch: Good catch — fixed in 11c5479. Worth noting the whole transcript is rendered into a single `user`-content string (`_format_channel` → `user_content`), so this value was never an actual API `role`, just a text prefix on each line. But the name and the mixed return value were genuinely confusing.

Rather than "user role + nation separately", I went (with your confirmation) fully symmetric: every line is now labelled by the sender's nation, including the bot's own past messages, and the board-state section names which nation the bot is playing (`You are playing France.`, derived from the game members). So the assistant/user concept is dropped from the transcript entirely and there's no more conflation:

```
You are playing France.
...
France: I propose we DMZ the Channel.
England: Agreed, but hold Burgundy.
Germany: I'd rather move to Ruhr.
```

`_message_role` is renamed to `_sender_label` and no longer branches on `is_current_user`.

---

service/bot/tasks.py
```diff
@@ -13,13 +13,21 @@
...
 
 logger = logging.getLogger(__name__)
 
 
-def _submit_orders_from_context(api, data, game_id, user_id, label, stage):
+def _persona_block(user_id) -> str:
+    profile = BotProfile.objects.filter(user_id=user_id).first()
+    if profile is None:
+        return ""
+    return f"Your persona:\nDisposition: {profile.disposition}\nVoice: {profile.voice}"
```
johnpooch: I feel like the persona and disposition section deserves an entire prompt file which would be injected into the system prompt. It should be a paragraph which prefixes the disposition and persona, explaining what they mean and when they should be used. 

Something to this effect
```
disposition — governs how the bot plays strategically. It decides moves. It does not influence how the bot communicates.
voice — governs how the bot communicates (tone, register, negotiation style). It affects wording only, never strategic decisions.
```

johnpooch: Done in a7257c6. Added `bot/prompts/persona_system.txt`:

```
You play with a persona made up of two parts, and you must keep them separate:

- Disposition governs how you play strategically. It decides your moves and what you pursue in negotiation. It does not influence how you communicate.
- Voice governs how you communicate: your tone, register, and negotiation style. It affects your wording only, never your strategic decisions.
```

`_persona_block` now loads that preamble and appends the bot's concrete `Disposition:`/`Voice:` values, and the whole block is injected into the system prompt for both order selection (plan/finalize) and replies.

---

PR #966, opened 2026-07-03
Add Android release CI/CD workflow

### LINKED ISSUE #925: Android release CI/CD workflow
## Goal

Add a GitHub Actions workflow that builds a signed Android AAB and uploads it to the Play Store beta track on every push to `main` that touches `packages/web/**`.

## Context

iOS already has an equivalent `ios-release.yml` workflow. The Fastfile already has an `android release` lane that builds a signed AAB and calls `supply`. The workflow should set `PLAY_TRACK=beta` so builds land in the Play Store beta track rather than internal testing, allowing controlled promotion to production.

## Approach

- Use `ubuntu-latest` runner (not macOS — Android doesn't need it).
- Add `actions/setup-java` step with `java-version: 21` (required by Capacitor Android 8).
- Add `android-actions/setup-android` step and explicitly accept SDK licenses to ensure SDK 36 is available — otherwise Gradle fails silently waiting for license acceptance.
- Mirror the rest of the structure from `ios-release.yml`: checkout, set up Node + Ruby, install deps, build web app (`VITE_GOOGLE_CLIENT_ID`, `VITE_DIPLICITY_API_BASE_URL`, `VITE_MAINTENANCE_MODE`), sync Capacitor, run `bundle exec fastlane android release`.
- Set `PLAY_TRACK=beta` in the workflow env.
- Pass secrets: `ANDROID_KEYSTORE_BASE64`, `ANDROID_KEYSTORE_PASSWORD`, `ANDROID_KEY_ALIAS`, `ANDROID_KEY_PASSWORD`, `PLAY_SERVICE_ACCOUNT_JSON`.
- Write the `GOOGLE_SERVICES_JSON` secret to `packages/web/android/app/google-services.json` before the Fastlane step. The file is gitignored, and `android/app/build.gradle` wraps the google-services plugin in a try/catch — so without this step the build **succeeds silently without Firebase** and the beta release ships with push notifications broken.
- Remove `bump_patch_version` from the `android release` Fastlane lane. Version bumps should be a deliberate developer action, not a CI side effect. The `versionCode` is already a Unix timestamp so Play Console accepts every upload regardless of `versionName`.

johnpooch:
## What this PR does

Adds `android-release.yml`, mirroring the existing `ios-release.yml`: on every push to `main` that touches `packages/web/**`, it builds a signed Android AAB and uploads it to the Play Store **beta** track via the existing `bundle exec fastlane android release` lane. It runs on `ubuntu-latest` (no macOS needed for Android), sets up JDK 21 and the Android SDK, and explicitly accepts SDK licenses — otherwise Gradle silently stalls waiting for license acceptance.

It also writes the `GOOGLE_SERVICES_JSON` secret to `packages/web/android/app/google-services.json` before the Fastlane step. That file is gitignored and `build.gradle` applies the Firebase plugin conditionally, so without this step the build would succeed silently without Firebase and the beta release would ship with push notifications broken (flagged in issue comments).

Per feedback on the issue, this also removes `bump_patch_version` from the `android release` Fastlane lane (and the now-unused private lane itself) — version bumps should be a deliberate developer action, not a CI side effect, and `versionCode` (already a Unix timestamp) lets Play Console accept every upload regardless of `versionName`.

**Note for reviewer:** the first CI run will only succeed once the required secrets (`ANDROID_KEYSTORE_BASE64`, `ANDROID_KEYSTORE_PASSWORD`, `ANDROID_KEY_ALIAS`, `ANDROID_KEY_PASSWORD`, `PLAY_SERVICE_ACCOUNT_JSON`, `GOOGLE_SERVICES_JSON`) are set in the repo, at least one manual AAB has been uploaded to the Play Console beta track, and the Play service account has the Release Manager role — all called out in the issue thread.

Closes #925

---
_Generated by [Claude Code](https://claude.ai/code/session_0157u8SxvGdcNbxGPJ9Xp2Ya)_

### DIFF
--- a/.github/workflows/android-release.yml
+++ b/.github/workflows/android-release.yml
@@ -1,0 +1,73 @@
+name: Android Release
+
+on:
+  push:
+    branches:
+      - main
+    paths:
+      - "packages/web/**"
+
+concurrency:
+  group: android-release
+  cancel-in-progress: false
+
+env:
+  PLAY_TRACK: beta
+
+jobs:
+  release:
+    runs-on: ubuntu-latest
+    timeout-minutes: 30
+    defaults:
+      run:
+        working-directory: packages/web
+
+    steps:
+      - uses: actions/checkout@v7
+
+      - uses: actions/setup-java@v4
+        with:
+          distribution: "temurin"
+          java-version: "21"
+
+      - uses: android-actions/setup-android@v3
+
+      - name: Accept Android SDK licenses
+        run: yes | sdkmanager --licenses || true
+
+      - uses: actions/setup-node@v6
+        with:
+          node-version-file: ".nvmrc"
+          cache: "npm"
+          cache-dependency-path: packages/web/package-lock.json
+
+      - uses: ruby/setup-ruby@v1
+        with:
+          ruby-version: "3.2"
+          bundler-cache: true
+          working-directory: packages/web
+
+      - name: Install npm dependencies
+        run: npm ci
+
+      - name: Build web app
+        run: npm run build
+        env:
+          VITE_GOOGLE_CLIENT_ID: ${{ secrets.GOOGLE_OAUTH_CLIENT_ID }}
+          VITE_DIPLICITY_API_BASE_URL: ${{ vars.SERVICE_BASE_URL }}
+          VITE_MAINTENANCE_MODE: ${{ vars.MAINTENANCE_MODE }}
+
+      - name: Sync Capacitor
+        run: npx cap sync android
+
+      - name: Write google-services.json
+        run: echo '${{ secrets.GOOGLE_SERVICES_JSON }}' > android/app/google-services.json
+
+      - name: Build and upload to Play Console beta track
+        run: bundle exec fastlane android release
+        env:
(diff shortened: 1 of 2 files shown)


johnpooch (commented):
### PR Review

Adds `android-release.yml` mirroring `ios-release.yml`, and removes `bump_patch_version` (plus its now-unused private lane) from the Fastlane `android release` lane, per the linked issue's approach.

Verified: the `google-services.json` write path (`packages/web/android/app/google-services.json`) matches the conditional `apply plugin` check in `build.gradle` and the repo's `.gitignore` entries, and is written before the Fastlane step that triggers the Gradle build. `PLAY_TRACK=beta` is set as a workflow-level env var, which Fastfile's `upload_to_play` correctly reads via `ENV.fetch("PLAY_TRACK", "internal")`. `bump_patch_version` has no other callers (`pr_build` never called it), so removing it doesn't break anything. No blocking issues found.

---

JorenC:
@johnpooch I think we can pull this. Fable double checked - please read the comment about the GOOGLE_SERVICES_JSON thing

PR #974, opened 2026-07-03
Poll channel messages and phase states while a game is active

### LINKED ISSUE #973: Channel messages and phase states don't refresh live on game detail screens
## Goal

Make channel messages and per-player phase state refresh on the game detail screens without the user having to manually refresh or navigate away and back.

## Current behaviour

The channel thread reads messages out of `useGamesChannelsListSuspense`, which has no `refetchInterval`, so incoming press messages only appear when something else invalidates the channels list (sending a message, marking read, or navigating). The FCM foreground handler that would invalidate queries (`useMessaging.ts`) is only mounted on the Account screen and in `useLogout`, not on any game detail screen, so a received push does not refresh channel data while the user is viewing a game. Phase states (`useGamePhaseStatesListSuspense`) are likewise fetched once with no polling, so other players' and bots' order confirmations do not appear until the screen is refetched for some other reason. The only active polling is `useGameRetrieve` in `GameDetailLayout.tsx`, refetching every 10s while the game is active.

## Expected behaviour

While a game is active and the user is on a game detail screen, new channel messages and updated phase states appear on their own within a few seconds, without a manual refresh or navigation.

## Where

- `packages/web/src/screens/GameDetail/ChannelScreen.tsx` — channel thread via `useGamesChannelsListSuspense`
- `packages/web/src/screens/GameDetail/OrdersScreen.tsx` — phase states via `useGamePhaseStatesListSuspense`
- `packages/web/src/components/GameDetailLayout.tsx` — existing `useGameRetrieve` poll (10s)

## Approach

Add a `refetchInterval` to the channels-list and phase-states queries on the game detail screens, matching the existing pattern in `GameDetailLayout.tsx` of polling only while the game status is `active`. Reduce the existing game-retrieve polling interval from 10s to 5s.

johnpooch:
## What this PR does

Channel messages and phase states didn't refresh live on the game detail screens — `useGamesChannelsListSuspense` and `useGamePhaseStatesListSuspense` were fetched once on mount with no `refetchInterval`, so incoming press messages and other players'/bots' order confirmations only appeared after something else (sending a message, marking read, navigating) invalidated the query.

This adds a `refetchInterval` to both queries, gated on `game.status === "active"`, matching the existing polling pattern already used for `useGameRetrieve` in `GameDetailLayout.tsx`. The existing `useGameRetrieve` poll is also reduced from 10s to 5s so all three queries refresh at the same cadence.

Closes #973

## Checklist

- [x] Tests cover the change — existing `ChannelScreen.test.tsx` and `OrdersScreen.test.tsx` suites pass unchanged (11 tests)
- [x] Screenshots — not applicable, no visual change (polling behaviour only)

---
_Generated by [Claude Code](https://claude.ai/code/session_01Ncw9L3e1h8tpDvqWXnVBTc)_

### DIFF
--- a/packages/web/src/screens/GameDetail/ChannelScreen.tsx
+++ b/packages/web/src/screens/GameDetail/ChannelScreen.tsx
@@ -108,7 +108,7 @@ const ChannelScreen: React.FC = () => {
   const { data: game } = useGameRetrieveSuspense(gameId);
   const { data: channels } = useGamesChannelsListSuspense(gameId, {
     query: {
-      refetchInterval: game.status === "active" ? 5000 : false,
+      refetchInterval: 5000,
     },
   });
   const { data: variants } = useVariantsListSuspense();
--- a/packages/web/src/screens/GameDetail/ChannelScreen.tsx
+++ b/packages/web/src/screens/GameDetail/ChannelScreen.tsx
@@ -106,7 +106,11 @@ const ChannelScreen: React.FC = () => {
   const [, setSearchParams] = useSearchParams();
 
   const { data: game } = useGameRetrieveSuspense(gameId);
-  const { data: channels } = useGamesChannelsListSuspense(gameId);
+  const { data: channels } = useGamesChannelsListSuspense(gameId, {
+    query: {
+      refetchInterval: game.status === "active" ? 5000 : false,
+    },
+  });
   const { data: variants } = useVariantsListSuspense();
   const createMessageMutation = useGamesChannelsMessagesCreateCreate();
   const markReadMutation = useGamesChannelsMarkReadCreate();
--- a/packages/web/src/screens/GameDetail/OrdersScreen.tsx
+++ b/packages/web/src/screens/GameDetail/OrdersScreen.tsx
@@ -168,11 +168,7 @@ const OrdersScreen: React.FC = () => {
   const { data: phase } = useGamePhaseRetrieveSuspense(gameId, selectedPhase);
   const { data: orders } = useGameOrdersListSuspense(gameId, selectedPhase);
   const { data: variants } = useVariantsListSuspense();
-  const { data: phaseStates } = useGamePhaseStatesListSuspense(gameId, {
-    query: {
-      refetchInterval: game.status === "active" ? 5000 : false,
-    },
-  });
+  const { data: phaseStates } = useGamePhaseStatesListSuspense(gameId);
 
   const deleteOrderMutation = useGameOrdersDeleteDestroy();
   const confirmOrdersMutation = useGameConfirmPhasePartialUpdate();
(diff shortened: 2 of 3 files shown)


johnpooch (commented):
### PR Review

Adds `refetchInterval` (gated on `game.status === "active"`) to the channels-list and phase-states queries in `ChannelScreen.tsx` and `OrdersScreen.tsx`, and reduces `GameDetailLayout`'s existing game poll from 10s to 5s — matching the pattern the issue asked for.

Verified: `useGameRetrieve` (parent) and `useGameRetrieveSuspense` (children) share the same query key (`/game/${gameId}/`), so `GameDetailLayout`'s poll keeps `game.status` fresh for the children's `refetchInterval` gate even though the children don't poll the game themselves. Also verified `/game/${gameId}/phase-states/` is a game-scoped endpoint (not phase-scoped), so gating its poll on `game.status` rather than the selected phase's status is correct — the same data feeds every phase view via one shared query.

**0** inline comment(s) (0 blocking, 0 suggestion). No issues found — the change is small, correct, and consistent with the existing polling pattern.

---

packages/web/src/screens/GameDetail/ChannelScreen.tsx
```diff
@@ -106,7 +106,11 @@ const ChannelScreen: React.FC = () => {
   const [, setSearchParams] = useSearchParams();
 
   const { data: game } = useGameRetrieveSuspense(gameId);
-  const { data: channels } = useGamesChannelsListSuspense(gameId);
+  const { data: channels } = useGamesChannelsListSuspense(gameId, {
+    query: {
+      refetchInterval: game.status === "active" ? 5000 : false,
```
johnpooch: In the future, we will support chat channels before the game is active and after the game is finished. I think we should remove this condition here.

packages/web/src/screens/GameDetail/OrdersScreen.tsx
```diff
@@ -168,7 +168,11 @@ const OrdersScreen: React.FC = () => {
   const { data: phase } = useGamePhaseRetrieveSuspense(gameId, selectedPhase);
   const { data: orders } = useGameOrdersListSuspense(gameId, selectedPhase);
   const { data: variants } = useVariantsListSuspense();
-  const { data: phaseStates } = useGamePhaseStatesListSuspense(gameId);
+  const { data: phaseStates } = useGamePhaseStatesListSuspense(gameId, {
+    query: {
+      refetchInterval: game.status === "active" ? 5000 : false,
+    },
+  });
```
johnpooch: Do we really need this? When does the phase state change? Is it possible that phase state can change through a non-user action? I wonder if this is unnecessary.

packages/web/src/screens/GameDetail/ChannelScreen.tsx
johnpooch: Good point, done — removed the `game.status === "active"` gate, `refetchInterval` is now unconditional (5s) whenever the screen is mounted.

---

packages/web/src/screens/GameDetail/OrdersScreen.tsx
johnpooch: You're right, this was unnecessary — dropped it. Checked `PhaseStateListView.get_queryset()` (`service/phase/views.py:50-53`): it filters to `current_phase.phase_states.filter(member__user=self.request.user)`, so this endpoint only ever returns the requesting user's own phase state (confirmed by `test_list_orderable_provinces_success` asserting `len(response.data) == 1` for a normal member). It never returns other players' or bots' state, so polling it can't surface bot/other-player confirmations regardless — the issue's premise for this particular query was wrong. Removed the added `refetchInterval`.

---

PR #990, opened 2026-07-03
Redesign My Games active tab with triage lanes

johnpooch:
## What this PR does

Redesigns the **active ("Started") tab** of My Games so a player can triage what needs their attention at a glance, replacing the dense row of status chips on each card.

**Triage lanes.** Active games are grouped into priority-ordered lanes, reusing the section-header styling from the Find Games "Fastest Start" section. Each game lands in exactly the highest lane it qualifies for:

1. **Civil Disorder — submit to rejoin** (`memberStatus` includes `civil_disorder`)
2. **No moves received last phase** (`memberStatus` includes `nmr`)
3. **Needs your orders** (`orderStatus === "orders_required"`)
4. **Waiting on others** (orders submitted/confirmed / nothing to do)
5. **Eliminated** (current user eliminated — kept as before, now a proper lane)

As backend state changes, games move between lanes naturally (e.g. rejoin from CD → NMR lane → submit orders → Waiting on others).

**GameCard `active` mode.** A new `mode` prop (default `"default"`, so Find Games / Staging / Finished are untouched) switches the card presentation:

- The nation moves out of the chip row to a **floating chip over the top-left of the map** — white background, border in the nation colour, and the nation's flag as an icon.
- The order-status chip becomes a **status line**: "Orders due in X" (required), "Orders not confirmed. Phase resolves in X", "Orders confirmed. Phase resolves in X", or "No orders required. Phase resolves in X".
- **Game-mode indicators** (gunboat, sandbox) move inline beside the private-lock icon, and the redundant phase-duration text ("24 hours" / "Resolve when ready") is dropped.
- A **paused** game dims the map with a centred "Paused" overlay and shows "Paused — resumes when unpaused" instead of a countdown.

Supporting change: `formatRemainingTime` gains an optional `withSuffix` argument so the status line can render a bare "3h 40m" without the " remaining" suffix. A mock-fixture fix sets the civil-disorder game's game-level `memberStatus` to match the real API shape.

## Checklist

- [x] Tests cover the change — lane grouping + priority in `MyGames.test.tsx`; active-mode status lines, floating nation chip, and paused treatment in `GameCard.test.tsx` (full suite: 510 passing)

## Screenshots

Captured via `npm run dev:mocks` against the seeded fixtures (Civil Disorder, Needs your orders, Waiting on others, Eliminated lanes all visible), desktop 900×1500 and mobile 400×1100. Images shared in the Claude Code session for the author to attach here.

- **Desktop** — Civil Disorder lane (red header + person-x icon), Needs-your-orders lane with amber "Orders due in X" lines and floating England nation chips, Waiting-on-others lane, Eliminated lane.
- **Mobile** — card stacks vertically with the floating nation chip anchored top-left of the map.

---
_Generated by [Claude Code](https://claude.ai/code/session_01QHLsffHXc7Q5L9iAJYRWa2)_

### DIFF
--- a/packages/web/src/components/GameCard.test.tsx
+++ b/packages/web/src/components/GameCard.test.tsx
@@ -143,6 +143,90 @@ describe("GameCard", () => {
     });
   });
 
+  describe("active mode", () => {
+    const activePhase = {
+      id: 5,
+      ordinal: 3,
+      season: "Fall",
+      year: 1901,
+      name: "Fall 1901 Movement",
+      type: "Movement",
+      status: "active",
+      scheduledResolution: "2099-01-18T12:00:00Z",
+      remainingTime: 13200,
+      units: [],
+      supplyCenters: [],
+    };
+    const activeGame = {
+      ...mockActiveGames[0],
+      currentPhase: activePhase,
+      isPaused: false,
+    };
+
+    it("shows 'Orders due in' when orders are required", () => {
+      renderGameCard({
+        game: { ...activeGame, orderStatus: "orders_required" },
+        mode: "active",
+        ...defaultProps,
+      });
+      expect(screen.getByText(/Orders due in 3h 40m/)).toBeInTheDocument();
+    });
+
+    it("shows 'Orders not confirmed' when orders are not confirmed", () => {
+      renderGameCard({
+        game: { ...activeGame, orderStatus: "orders_not_confirmed" },
+        mode: "active",
+        ...defaultProps,
+      });
+      expect(
+        screen.getByText(/Orders not confirmed\. Phase resolves in 3h 40m/)
+      ).toBeInTheDocument();
+    });
+
+    it("shows 'Orders confirmed' when orders are submitted", () => {
+      renderGameCard({
+        game: { ...activeGame, orderStatus: "orders_submitted" },
+        mode: "active",
+        ...defaultProps,
+      });
+      expect(
+        screen.getByText(/Orders confirmed\. Phase resolves in 3h 40m/)
+      ).toBeInTheDocument();
+    });
+
+    it("shows 'No orders required' when no orders are required", () => {
+      renderGameCard({
+        game: { ...activeGame, orderStatus: "no_orders_required" },
+        mode: "active",
+        ...defaultProps,
+      });
+      expect(
+        screen.getByText(/No orders required\. Phase resolves in 3h 40m/)
+      ).toBeInTheDocument();
+    });
+
+    it("renders the nation as a floating chip on the map", () => {
+      renderGameCard({
+        game: { ...activeGame, orderStatus: "orders_required" },
+        mode: "active",
+        ...defaultProps,
+      });
+      expect(screen.getByText("Turkey")).toBeInTheDocument();
+    });
+
+    it("shows a paused treatment instead of a countdown when paused", () => {
+      renderGameCard({
+        game: { ...activeGame, orderStatus: "orders_required", isPaused: true },
+        mode: "active",
+        ...defaultProps,
+      });
+      expect(
+        screen.getByText(/Paused — resumes when unpaused/)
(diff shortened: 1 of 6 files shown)


JorenC:
I think this has some good points in sorting, but it's getting a bit unclear to the user what the status is - especially on mobile games, because you're removed the status from the card. If I miss the header section (which is very easy to miss), I don't get ANY indicator of status. I do like the flag next to the title, and moving them. 

1. We should not colour code the 'No moves received' or 'civil disorder' section titles. Section titles should be kept consistent. 

2. These section titles are very easily missed (especially on mobile) - they are a small line of text in a list of games where each card has a lot of text..
image
I would suggest adding a line divider as well, and give the section headers quite soome more vertical padding (more above and a bit less more below). For desktop, I think the padding is fine.

3. I think we should keep the "Civil Disorder' and 'No Moves Received' status badges next (or underneath) the game name. This helps clarifying - even if the user misses the section header, they can still gleam the information from the cards.

4. I don't like the "submit to rejoin" in the Civil disorder header. For a new player, it's both unclear what CD is, as well as 'what does submit to rejoin mean"? Even I don't know - does it mean I need to submit orders? Confirm them? 

I think this would be better be "Civil disorder" with a question mark icon. Tapping (mobile) or hovering (desktop/mouse device) would give a tooltip that both explains what CD is (how did I get there), what the effect of it is (game doesn't wait on me) as well as how to get out of it (submit any order). That way, there is progressive disclosure for those that need it. 
Something similar can be done for No Moves Received, although I think that's less bad.


_All the above can be picked up by Claude. Questions for @johnpooch - what was the reason to adjust this?_

johnpooch:
@JorenC I agree with your feedback. This PR was a bit of a half-baked effort. I'm going to close it, but extract the main decisions and open questions to a discussion item for future. Maybe respond to the discussion item if you have any comments.

> We should not colour code the 'No moves received' or 'civil disorder' section titles. Section titles should be kept consistent.

Agreed, section headers should not be colour coded.

> These section titles are very easily missed (especially on mobile) - they are a small line of text in a list of games where each card has a lot of text..

>  I would suggest adding a line divider as well, and give the section headers quite soome more vertical padding (more above and a bit less more below). For desktop, I think the padding is fine.

I think a divider line would help agreed. Maybe we should do the same for consistency on the Find Games screen.

> I think we should keep the "Civil Disorder' and 'No Moves Received' status badges next (or underneath) the game name. This helps clarifying - even if the user misses the section header, they can still gleam the information from the cards.





I think I agree. I think the groups/lanes is helpful for surfacing the most urgent games at the top. The issue that I want to solve is the fact that we are stuffing a lot of badges into the same row that represent different things. Right now, a user can have a game card with the following badges all on the same row:
```
[Italy] [Orders Submitted] [NMR] [CD] [Gunboat]
```
That's nation, orders status, nmr/cd status, and game mode all expressed using the same element in the same row. It doesn't really express urgency and doesn't really have much information hierarchy.

I'm trying to move stuff into separate places to express importance, hierarchy, and also separation just to show the user that these are different things.

I feel like the nation appearing on top of the map is nice. It just looks good and gets it out of the messy badge row.

But we need to decide where game mode, nmr/cd status, and orders status should appear.

Side note, I think showing NMR is redundant if we are already showing CD, so they shouldn't appear together.

> I don't like the "submit to rejoin" in the Civil disorder header. For a new player, it's both unclear what CD is, as well as 'what does submit to rejoin mean"? Even I don't know - does it mean I need to submit orders? Confirm them?

I agree. It's kinda meaningless. We should have something more informative here.

> I think this would be better be "Civil disorder" with a question mark icon.

Yep, agreed. I like this suggestion.

---

The overall reason for adding this PR was to split up the messy badge row, which to me is a bit of a mess, and also to remove some info from the card, which seems a little bit redundant. 

But it's a bit of a distraction, so I think it probably belongs in a discussion item for now.

PR #969, opened 2026-07-03
Add always-populated admin field to Game and collapse manager checks

johnpooch:
## What this PR does

Adds a `Game.admin` FK to `User` (`SET_NULL`, mirroring `game_master`/`created_by`), always populated at creation — `game_master` for GM games, `created_by` otherwise — and backfilled on existing rows via `COALESCE(game_master, created_by)`. `can_manage()` and `IsGameManager` collapse to a single `admin_id == user.id` comparison, replacing the branchy game_master/created_by/membership check. No joins or `select_related` needed since `admin_id` is a plain column on the game row.

`game_master` and `created_by` are untouched and keep their existing meanings elsewhere (`IsNotGameMaster`, `can_delete`, serializers, history).

One behavioral note worth flagging for review: because `admin` is an FK to `User` rather than `Member`, deleting a `Member` row (kick/leave) never nulls it — so a creator who leaves a non-GM pending game now retains manage rights, whereas previously `can_manage` required them to still be a current member. This is intentional per the issue's design (it's the precursor to management transfer in #900), and is covered by a new test (`test_can_manage_true_for_creator_after_leaving_game`).

Closes #967

---
_Generated by [Claude Code](https://claude.ai/code/session_01ANbysWithU9Zpx5XxeLVh9)_

PR #993, opened 2026-07-03
Reassign game admin when they leave or enter civil disorder

### LINKED ISSUE #968: Reassign game admin when the current admin leaves or enters civil disorder
## Goal

When the game admin leaves a pending game or enters civil disorder, reassign `admin` to a random eligible member and notify them. Depends on #967.

## Context

Implements the transfer behaviour from #900 on top of the always-populated `admin` field. PR #901's review established that the kick path can never remove the manager (only the manager can kick, self-kick is blocked), so the reachable triggers are leaving and civil disorder; elimination is deliberately not a trigger, since an eliminated player is still a member and can keep managing.

## Approach

- Add a small `reassign_admin()` method on `Game`: pick a random member with `kicked=False, civil_disorder=False, user__isnull=False`, excluding the current admin; if none exists, do nothing. Set `admin` and send one push notification to the new admin explaining they are now the game manager and why.
- Two explicit trigger call sites, each one guarded line (`if <exiting user id> == game.admin_id: game.reassign_admin()`): the leave path in `MemberDeleteView.perform_destroy` (`service/member/views.py`) and the civil-disorder check in phase resolution (`service/phase/models.py`).
- Random selection only — no reliability ranking — and explicit calls rather than signals.

johnpooch:
## What this PR does

Adds `Game.reassign_admin()`, which picks a random eligible member (`kicked=False, civil_disorder=False, user__isnull=False`), excluding the current admin, transfers `admin` to them, and sends a push notification explaining they are now managing the game. If no eligible member exists, it's a no-op.

Two explicit trigger sites call it, each guarded by a single check against `game.admin_id`:
- The leave path in `MemberDeleteView.perform_destroy` (`service/member/views.py`)
- The civil-disorder check in phase resolution, `Phase._check_civil_disorder` (`service/phase/models.py`)

Elimination and kick are deliberately not triggers — the kick path can never remove the manager (only the manager can kick, and self-kick is blocked), and an eliminated player is still a member and can keep managing, per the design discussion on #900/#967/#968.

One test from the precursor PR (#969) asserted that a creator retains manage rights after leaving a pending game — that was the pre-transfer behavior this issue explicitly changes, so it's been updated (`test_admin_reassigned_when_current_admin_leaves_game`) to assert the new reassignment instead.

Closes #968

## Checklist

- [x] Tests cover the change
- [ ] Screenshots embedded in the PR description for any visual changes (see CLAUDE.md) — not applicable, backend-only change

---
_Generated by [Claude Code](https://claude.ai/code/session_013isFFbk38uxhUQgDqmh4TT)_

### DIFF
--- a/service/game/models.py
+++ b/service/game/models.py
@@ -29,6 +29,7 @@ from common.constants import (
     duration_to_seconds,
 )
 from common.models import BaseModel
+from notification.tasks import send_notification
 from phase.models import Phase, PhaseState
 from phase.utils import calculate_next_fixed_deadline, FREQUENCY_INTERVALS
 from member.models import Member
@@ -592,6 +593,29 @@ class Game(BaseModel):
     def can_manage(self, user):
         with tracer.start_as_current_span("game.models.can_manage"):
             return self.admin_id == user.id
+
+    def reassign_admin(self):
+        with tracer.start_as_current_span("game.models.reassign_admin"):
+            import random
+
+            candidates = list(
+                self.members.filter(kicked=False, civil_disorder=False, user__isnull=False)
+                .exclude(user_id=self.admin_id)
+            )
+            if not candidates:
+                return
+
+            new_admin = random.choice(candidates).user
+            self.admin = new_admin
+            self.save(update_fields=["admin"])
+
+            send_notification.defer(
+                user_ids=[new_admin.id],
+                title=self.name,
+                body="The previous manager is no longer available, so you are now managing this game.",
+                notification_type="game_admin_reassigned",
+                data={"game_id": str(self.id), "link": f"{settings.FRONTEND_URL}/game/{self.id}"},
+            )
 
     def notification_user_ids(self, exclude_user_id=None):
         user_ids = {
--- a/service/game/models.py
+++ b/service/game/models.py
@@ -1,4 +1,5 @@
 from datetime import timedelta
+import random
 import re
 import uuid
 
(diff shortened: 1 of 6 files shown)


johnpooch (commented):
### PR Review

This PR adds `Game.reassign_admin()` and wires it to the two triggers described in the issue (leave, civil disorder), correctly excluding kick and elimination per the linked design discussion. I traced both call sites in `member/views.py` and `phase/models.py`: the guard conditions correctly identify only the outgoing/CD admin, the candidate filter (`kicked=False, civil_disorder=False, user__isnull=False`) correctly excludes ineligible members, and the local `import random` and raw-string `notification_type` both match existing conventions elsewhere in `game/models.py`. Tests cover the reassignment, exclusion, and no-op-when-no-candidates cases at both the unit and integration level, and the fixtures used all exist and are wired correctly.

**0** inline comment(s) (0 blocking, 0 suggestion) — no issues found.

---

service/game/models.py
```diff
@@ -593,6 +594,29 @@ def can_manage(self, user):
         with tracer.start_as_current_span("game.models.can_manage"):
             return self.admin_id == user.id
 
+    def reassign_admin(self):
+        with tracer.start_as_current_span("game.models.reassign_admin"):
+            import random
```
johnpooch: Move import to the top of the file. Update CLAUDE.md to clearly instruct future Claude sessions not to use inline imports unless required to avoid circular imports.

PR #994, opened 2026-07-04
Make GET /variants/ a shared cacheable published catalogue

### LINKED ISSUE #992: Make GET /variants/ a shared cacheable published catalogue and resolve variants by id
## Goal

Serve `GET /variants/` as a published-only, user-independent catalogue that can be shared across users and a CDN, and stop game screens depending on a user-scoped variant list — while keeping all variant board data on cacheable endpoints.

## Context

Follows the render-cache work in #963. With colorblind mode now removed, the published payload is already user-independent in content, but `GET /variants/` is still user-scoped (`visible_to`: published + own drafts + member-game variants), the ETag still includes `user.id`, and `Cache-Control` is `private, must-revalidate, max-age=60` (`service/variant/views.py`) — so it can't be shared across users. Games reference variants only by `variantId`; embedding variant data into the volatile, per-user games list would destroy cacheability, so variants must stay their own cacheable resource referenced by id. The only reason game screens currently need a user-scoped list is to resolve a draft-backed variant (a private game on the owner's draft variant), which the existing per-variant detail endpoint can serve instead.

## Approach

- Make `GET /variants/` published-only and user-independent: serve `Variant.objects.filter(status=PUBLISHED).with_related_data()`, set `ETag = sha256(variant_max|flag_max)` (drop `user.id`), and `Cache-Control: public`. Reuse the render cache from #963 as the sole path (the published-only gate query is no longer needed).
- Resolve draft-backed game variants via the existing `GET /variants/<id>/` detail endpoint (already permission-scoped via `visible_to`, so a game member can fetch it); add an ETag / `Cache-Control` so each variant is individually cacheable. The client resolves a game's variant from the cached published catalogue and, on a miss (a private game on a draft variant), fetches that one variant by id — hidden behind a small `useGameVariant(game)` hook.
- Keep the games list carrying only `variantId` (no embedded variant data).
- Give the Variants management screen a small user-scoped "my drafts" enumeration (the user's own drafts) — the only remaining user-scoped listing, kept only so other users' drafts aren't shown.
- Repoint the frontend consumers of the current `useVariantsListSuspense` hook: CreateGame / FindGames / Tutorial and the game-detail screens resolve from the published catalogue (with the per-id fallback for drafts); the management screen uses the drafts enumeration. Regenerate the API client (`docker compose up codegen`) and update the MSW handlers/fixtures (add a draft-variant fixture).
- Delete the unused `GameListVariantSerializer` (`service/variant/serializers.py`).

johnpooch:
## What this PR does

`GET /variants/` is now published-only and user-independent: the ETag drops `user.id`, `Cache-Control` becomes `public`, and the shared render cache added in #963 is the sole path (the per-user "does this user see only published variants" gate query is gone). Games keep referencing variants only by `variantId` — no variant data is embedded in the volatile per-user games list.

A new small `useGameVariant(game)` hook resolves a game's variant from the cached published catalogue and falls back to fetching it by id (now also ETag/`Cache-Control`-cacheable) for the one case that can't come from the published list: a private game running on its owner's still-draft variant. `CreateGame`, `MyGames`, and the game-detail screens (`GameInfoContent`, `PlayerInfoContent`, `OrdersScreen`, `DrawProposalsScreen`, `ProposeDrawScreen`, `ChannelListScreen`, `ChannelScreen`, `GameInfo`) are repointed accordingly. `FindGames` needed no change: it lists `can_join=true` games, which `game/filters.py`'s `filter_can_join` restricts to `private=False` and excludes games the requesting user is already a member of — draft variants can only back private games, so a joinable game's variant is always in the published catalogue for a browsing, not-yet-a-member user, both before and after this PR.

The Variants management screen gets a new `GET /variants/mine/` endpoint enumerating just the caller's own drafts, since those no longer appear in the published list, and merges it with the published catalogue for display.

**Trade-off worth flagging:** `MyGames` isn't named in the issue's approach, but it lists private games too (via `mine=true`), which can be draft-backed (a user's own sandbox or private game created from their own draft). Without the same per-game fallback there, those games would have silently disappeared from "My Games" once their variant stopped being in the published list. I applied the same `useGameVariant` treatment there to avoid that regression.

Regenerated `service/openapi-schema.yaml` and `src/api/generated/endpoints.ts` via the underlying `manage.py spectacular` + `orval` commands (no local Docker daemon in this environment); the diff is limited to the new `variants/mine/` endpoint plus a harmless interface-declaration reorder. Also deleted the now-unused `GameListVariantSerializer` and updated MSW handlers/fixtures with a draft-variant fixture.

Closes #992

---
_Generated by [Claude Code](https://claude.ai/code/session_01PrQeTNvcfKmjeBXG7gSMZf)_

### DIFF
--- a/packages/web/src/api/generated/endpoints.ts
+++ b/packages/web/src/api/generated/endpoints.ts
@@ -898,6 +898,11 @@ export interface VictoryConditions {
   drawAfterYear: number | null;
 }
 
+export interface VariantProvinceAdjacency {
+  to: string;
+  pass: string;
+}
+
 export interface VariantProvince {
   id: string;
   name: string;
@@ -906,11 +911,6 @@ export interface VariantProvince {
   /** @nullable */
   parentId: string | null;
   readonly adjacencies: readonly VariantProvinceAdjacency[];
-}
-
-export interface VariantProvinceAdjacency {
-  to: string;
-  pass: string;
 }
 
 export interface VariantTemplateNationRef {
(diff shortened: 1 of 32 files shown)


johnpooch (commented):
### PR Review

The PR correctly makes `GET /variants/` published-only and shared-cacheable (ETag drops `user.id`, `Cache-Control` is `public`, the per-user gate query is removed), adds a properly authenticated/owner-scoped `GET /variants/mine/`, and repoints the named game-detail screens, `CreateGame`, and `MyGames` to a new `useGameVariant` hook with a sound per-game fallback for draft-backed variants.

🛑 **blocking**: The PR description states `FindGames` was repointed to the new variant-resolution approach, but `packages/web/src/screens/Home/FindGames.tsx` has no changes in this diff. It still calls `useVariantsListSuspense()` (line 51) and filters via `variantMap.has(game.variantId)` (line 62), which silently drops any joinable game whose variant isn't in the published catalogue — the exact regression class this PR fixes for `MyGames` via `useGameVariant`. A variant can be archived after backing already-created public games, so those games would silently vanish from Find Games results with no fallback. Since the file has no diff hunk, I can't attach this inline — flagging it here instead. Please either repoint `FindGames` the same way as `MyGames`/`CreateGame`, or correct the PR description to reflect that it wasn't addressed.

**1** issue flagged (blocking), 0 suggestions.

---

johnpooch:
Thanks for catching the description inaccuracy — fixed (`FindGames` is no longer listed as "repointed", with an explanation of why it needed no change).

On the substance: I don't think this is a regression, and I don't think `FindGames` needs the same `useGameVariant` fallback. Walking through the actual constraints:

- `FindGames` queries `useGamesListInfinite({ can_join: true, eligible_only: true, ... })`. `GameFilter.filter_can_join` (`service/game/filters.py:37-43`) restricts this to `status=PENDING, private=False`, and additionally excludes any game the requesting user is already a member of.
- Draft variants can only ever back **private** games — enforced server-side in `GameCreateSerializer.validate()` (`service/game/serializers.py:474-485`): `"Draft variants can only be used for private games."` So a `private=False` game's variant was published at creation time, never draft.
- The only way such a game's variant could later be missing from `GET /variants/` is if it was subsequently archived. But `VariantQuerySet.visible_to(user)` — the *pre-existing* logic this PR replaces for the list endpoint — only ever surfaced a non-published variant to a user via `Q(games__members__user=user)`, i.e. only to users who are already members of a game on that variant. A `can_join` game by definition excludes users who are already members. So a browsing, not-yet-joined user could never have resolved that archived-variant game's variant from `GET /variants/` **before** this PR either — `variantMap.has(game.variantId)` already silently filtered it out under the old code.

So this specific "vanishing game" failure mode is real, but it's identical before and after this PR — not something introduced or fixed here, and arguably a separate, narrower issue (stale/archived variants on already-created public games) than what #992 scopes. Happy to be shown a path I'm missing, but I don't think it blocks this PR. Let me know if you'd like it filed as a follow-up issue instead.

---

Issue #942, opened 2026-07-01
Procrastinate worker races migrations on deploy, causing transient "column does not exist" errors


johnpooch:
Source: sentry/DIPLICITY-API-8D

**Sentry issue:** https://johnpooch.sentry.io/issues/DIPLICITY-API-8D

**Regression of:** #737 (closed as completed 2026-06-16, but the Sentry issue has since recurred — occurrences on 2026-06-30 postdate that closure)

## Error

```
ProgrammingError: column variant_variant.dominance_rules does not exist
LINE 1: ..."."phase_progression", "variant_variant"."rules", "variant_v...
                                                             ^
```

- **Task:** `phase.sweep_due_phases` (procrastinate background worker)
- **Culprit:** `phase/models.py:122` in `get_phases_to_resolve` → `with_adjudication_data()`
- **Frequency:** 9 events over 26 days (first seen 2026-06-05, last seen 2026-06-30)
- **Users affected:** 0 (background job, not user-facing)

## Root cause

`service/entrypoint.sh` starts the procrastinate worker immediately with no migration step:

```sh
if [ "$PROCESS_ROLE" = "worker" ]; then
    echo "Starting Procrastinate worker..."
    exec python manage.py procrastinate worker
fi
```

Only the `web` role runs `manage.py migrate` before starting Gunicorn. On deploy, the `worker` service's new container (already running code that references new model fields) starts querying the database while the `web` service's container is still mid-`migrate`. This is a race condition, not a permanently missing column — the previous issue (#737) confirmed via pgweb that migration `variant/0021_add_dominance_rules` was applied on 2026-05-29 and the column exists in production today.

**Confirming evidence:** a related event in the same issue group, `column game_game.created_by_id does not exist`, fired at `2026-06-12T22:10:00Z` — exactly **one second before** migration `game/0016_game_created_by` completed in production (`2026-06-12T22:10:01.613365Z`, verified via pgweb against `django_migrations`). This pins the failure window to the exact moment migrations were still being applied, ruling out a genuinely missing column.

## Fix

Have the worker entrypoint wait for `manage.py migrate --check` to succeed (bounded retry) before starting the procrastinate worker, so it never queries the database mid-migration. This fixes the race for any future migration, in both production and PR staging deploys (which share the same entrypoint script).

PR #996, opened 2026-07-04
Serialize manage.py migrate across concurrent web containers

### LINKED ISSUE #995: Concurrent `manage.py migrate` invocations race during rolling deploys, corrupting schema mid-migration
Source: sentry/DIPLICITY-API-8D

**Sentry issue:** https://johnpooch.sentry.io/issues/DIPLICITY-API-8D

**Regression of:** #942 (closed as completed 2026-07-01, but the Sentry issue has continued firing — most recently 2026-07-03T23:47:00Z, after the fix was merged and deployed)

## Error

```
ProgrammingError: column variant_variant.dominance_rules does not exist
LINE 1: ...".`phase_progression`, "variant_variant"."rules", "variant_v...
```

- **Task:** `phase.sweep_due_phases` (procrastinate background worker)
- **Culprit:** `phase/models.py:122` in `get_phases_to_resolve` → `with_adjudication_data()`
- **Frequency:** 29 events over ~29 days (first seen 2026-06-05, last seen 2026-07-03T23:47:00Z)
- **Users affected:** 0 (background job, not user-facing)

## Root cause

#942/#943 fixed one race (worker starting before web finished migrating) by making the worker poll `manage.py migrate --check` before starting. That fix is necessary but not sufficient, because `manage.py migrate` itself has no protection against **concurrent invocations** — and on 2026-07-03 at 23:27–23:28 two other Sentry issues fired in the same window that prove concurrent migrate races are still happening:

- **[DIPLICITY-API-6E](https://johnpooch.sentry.io/issues/DIPLICITY-API-6E)** — `ProgrammingError: relation "variant_variant" already exists`, raised from `manage.py migrate` itself (`CREATE TABLE`), culprit `manage.py:18` in `main`. **432 occurrences since 2026-03-29** — this race predates #942/#943 entirely and is unrelated to the worker.
- **[DIPLICITY-API-96](https://johnpooch.sentry.io/issues/DIPLICITY-API-96)** — `ProgrammingError: relation "unit_unit" does not exist`, raised from `manage.py migrate` (`ADD COLUMN`), same window, same server.

`service/entrypoint.sh`'s `web` role runs `python manage.py migrate` unconditionally on every container start with no locking:

```sh
echo "Running database migrations..."
if python manage.py migrate; then
```

When two `web` containers overlap — a rolling deploy where the old instance hasn't stopped before the new one starts, or Railway briefly running >1 replica — both run `manage.py migrate` against the same production database concurrently. Postgres serializes the conflicting `CREATE TABLE`/`ADD COLUMN` DDL at the catalog level, so the loser gets `relation already exists` (#6E) or, if the migration graph's create-step hasn't landed yet in that process's view of `django_migrations`, `relation does not exist` (#96) when it tries to alter a table it doesn't think has been created yet. Either way, one process's `manage.py migrate` exits with an error mid-run, leaving the schema (and `django_migrations` bookkeeping) in a state where a later migration — e.g. `variant/0021_add_dominance_rules` — hasn't actually been applied even though other migrations after it may have raced ahead. That's what the procrastinate worker (#8D) then hits: `migrate --check` can pass while a specific column is genuinely still missing, because `--check` is on the *worker's* view of the same racing `django_migrations` table.

## Fix

Serialize `manage.py migrate` across concurrent `web` containers using a Postgres session-level advisory lock (`pg_advisory_lock`) acquired before migrating and released after, in `service/entrypoint.sh`. This makes concurrent deploys queue for the lock instead of racing, so migrations always run to completion one at a time regardless of how many `web` replicas start simultaneously. This is expected to resolve DIPLICITY-API-8D, DIPLICITY-API-6E, and DIPLICITY-API-96 together, since they share this one root cause.

johnpooch:
## What this PR does

Wraps the `web` role's `manage.py migrate` in `service/entrypoint.sh` with a Postgres session-level advisory lock (`pg_advisory_lock`/`pg_advisory_unlock`), so concurrent `web` containers queue for the lock instead of racing each other's migration DDL.

**Problem:** the `web` entrypoint runs `manage.py migrate` unconditionally on every container start with no coordination. When two `web` containers overlap (a rolling deploy where the old instance hasn't stopped before the new one starts, or a brief multi-replica window), both invoke `manage.py migrate` against the same production database concurrently. Postgres serializes the conflicting `CREATE TABLE`/`ADD COLUMN` DDL at the catalog level, so the losing process gets `relation already exists` ([DIPLICITY-API-6E](https://johnpooch.sentry.io/issues/DIPLICITY-API-6E), 432 occurrences since 2026-03-29) or `relation does not exist` ([DIPLICITY-API-96](https://johnpooch.sentry.io/issues/DIPLICITY-API-96)) and exits mid-run, leaving a later migration unapplied even though `django_migrations` bookkeeping may look otherwise consistent. That's what the procrastinate worker then hits when it queries a column that isn't actually there yet ([DIPLICITY-API-8D](https://johnpooch.sentry.io/issues/DIPLICITY-API-8D)) — a regression of #942/#943, which fixed the worker-vs-web race but not this web-vs-web race.

**Fix:** acquire a Postgres advisory lock before calling `migrate` and release it in a `finally`, via a `manage.py shell` heredoc (matching the existing superuser-creation pattern in this file). A second container's migrate call now blocks until the first finishes, instead of racing its DDL.

Closes #995

## Checklist

- [x] Tests cover the change — no Python/TS test suite applies (shell-script-only change); verified via `sh -n` syntax check and by manually exercising the lock against the local Postgres cluster: (1) the migrate-with-lock script runs cleanly end-to-end, (2) a second concurrent process attempting `pg_advisory_lock` on the same key blocks until the first process explicitly unlocks, confirmed by timing the second process's acquisition against the first's release.
- [ ] Screenshots embedded in the PR description for any visual changes (see CLAUDE.md) — not applicable, deploy/infrastructure-only change with no user-facing surface

---
_Generated by [Claude Code](https://claude.ai/code/session_015AaZbtmQy5KR3CCiKrecrJ)_

### DIFF
--- a/service/entrypoint.sh
+++ b/service/entrypoint.sh
@@ -22,7 +22,21 @@ if [ "$PROCESS_ROLE" = "worker" ]; then
 fi
 
 echo "Running database migrations..."
-if python manage.py migrate; then
+if python manage.py shell << 'EOF'
+from django.core.management import call_command
+from django.db import connection
+
+MIGRATION_LOCK_ID = 727272001
+
+with connection.cursor() as cursor:
+    cursor.execute("SELECT pg_advisory_lock(%s)", [MIGRATION_LOCK_ID])
+try:
+    call_command("migrate")
+finally:
+    with connection.cursor() as cursor:
+        cursor.execute("SELECT pg_advisory_unlock(%s)", [MIGRATION_LOCK_ID])
+EOF
+then
     echo "Migrations completed successfully."
 else
     echo "Migration failed!" >&2


johnpooch (commented):
### PR Review

Wraps the `web` role's `manage.py migrate` in a session-level Postgres advisory lock so overlapping `web` containers serialize their migration DDL instead of racing. The change is correct and well-scoped. I verified against the local Django 6 / Postgres environment: (1) `manage.py shell` still exits non-zero when the migration raises, so a failure is detected exactly as before; (2) `call_command("migrate")` runs on the same DB session/backend PID that holds the lock, so the lock actually covers the whole migrate; (3) it's a session-level lock, so it auto-releases if a container dies mid-migration (no permanent deadlock); (4) the quoted `<< 'EOF'` heredoc correctly avoids shell expansion, and it's the only advisory lock in the codebase so there's no key collision. No issues to flag.

---

PR #1005, opened 2026-07-06
[ONE LINE CODE CHANGE] Make the map's out-of-bounds background follow the app theme

JorenC:
## What this PR does

In dark mode, the area around the game board (visible when the map is letterboxed in fit mode, while panning past the edge, or during load) stayed white. The white came from a hardcoded `background: "#fff"` on the Leaflet container div in `GameMapCanvas.tsx` — the rasterized board PNG is transparent outside painted areas, so the container color shows through.

The container now uses `background: "var(--background)"`, the theme token the rest of the app is built on, so the surround is near-black in dark mode and white in light mode as before. It stays an inline style because `leaflet/dist/leaflet.css` sets its own `.leaflet-container { background: #ddd }`, and the inline style overrides it regardless of stylesheet order.

The colored area *inside* the board rectangle comes from the variant's dSVG artwork (`rootFill` / background layer) and is intentionally untouched.

### Screenshots

**Dark mode (fixed)** — letterboxed board on `/game/active-movement/phase/101`, surround now follows the theme:

map dark mode

**Light mode (unchanged)** — surround remains white, which is also what dark mode looked like before this fix:

map light mode

## Checklist

- [x] For PRs of any significant complexity: I ran `/review-pr` against this PR in Claude Code and addressed (or responded to) its findings — n/a, one-line style change
- [x] Tests cover the change — n/a, inline style only; verified via the screenshots above in both themes

### DIFF
--- a/packages/web/src/components/GameMapCanvas/GameMapCanvas.tsx
+++ b/packages/web/src/components/GameMapCanvas/GameMapCanvas.tsx
@@ -254,7 +254,7 @@ const GameMapCanvas: React.FC<GameMapCanvasProps> = (props) => {
       <div
         ref={containerRef}
         data-map-impl="leaflet"
-        style={{ width: "100%", height: "100%", background: "#fff" }}
+        style={{ width: "100%", height: "100%", background: "var(--background)" }}
       />
       {showFillToggle && (
         <Button


JorenC (commented):
### PR Review

This PR replaces the hardcoded white background on the Leaflet map container with the app's --background theme token so the out-of-bounds area follows dark mode. I verified the token is defined on :root (white, identical to the previous #fff) with a .dark override in index.css, that nothing else in the map module depends on a white container (rasterization is independent of the container element), and that the inline style correctly overrides leaflet.css's .leaflet-container background. The change is minimal, correct, and matches the description and screenshots — no issues to flag.

PR #1007, opened 2026-07-06
Enable Public Press chat during staging (backend)

### LINKED ISSUE #902: Enable Public Press chat during staging (backend)
## Goal

Make Public Press usable during staging: joined players can post, anyone who can see the game can read, and an optional intro message is collected on join. **Public Press is the lobby** — no new channel type (decision by @johnpooch).

## Context

Replaces the earlier Lobby-channel design, which @johnpooch rejected in favour of the simplest approach: reuse Public Press, render messages without a nation pre-start, and let them re-attribute to nations automatically once the game starts.

Deliberately cut from the old plan:

- Lobby channel, type discriminator, backfill migration, staging/post-game visibility flipping, post-game re-show (#904 closed).
- Anonymity exemptions — the existing masking in `BaseMemberSerializer._is_masked` stands, so anonymous games chat as "Anonymous" pre-start (self-consistent, and showing real names would leak identity after nation re-attribution).
- Private channels during staging — channel creation stays active-game-only, avoiding the nation-based naming problem (`create_from_member_ids` uses `m.nation.name`, null pre-start).
- No-press games stay chat-blocked pre-start (`IsNotNoPressActiveGame` already excludes `PENDING`; unchanged).
- Unread-count wiring — the join path already creates `ChannelMember` rows for every public channel (`member/serializers.py`, `MemberSerializer.create`).
- **Making the intro message required** — deferred to a follow-up PR (decision by @johnpooch), so it lands as an independent change that can be verified separately and reverted easily. The rationale for eventually requiring it (Joren): it acts as a commitment filter — a player who can't be bothered to write a one-line introduction is not going to be an active player.

## Approach

- Relax the game-status permission on `ChannelListView`, `ChannelMessageCreateView`, and `ChannelMarkReadView` (`channel/views.py`) to also allow `PENDING`. `ChannelCreateView` keeps `IsActiveOrCompletedGame`. Non-member read access falls out of `accessible_to_user` (`channel/models.py:15-17`).
- Make the sender's nation nullable in `ChannelMemberSerializer` (`channel/serializers.py:16`, currently a non-nullable `NationSerializer()` that breaks for nation-less staging members).
- Add a write-only **optional** `intro_message` field to the join endpoint (`MemberSerializer.create`), creating the Public Press `ChannelMessage` in the same transaction as the member row when provided. Users are prompted for an intro on join but may skip it. Not applicable for sandbox, no-press, and anonymous games (where chat is unavailable or the sender would show as "Anonymous"). Accepted quirk: leaving a pending game cascade-deletes the member's messages, intro included.
- Notifications need no changes — the existing `post_save` signal already covers Public Press.
- Tests, then codegen and `npx tsc -b --noEmit` in `packages/web`.

## Dependencies

```
```

- **#902** — no dependencies

johnpooch:
## What this PR does

Makes Public Press usable during staging (pending games), per #902: joined players can post, and anyone who can see the game can read — no new Lobby channel, Public Press doubles as the lobby.

- `ChannelMessageCreateView` and `ChannelMarkReadView` no longer carry a game-status permission (only `IsGameMember`/`IsChannelMember` etc.); `ChannelListView` uses `permissions.AllowAny`. None of the three needs to restrict by game status — `ChannelCreateView` is untouched, so private channels still only exist once a game is active.
- `ChannelMemberSerializer.nation` is now nullable, since staging members have no nation assigned yet. The two chat screens (`ChannelScreen`, `ChannelListScreen`) that read `sender.nation.name` now fall back to the member's display name when nation is null.
- Regenerated `openapi-schema.yaml` and the frontend TS client for the nullable `nation` field.
- Added a CLAUDE.md guideline against permission classes whose condition covers every possible value of the field they inspect (discovered during review: an earlier version of this PR added exactly that).

Out of scope, per review: an optional intro-message-on-join feature was cut from this PR in favor of a simpler two-chained-request frontend approach (join, then post to Public Press) in a follow-up.

Closes #902

## Checklist

- [x] For PRs of any significant complexity: I ran `/review-pr` against this PR in Claude Code and addressed (or responded to) its findings — the `gh`-CLI-based skill isn't available in this sandbox, so I performed an equivalent manual read-through of the diff instead, and separately addressed human review comments.
- [x] Tests cover the change
- [ ] Screenshots embedded in the PR description for any visual changes — this PR is backend-only; the two frontend edits are type-safety fixes for the new nullable field, not visible changes.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -282,7 +282,8 @@ python -m pytest -n auto --reuse-db # full suite
 8. **Write tests alongside features** — not as an afterthought. All tests for a single app live in that app's `tests.py`; do not split them across multiple test modules (e.g. `tests_emit.py`).
 9. **Self-review non-trivial PRs with `/review-pr`** before requesting human review. Address or explicitly respond to all findings. Trivial PRs (typo fixes, dep bumps, doc-only) are exempt.
 10. **PR description must match the diff** — run `git diff main` and confirm every described change is visible. Do not describe work from a prior PR or session.
-11. **Python imports go at module top-level** — do not add an inline `import` inside a function/method body, even if you find an existing one nearby to copy. The only exception is breaking a genuine circular import, and that exception should be rare enough to call out in a PR description when used.
+11. **Never add a vacuous permission class** — if a Django REST `BasePermission` check's condition covers every possible value of the field it inspects (e.g. every `GameStatus`), it always evaluates `True` and gates nothing. Before adding a new permission class, check whether it excludes at least one real case; if it doesn't, don't add it — use `permissions.AllowAny` (or omit the check) instead.
+12. **Python imports go at module top-level** — do not add an inline `import` inside a function/method body, even if you find an existing one nearby to copy. The only exception is breaking a genuine circular import, and that exception should be rare enough to call out in a PR description when used.
 
 ---
 
--- a/packages/web/src/api/generated/endpoints.ts
+++ b/packages/web/src/api/generated/endpoints.ts
@@ -103,7 +103,7 @@ export interface ChannelMember {
   readonly isBot: boolean;
   /** @nullable */
   readonly commitment: string | null;
-  nation: Nation;
+  nation: Nation | null;
 }
 
 export interface ChannelMessage {
--- a/packages/web/src/screens/GameDetail/ChannelListScreen.tsx
+++ b/packages/web/src/screens/GameDetail/ChannelListScreen.tsx
@@ -34,7 +34,7 @@ const getLatestMessagePreview = (
   const latestMessage = messages[messages.length - 1];
   const senderLabel = latestMessage.sender.isCurrentUser
     ? "You"
-    : latestMessage.sender.nation.name;
+    : (latestMessage.sender.nation?.name ?? latestMessage.sender.name);
   return `${senderLabel}: ${latestMessage.body}`;
 };
 
--- a/packages/web/src/screens/GameDetail/ChannelScreen.tsx
+++ b/packages/web/src/screens/GameDetail/ChannelScreen.tsx
@@ -62,17 +62,19 @@ const buildMessageItems = (
   messages: readonly ChannelMessageType[]
 ): MessageDisplayItem[] => {
   return messages.map((msg, index) => {
-    const showAvatar =
-      index === 0 ||
-      messages[index - 1].sender.nation.name !== msg.sender.nation.name;
+    const nationName = msg.sender.nation?.name ?? msg.sender.name;
+    const previousNationName = index > 0
+      ? messages[index - 1].sender.nation?.name ?? messages[index - 1].sender.name
+      : null;
+    const showAvatar = index === 0 || previousNationName !== nationName;
 
     return {
       id: msg.id,
       body: msg.body,
       createdAt: msg.createdAt,
       sender: {
-        nationName: msg.sender.nation.name,
-        nationColor: msg.sender.nation.color,
+        nationName,
+        nationColor: msg.sender.nation?.color ?? "#808080",
         picture: msg.sender.picture,
       },
       isCurrentUser: msg.sender.isCurrentUser,
--- a/service/channel/serializers.py
+++ b/service/channel/serializers.py
@@ -13,7 +13,7 @@ Member = apps.get_model("member", "Member")
 
 
 class ChannelMemberSerializer(BaseMemberSerializer):
-    nation = NationSerializer()
+    nation = NationSerializer(allow_null=True)
 
 
 class ChannelMessageSerializer(serializers.Serializer):
--- a/service/channel/tests/test_channel.py
+++ b/service/channel/tests/test_channel.py
@@ -587,6 +587,61 @@ class TestChannelMarkReadView:
         assert response2.status_code == status.HTTP_204_NO_CONTENT
 
 
+class TestChannelPendingGameAccess:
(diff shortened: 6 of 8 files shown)


service/member/serializers.py
```diff
@@ -95,8 +106,19 @@ def create(self, validated_data):
         game = self.context["game"]
         user = self.context["request"].user
         member = game.members.create(user=user)
```
johnpooch, line 106: 💡 **suggestion**: The PR description says the intro message "posts to Public Press in the same transaction as the member row," but `create()` isn't wrapped in `transaction.atomic()` and `ATOMIC_REQUESTS` isn't set in settings — so the `Member`/`ChannelMember` writes and the `ChannelMessage.objects.create()` a few lines down are three independent commits. If the last write fails, the user ends up joined with no intro posted, silently. `game/serializers.py`'s `GameCreateSerializer.create` wraps its analogous multi-step writes in `with transaction.atomic():`; doing the same here would make the description accurate and avoid the partial-join state.

---

johnpooch (commented):
### PR Review

This PR correctly implements staging chat per #902: the new `IsActiveCompletedOrPendingGame` permission is applied consistently across list/create/mark-read views, `ChannelMemberSerializer.nation` is properly nullable, frontend fallbacks to display name are correct and match the existing `#808080` fallback-color convention, and the `IsNotNoPressActiveGame`/sandbox-join exclusions correctly stay blocking (sandbox games are always created pre-filled and active, so the join path is unreachable there anyway). Traced `public_channels[0]` — safe, since only one public (non-private) channel is ever created per game across the codebase. One suggestion found regarding a description/implementation mismatch around transactional atomicity of the join + intro-message write.

**1** inline comment(s) (0 blocking, 1 suggestion).

---

service/member/serializers.py
johnpooch: Good catch, fixed in b6483d5 — wrapped the member/channel-member/intro-message writes in `transaction.atomic()`.

---

service/common/permissions.py
```diff
@@ -28,6 +28,15 @@ def has_permission(self, request, view):
         return game.status in (GameStatus.ACTIVE, GameStatus.COMPLETED, GameStatus.ABANDONED)
 
 
+class IsActiveCompletedOrPendingGame(BasePermission):
```
johnpooch: I don't understand what the purpose of this permission class is. All games belong to this check, don't they? 

We should just remove the permission. 

We should add some guidance to claude.md to tell it to never create permissions classes that include every possible state.

service/member/serializers.py
```diff
@@ -83,6 +86,15 @@ class MemberSerializer(BaseMemberSerializer):
...
     replaceable = serializers.BooleanField(read_only=True)
+    intro_message = serializers.CharField(
+        required=False,
+        allow_blank=True,
+        write_only=True,
+        max_length=settings.CHAT_MESSAGE_MAX_CHARS,
+        error_messages={
+            "max_length": f"Messages cannot be longer than {settings.CHAT_MESSAGE_MAX_CHARS} characters."
+        },
+    )
```
johnpooch: This doesn't seem like the right approach to me. 

This doesn't seem like the right approach to me. I feel like a simpler approach would be to have the front end call the join game endpoint and then immediately call the create message endpoint. Which would mean that we don't really need this change here at all. It could be a front end only change.

service/member/serializers.py
```diff
@@ -94,9 +106,22 @@ def get_is_game_creator(self, obj):
...
+
+        with transaction.atomic():
+            member = game.members.create(user=user)
+            public_channels = list(game.channels.filter(private=False))
+            ChannelMember.objects.bulk_create(
+                [ChannelMember(member=member, channel=ch) for ch in public_channels]
+            )
+
+            intro_message = validated_data.get("intro_message")
+            public_channel = public_channels[0] if public_channels else None
```
johnpooch: There is only ever one public channel. The logic a few lines above is misleading, and this line compounds that misleading logic. 

Either way, I think that this problem is resolved by moving all of this logic onto the frontend. 

Typically speaking, I prefer keeping business logic on the backend, but in this case I think two chained requests are the better fit than a single request. It means that there is only one way that messages are created, rather than multiple endpoints that support message creation.

service/common/permissions.py
johnpooch: Agreed, you're right — `GameStatus` only has 4 values and the check covered all of them, so it always evaluated true. Removed it in 78612cb; the three views now either carry no game-status permission (nothing in this PR needs one) or use `permissions.AllowAny` where that's the accurate description. Added the guardrail to CLAUDE.md as requested.

---

service/member/serializers.py
johnpooch: Makes sense, especially since the permission relaxation in this PR already makes the two-call flow possible. Reverted `intro_message` and the join-time message creation entirely in 78612cb — this becomes a frontend-only follow-up (join, then post to Public Press).

---

service/member/serializers.py
johnpooch: Agreed on both counts. Removed this logic entirely in 78612cb rather than fix the misleading indexing — one message-creation endpoint is the better invariant.

---

johnpooch:
Still ready for review — CI is green and up to date with `main`, just waiting on a human review pass (tracked via `workflow: needs-pr-review` on #902).

---

JorenC:
@johnpooch how is your holiday ;)

Issue #906, opened 2026-06-26
Add the Chat tab to pending games with a join intro message


JorenC:
## Goal

Add a Chat tab to pending games showing Public Press with player names and avatars, a required intro message on join, the message input limited to joined players, and live polling during staging.

## Context

Depends on **#902** (staging chat + codegen) and **#903** (pending games routed into the GameDetail shell). There is no Lobby channel: during staging the channel list naturally contains only Public Press, so no direct-open logic or channel-type branching is needed.

## Approach

- Add the **Chat** nav item for pending games in `GameDetailLayout`; it opens the normal channel list.
- Chat components (message rendering, `ChannelAvatar`, channel list) tolerate `nation: null` by falling back to member `name`/`picture`. Once the game starts, nations are assigned and older messages re-render with the appropriate nation automatically.
- Enable polling during staging: `GameDetailLayout.tsx:44` currently polls only when `status === "active"`; include pending.
- Hide the message input for non-members (the backend 403s them regardless).
- The join flow gains the **required** intro field, posted via #902's `intro_message` join parameter (hidden for sandbox, no-press, and anonymous games, matching the backend rule). Requiring it is a deliberately opinionated call (Joren): a one-line introduction is a minimal commitment filter that ensures players *will* be active — see #902.
- Add an MSW fixture for a pending game with messages; include PR screenshots per CLAUDE.md.

## Dependencies

```
#902 (backend: staging chat) ──┬─→ #906 (frontend: Chat tab + intro) ←── #903 (pending games in shell)
                               └─→ #905 (GM/admin participation)
```

- **#902** — no dependencies
- **#903** — no dependencies (can run in parallel with #902)
- **#906** — depends on #902 and #903
- **#905** — depends on #902; its frontend display reuses #906's null-nation rendering

JorenC:
@johnpooch Rewritten as the simplified implementation per your direction on #902 — I think we can push ahead with this:

> All we would have to do is update the chat message component to be able to render a member that does not have a nation. That would be very straightforward.
>
> Once the game starts, the nations would be assigned, and then the previous messages would all appear with the appropriate nation.

That's exactly what this is now: a Chat tab on pending games showing the normal channel list (only Public Press exists during staging), with null-nation senders falling back to player name/avatar.

One opinionated addition: the join intro message is **required** (see the body and #902) — a one-line introduction is a minimal commitment filter that ensures players will actually be active.

JorenC:
Relevant

[...]

PR #1008, opened 2026-07-06
Migrate pending games into the GameDetail shell

### LINKED ISSUE #903: Migrate pending games into the GameDetail shell
## Goal

Route pending (staging) games through the existing GameDetail shell with a working Game Info tab, replacing the standalone Home game-info screen. Non-members can open the pending shell too (read-only), per @johnpooch's spectator stepping-stone idea — scoped to pending games only.

## Context

Pure frontend navigation, independent of the backend staging-chat work (#902). Pending games already have a `PENDING` phase (created by `create_from_template`), so the phase-based shell routes resolve without inventing a placeholder.

## Approach

- Update `getGameLandingPath` (`util.ts:20`) and `GamePhaseRedirect` so pending games land on `…/phase/:phaseId/game-info` in the shell instead of the Home `/game-info/:gameId` screen.
- `GameDetailLayout`: pending games show only a **Game Info** nav item — Map and Orders are omitted, mirroring the existing sandbox nav filter. The **Chat** tab is added in #906.
- Lift the pending-action block (join / leave / share / reliability-gated states, plus mutation side-effects: query invalidation, `checkNotificationPermission()`, toasts) from `Home/GameInfo.tsx` into the shared `GameInfoContent` so both entry points render identically, then retire the Home `/game-info/:gameId` route.
- Spectator scope is pending games only: game retrieve already works for non-members, and combined with #902's channel-list relaxation that is the complete non-joined view. Active-game spectating stays out of scope.

## Dependencies

```
```

- **#902** — no dependencies

johnpooch:
## What this PR does

Pending (staging) games now render inside the existing GameDetail shell using the exact same structure active games already use, rather than a bespoke layout:

- **Desktop**: a narrow Game Info panel on the left (occupying the same slot Orders uses for active games), with the full board always visible on the right.
- **Mobile**: `Map | Game Info` bottom tabs (Chat is still deferred to #906, per the issue's dependency chain).

The map is reachable and shows the real starting deployment — `create_from_template` places units immediately on the backend — but it's read-only. `GameMap` already collapses into a no-op click handler whenever order options are empty, so the only change needed to make it inert for a pending game was skipping the options fetch (which 403s under `IsActiveGame` anyway); no new interaction-blocking logic was required.

`GameInfoContent` drops its own map preview now that the real map is always shown alongside it. `GameInfoScreen` gets a `GameDropdownMenu` when the game is pending, matching how the Map/Orders screens already surface it for active games — this also restores the Delete/Clone actions that were unreachable from a pending game's own view. Home's now-redundant `PlayerInfo` screen is retired alongside `GameInfo`; `GameCard`'s dropdown routes both through the shell via a shared `getGamePhasePath` helper.

The Map tab's phase selector was also replaced with the game name for pending games, since the phase never changes while a game is staging — there's nothing to select between. The join/leave/share/add-bot row now appears directly beneath it on mobile too, in addition to the Game Info panel. That row's mutations (join, leave, share, add-bot sheet state) were extracted into a shared `PendingGameActions` component so `GameInfoContent` and `MapScreen` don't duplicate them.

This direction came out of a design discussion on the PR with @johnpooch and @JorenC after an earlier iteration (isolated single-tab shell, no visible map) didn't match the intended UX — see the thread below for the full back-and-forth.

### Screenshots

**Desktop** — pending game the current user manages: narrow Game Info panel, full board on the right showing the starting deployment:

desktop

**Mobile** — `Map | Game Info` tabs:

mobile

**Mobile Map tab (member)** — game name + Add AI player/Leave/Share row in place of the phase selector:

mobile map member

**Mobile Map tab (non-member)** — Join game state:

mobile map join

**Full panel (mobile, tall viewport)** — confirms no leftover map preview anywhere in the Game Info panel; content ends cleanly at Rules:

bottom of panel

**Join state (Game Info panel)** — pending game the current user hasn't joined:

join

Closes #903


johnpooch (commented):
### PR Review

This PR routes pending games through the shell's Game Info tab instead of the standalone Home screen, retires the old `/game-info/:gameId` route, and moves the join/leave/share/add-bot action block into the shared `GameInfoContent` so both entry points render identically.

I traced the key correctness-sensitive paths:
- `getGameLandingPath`/`getGameInfoPath` navigation logic, including the fallback to `/` when `currentPhaseId` is null — matches updated test expectations and the mock fixture fix (pending games now always carry a real phase, consistent with backend's `create_from_template` guarantee).
- Spectator access: confirmed `GameRetrieveView` and `PhaseRetrieveView` both use `permissions.AllowAny` server-side, so the new non-member spectator path in `GameInfoContent` (join button, map preview via `useGamePhaseRetrieve`) doesn't hit a permission wall.
- `GameDetailLayout`'s nav-item filtering and conditional map panel/content-width logic for `status === "pending"` — no dangling references to the removed Home `/game-info` route or `Home.GameInfoScreen` export.
- Fixture builder change (`makeGame` now takes `phases[phases.length - 1]` as current phase instead of searching for "active") — verified against all existing callers in `games.ts`; order is always chronological with the active/final phase last, so no behavior change for existing fixtures.

Ran `npm run test` (510 tests, all passing), `npx tsc -b --noEmit` (clean), and `npm run lint` (clean).

No issues found — nothing to flag.

---

JorenC:
<@688641092278681600> I feel like this change is totally wrong: https://github.com/johnpooch/diplicity-react/pull/1008

What do you think? I would have assumed the player could see the map during the pending state. This looks totally weird.

Also, these three PRs all feel like a stack:
https://github.com/johnpooch/diplicity-react/pull/1009
https://github.com/johnpooch/diplicity-react/pull/1008
https://github.com/johnpooch/diplicity-react/pull/1007

I feel like we should address these one at a time. Otherwise it's overwhelming
--白尤恩 — Yesterday at 22:53
I thought the same, but actually if you look at the current pending/game info its the same screen. The map comes just below the rules section - it should be there but is just below the fold
This is a stack- there are two that Claude could do in parallel, but the bot automatically took the additional ones and implemented them in one go
If you look at the stories, you can see i only added two (when there were 4 PRs only) issues to a PR
And the idea was to check more later. 

First i think only "add backend" and "add game-info as shell" (so later chat - and possibly map - can be added) should run
I say this from my memory
But the bot took up more stories, i think by reading the dependencies and connections
image
image
Current game info screen has the same.
The stories make 'game info* as a tab, similar to chat so later we can add chat
I believe
And make it simple, same structure
Johnpooch — Yesterday at 22:57
But I don't understand what we're doing with this first PR. Why are we moving into the game detail shell? My vision of how this would work was totally different.

Effectively, in my vision, the game detail screen becomes the lobby where the players are waiting for the game to start. It would appear the exact same as it does after the game starts, but the player just can't interact with the map yet.
白尤恩 — Yesterday at 22:57
I dont understand that?
"becomes the lobby"
In my view, we have map/orders/chat/game info
We just disclose game info now as an option
Staging games SHOW the game info and chat (same as a normal game, just hide orders and map)
Running games SHOW the chat, orders and map, hide the game info (entrance via overflow)
That way they all use the same logic, we don't need new UI (if the lobby is the game info, then where does the chat UI live?)
So a staging game has the game info screen/tab on mobile, and a chat one. The chat one is populated with only the Public Press
Keeps everything the same. Consistent.
Simple
Johnpooch — Yesterday at 23:00
Why do we need to hide the map and orders? Why not just have the orders show a message that explains that the game is pending?
白尤恩 — Yesterday at 23:01
That is also possible, but seems to be (a) extra use case (need to make that UI which seems more complex than just hiding it) and (b) shows unnecessary information/clutter to the user
Why not just hide it?
I mean. The map maybe, but you already made the zoom and thingy in the map preview so it seemed a bit overkill to have two maps
And it means we have four navigation items in the mobile bottom bar - once again, I need to PRESS THEM FIRST to figure out what is "behind the curtain" only to find out "sorry! Nothing here until the game starts!"
Then why not just hide the entrance? User sees two things. 

Game info: all about the game and zoomy map
Chat: public press chat interface (this is an unnecessary extra click but (a) is consistent with the pattern across the whole platform and (b) keeps it simple, no customized UI.


Ideally I'd have the "chat" tab open the public press CONTENT immediately, but I thought that was unnecessary complexity.
And could be a later improvement
Johnpooch — Yesterday at 23:11
To be honest, I strongly disagree with this direction. Here are my thoughts:

The game info screen looks horrible when stretched out like it is in the PR.
The UI in the PR breaks the pattern that is followed everywhere else in the GameDetail layout (command panel to the left, map to the right)
The proposed direction is more complex than it needs to be.
The way the UI changes after the game starts bugs me. Right now the game detail screen has everything, orders, chat, map, player info, game info. All we need to do is make the map uninteractive.
Making the map uninteractive and having a placeholder for orders panel is a pre-requisite for the spectator view anyways.

This just doesn't feel properly thought through. By far the simplest approach is to do what I'm describing.
We need to keep the number of branching paths low. The way I'm thinking about this, we use effectively the same ui for:
pending game (member)
active game (member)
pending game (non-member)
active game (non-member)
白尤恩 — Yesterday at 23:14
....but I think that is what I'm describing too?
I don't even see a distinction between member and non member....
Let me climb out of bed
Mock something up
Johnpooch — Yesterday at 23:15
You don't need to do that!
Do what you like, but I really don't think there is any need to mock anything up. All we need to do is show the game detail screen instead of the game info screen when someone clicks on a game. And then just add a notice to explain that the game will start soon, etc.
I'm okay with just hiding the orders tab if we like too. That wouldn't bother me much.
白尤恩 — Yesterday at 23:24
This is what we have NOW. The left is the game screen (with sidebar - bottom on mobile). The right is the pending game screen (with the join button)
image
This is the way I see it The left being the 'game info screen' - so it goes INTO a game when you open a game, rather than stay on the 'community' tab - consistent with all other games you enter (ongoing/finished). 

When you're INTO a game it now shows the 'info' sidebar item (bottom on mobile) where you can see the same info as now, same join/leave/share info - it's the same for members/non-members, only the join element in the game info changes (like now).

But in ADDITION you have the 'Chat' sidebar item (bottom on mobile) where you can see the chats (same for members/non members, only members can add to chat). Remove the 'add channel' in this mock up - I forgot to remove that 
image
It actually improves a navigation issue
Because right now, if I go to 'find games', and I click a game, it replaces the find-games content. There is no 'back' button to go where I came from - so I need to go back to the find games top hierarchical layer, and find the next game.

The same for a pending game - 'My games > Staging games > Game content. If I want to go to another staging game, the navigation should be Game content A > Staging Games > Game content B, but I can't - I have to go to Game Content A > My Games > Staging Games > Game Content B'. 

--- but with this new game shell, I can just press the back button in the top left corner, and I'm back in Staging games (or wherever I came from). This is a bit of a nit but it's additional improvement. Consistency in code, in approach, and in UX
白尤恩 — Yesterday at 23:31
This makes more sense (back button)
image
This makes less sense (no back button) - and is current behaviour. It's the same thing, just with the 'Game content' being consistent and the same
image
No special UI for special cases, no notifiactions that stuff doesn't exist etc
I don't see the issue? Or is there some magical code difficulty I'm missing? I thought you were all for consistency and simplification?
(Note the small red 'i' I made in the sidebar - that would be a game info or game settings or something icon)
So I think we're actually describing the same situation. Maybe Claude didn't interpret or capture it correctly. 

The UI when you join a game, I would actually make a pop-up dialog (I know you hate them, but this warrants one as it's an 'important decision' and you can provide the answer. This is the right moment for it) where you write your first message. It then pops up in the Public Press channel, and new messages appear like an unread message as now - same channel, same notifiaction system, same indicators...
That I haven't discussed with Claude yet becuase that would be implemented last - first just have the chat available, and people who are in the game can chat. That's the MVP - the next story is adding the 'Write your intro message' dialog, which is a small PR I think
白尤恩 — Yesterday at 23:38
(Last thing and I'll climb back into bed) - the only question is whether you'd hide the orders tab or not
image
Map is sort of :/ I don't care - it can be there (when chat is selected, as now) on desktop, but it means we need to have a navigation item on mobile for it and I sort of don't like that
But I can go either way
That would become this (if we don't hide orders). I'd rather have INFO | MAP | CHAT   which changes into MAP | ORDERS | CHAT - consistency, etc etc etc
image
Anyway, please read the above - and if you disagree we can go back to making it a discussion 😄
I'm back to bed
I read the issues as the above btw
Backstabbr #release-notes
SERVER
 — 03:13
With apologies for a little blatant advertising (#sorrynotsorry), I just shared a post with our Patrons providing some details AND a sneak preview (a real, live, see it in your browser preview) on an exciting new feature in the works: skill ratings! And also, a few other things in the pipe. If you are just dying to know more, consider signing up at Patreon! It helps us pay the server costs. (Very occasionally, it also buys us beer. Maybe some nachos. But, like... 99% server costs.)
https://www.patreon.com/backstabbr/posts/we-finally-built-163199950
Patreon
[We Finally Built a Rating System. Turns Out, You’re All Cheaters....](https://www.patreon.com/backstabbr/posts/we-finally-built-163199950)
Hello, distinguished patrons! I know that it has been awhile since we posted anything here, but please don't be frightened. We're hard at wo
We Finally Built a Rating System. Turns Out, You’re All Cheaters....
Johnpooch — 06:50
Okay what you’re suggesting isn’t too far from what I’m thinking. 

Here’s what I think:

most minimal solution is to just render the pending game within game detail layout, making the orders panel just include a notice. Game info panel is available via sub menu, not ideal

a natural enhancement on top of that is to replace the orders panel (useless before game starts) with the game info panel which is more useful before the game starts.

I think my main objection is that the game info screen should be a panel in the left command section, keeping the existing layout. Naturally this would mean removing the map from the game info screen.

The join game button fits naturally where the confirm orders button is on the orders screen.
I agree that the problem with keeping orders and info for both pending and active is that the ordering needs to swap when the game starts which feels weird. I think demoting game info to the sub screen (behind menu) when the game starts feels natural and consistent with what we have at the moment
Johnpooch — 07:00
So then the game info and player info screens in the home layout become a bit redundant. Which is good. I feel like we can get rid of them.
白尤恩 — 09:40
Yes, I think we actually agree - as usual

most minimal solution is to just render the pending game within game detail layout, making the orders panel just include a notice. Game info panel is available via sub menu, not ideal
Considering the game info panel is where you join/leave/share a game, I think this must be the first page people see when they go to a pending game. Non-negotiable, as that is the NBA and everything else is secondary. 
a natural enhancement on top of that is to replace the orders panel (useless before game starts) with the game info panel which is more useful before the game starts.
That is indeed what I suggest - IF pending, hide orders. IF running/finished, hide game-info
I think my main objection is that the game info screen should be a panel in the left command section, keeping the existing layout. Naturally this would mean removing the map from the game info screen.
I don't know what you mean with 'left command section'? In my view, this would just replace ALL the left content and render the game-info in the full screen, like my mock up. Or, is that very hard? 

An alternative would be to have it in the small side panel, remove the map and leave the map on the big screen. For mobile, that would indeed be the 'map tab'. Both are fine by me
白尤恩 — 09:46
That I don't like so much, tbh. It introduces unnecessary change, requires us to shift logic, and I like it on the game details - it has the  rules, players, times etc - so it ensures the user is aware of what they sign up for.
白尤恩 — 09:46
Yes, we 'hide' game info but it remains accessible through the overflow menu



Claude, can you update this with the following feedback?

johnpooch:
Read the full discussion. Here's what I believe is agreed, and the one point I want to confirm before reworking, since it's still open as of the last message:

**Agreed (I'll implement this):**
- Pending games render inside `GameDetailLayout` (already true), but instead of showing *only* a Game Info tab with everything else hidden, the tab **swap** applies: `Map | Orders | Chat` becomes `Map | Game Info` while pending (Chat still deferred to #906), and reverts to `Map | Orders | Chat` once active, with Game Info demoted back to the overflow menu — same tab count/position, no new UI, no notice-in-orders-panel.
- Map stays reachable for a pending game and renders read-only/uninteractive (units already sit at their template starting positions per `create_from_template`, so there's real content to show, not a blank board).
- Home's `/game-info/:gameId` and `/player-info/:gameId` screens are redundant now and can go — I'll fold this in.

**Open question — Game Info's own layout:**
johnpooch: Game Info should live in the narrow left command panel (like Orders does), dropping its own map preview since the real map is already visible on the right/via the Map tab.
JorenC (last message, no reply yet): prefers keeping Game Info as the full "game details" experience — own map preview, rules, players, times — not confined to the narrow panel, since that's an unrelated shift in scope/logic.

These produce different amounts of rework and I'd rather build the right one once. @johnpooch — can you confirm which of these two you want, or if you and Joren want to settle it in the thread first? I'll hold off on that specific piece until then, but will start on the tab-swap + Map-uninteractive-view + Home-screen-removal now since those seem settled.

---

JorenC:
Let's have Game-Info live in the small panel and render the full map next to it. In that case, remove the map preview from the narrow Game-Info.

This means that on mobile, we DO have a 'map' tab, similar to how it currently works for running games.


So:
Staging:
Desktop: GAME-INFO | CHAT
Mobile: MAP | GAME-INFO | CHAT

Running/finished:
Desktop: ORDERS | CHAT
Mobile: MAP | ORDERS | CHAT

This should be okay to go - you can proceed, no need to wait for John

johnpooch:
Implemented the agreed design:

- Desktop: `GAME-INFO` narrow panel (Orders' slot) + full map on the right, always visible.
- Mobile: `MAP | GAME-INFO` tabs (Chat still deferred to #906).
- Map is real (starting deployment already placed by `create_from_template`) but read-only — no order interactions possible for a pending phase.
- `GameInfoContent` no longer carries its own map preview.
- Home's `GameInfo`/`PlayerInfo` screens are both retired now that they're redundant.

Screenshots updated in the PR description. Tests/lint/typecheck all green. Ready for another look whenever you get a chance.

---

JorenC:
The desktop version of the 'status' (Leave/Join/Share/Add AI player) needs to be changed to the mobile version; before it had the full width of the desktop, but now it is the side panel, it has the same space as a mobile. So - it needs to use the mobile version of that component.

Also, there should be some horizontal padding to the sides of the panel - right now it's very cramped.

johnpooch:
Good catch — the buttons were using Tailwind's viewport `sm:` breakpoint, which stayed "desktop" even though the actual panel is sidebar-width now. Switched to `@[500px]:` container queries (matching the threshold `ScreenCard` already uses for this same panel) so the layout responds to the panel's real width instead of the browser's. Also added horizontal padding to the panel — it was relying on padding from the old Home layout wrapper that doesn't exist in the GameDetail shell.

Updated desktop screenshot in the PR description.

---

JorenC:
Finally, can you add a screenshot of the bottom of the game-info screen - to validate that there is no 'map' anymore?

Secondly, it now says 'Game info' in the top - can we make that the NAME of the game (so we move that up and remove 'Game info'. Truncate the name.  (Then you also don't need the name duplication in the bottom of 'Gathering forces' - which is the game name).

johnpooch:
Both done:

- Added a tall-viewport screenshot to the PR description ("Full panel") confirming the content ends cleanly at Rules with no leftover map anywhere in the panel.
- Header now shows the game's own name (truncated) instead of "Game Info", and dropped the duplicate name that was in the metadata card below.

---

JorenC:
Add some slight vertical padding to the top of the 'join/leave/add AI player' bracket so it doesn't directly touch the top bar.

johnpooch:
Added a small gap above the alert box so it no longer touches the separator line. Updated desktop screenshot in the PR description.

---

JorenC:
Right now we have the phase selector in the map screen during staging games (on mobile). This is not relevant, as the phase is always stuck at the first one. Replace that with the game name, and add the "join/leave" component beneath it (mobile only).

JorenC:
Can you make screenshots of the latest commit you made and update them in the PR description? This is about the mobile view/map active.

JorenC:
Commit [040c9b4](https://github.com/johnpooch/diplicity-react/pull/1008/commits/040c9b46cfcc0c4cffe5530a45044cb02e0abb69) only added the 'open game' (join game) state. Can you show the other states?

johnpooch:
Sorry for the silence — my connection to GitHub dropped for an extended stretch and I couldn't post anything until now (I kept working locally in the meantime and pushed everything as soon as it came back).

To catch up:

- **Phase selector → game name**: implemented for real (not just the mockup). The Map tab now shows the game name in place of the phase selector for pending games, with the join/leave/share/add-bot row directly beneath it (mobile only). That row's logic was extracted into a shared `PendingGameActions` component so it's not duplicated between the Game Info panel and the Map tab.
- **Both states**: `pending-map-mobile.png` (member — Add AI player/Leave/Share) was already committed in the same push as the feature; it just wasn't visible because I couldn't update the PR description. Both it and the join-state screenshot (`pending-map-mobile-join.png`) are now in the description above.
- CI is green on the latest commit.

Let me know if the padding/sizing on the new Map-tab action row needs the same kind of pass the Game Info one did.

---

JorenC:
I think we're all good, except the Mobile — Game Info needs to have some padding above the join/leave/add AI bot element. For Mobile - Map view there is no padding, this is good. A bit nit picking, but I think it's important.

Specifically, it's here (I can't edit while on the move - I circled it in red)
Screenshot_20260716-082508

johnpooch:
I checked this against a clean checkout of the current HEAD (`6445846`, on top of `040c9b4` which already has the `b4087f4` padding fix from the earlier round) — the padding is there on my end:

**Mobile Game Info** — clear gap between the AppBar and the alert box:
game info padding

**Mobile Map tab** — for comparison, the action row still sits flush against the AppBar here, as intended:
map no padding

`GameInfoScreen`'s `Panel.Content` has had `px-4 pt-2` since that earlier fix, and nothing since has touched it. I can't fetch the screenshot you attached directly (this environment can't reach `github.com/user-attachments/...`), but given the code and a fresh render both show the gap, is it possible you were looking at a cached build — a previously-installed PWA or a browser cache on your phone? If you do a hard refresh / clear-cache-and-reload and it's still flush, let me know and I'll dig further — there could be a mobile-Chrome-specific rendering quirk I'm not catching in a headless screenshot.

---

JorenC:
@johnpooch I think this is good to merge.

The screenshots don't show, likely because Claude is not allowed to commit it. MAYBE (Claude, don't do this!) we should allow committing screenshots, then instruct Claude to always clean that folder before merging. That way the screenshots can be used (if my idea about the cause is correct). Can't check now, otherwise just leave it.

But let's merge this! And move towards prechat lobby!!!

johnpooch:
Rebased on `main` — this branch had fallen behind and picked up a real conflict from #1099 (the commitment-scoring system), not just a stale-bot false alarm.

`GameInfoContent`'s metadata row now shows "Commitment" (main's replacement for the old "Player reliability" row), and `PendingGameActions`' join-gating now checks `commitmentEligibility` (`committed_locked`/`low_locked`) instead of the retired `minReliability` field — matching the same pattern `GameCard.tsx` already uses for locked pending games in Find Games. The old `minReliability !== "open"` check would've been dead code going forward (that field is `"open"` everywhere now), so a locked join button reached through the GameDetail shell wouldn't have gated correctly without this fix.

Lint, typecheck, and the full test suite (555 tests) are all green on the merge commit. CI is running now.

---

JorenC:
This looks good to me, TBH.

I'd like to have the map icon in the desktop view, but that is a different
story.

I think this is mergable.

johnpooch:
When you are on the game info screen and you navigate to the player info screen and then press the X at the top of the window, it tries to go to the order screen and then crashes. 

Once that bug is fixed, I'm happy to merge this PR, but to be honest, I feel like approaching this incrementally like this is a weird fit. I feel like we haven't seen or agreed on the end state of what we're building. There are too many things on the table.

Personally, I would feel better if we designed the end state for the entire game detail shell experience using something like Figma. If we could really nail it across desktop and mobile for all screens in the game detail shell, we could pretty rapidly update the UI, probably in a single PR. 

If we had the end state clearly defined, I would be happy to move towards it either incrementally or in one big PR. My instinct is that one big PR would be the best fit. 

I think adding the backend parts incrementally makes sense.

JorenC:
That's fine as well. Back next week

johnpooch:
Found and fixed the crash. Root cause: navigating back from Player Info (or hitting any bare `/game/:id/phase/:id` URL) always rendered `OrdersScreen` on desktop / `MapScreen` on mobile via `GameIndexRoute`, with no awareness of game status. `GameDetailLayout`'s own nav already hides the Orders tab for pending games, but this index route never got that memo, so it fell straight through to Orders.

That's more than a wrong-tab bug: `OrdersScreen` calls `useGamePhaseStatesListSuspense`, which 403s for a pending game (`IsActiveOrCompletedGame` only allows active/completed/abandoned) and throws inside a Suspense boundary with no visible recovery — that's the actual crash.

`GameIndexRoute` now mirrors what `GamePhaseRedirect` already does via `getGameLandingPath`: check `game.status` before picking a screen, and send pending games to Game Info instead. Verified the fix reproduces the original bug and resolves it via a scripted Playwright repro (click Player Info's X on a pending game → lands on Game Info, no crash), plus a new test in `Router.test.tsx`. Lint, typecheck, and the full suite (556 tests) are green.

On the bigger question — happy to hold here and let you and Joren settle the Figma/end-state direction before I touch anything else on this PR. I won't push further UI changes to this branch unless asked.

---

johnpooch:
Closing this PR pending full design in the design playground. We can always open it again later or refer to it in future sessions.

PR #1009, opened 2026-07-06
Enable Public Press chat during staging

### LINKED ISSUE #902: Enable Public Press chat during staging (backend)

johnpooch:
## What this PR does

Makes Public Press usable before a game starts: joined members can post, anyone who can see the game can read, and an optional intro message is collected on join. Public Press doubles as the staging lobby — no new channel type, per the decision recorded on the issue.

- `ChannelListView`, `ChannelMessageCreateView`, and `ChannelMarkReadView` no longer require an active/completed game — a member of a pending game can now list, post to, and mark read the game's channels. `ChannelCreateView` is untouched (private channels stay active-game-only, since naming them requires assigned nations).
- `ChannelMemberSerializer.nation` is now nullable, since staging members have no nation yet.
- `MemberSerializer` (the join endpoint) gains an optional, write-only `intro_message` field. When provided, it's posted to Public Press in the same transaction as the member row — skipped for no-press and anonymous games (chat is unavailable or the sender would show as "Anonymous"). Making this required is deliberately deferred to a follow-up PR so it can be verified and reverted independently.
- Two frontend components (`ChannelListScreen`, `ChannelScreen`) fall back to the sender's name/a neutral color when `nation` is null, so existing chat rendering doesn't break once the schema changes.

**One incidental fix**: enabling staging chat means a human message in a pending game with a bot member (bots can be added to a game before it starts via `BotMemberCreateView`) now reaches the bot's reply signal, which always failed its order-options fetch (403 — game isn't active) and logged an error for every message. Added a guard so the bot only attempts a reply in active games, plus a regression test.

Screenshots aren't included — the changed rendering path (a staging-chat message with no nation yet) isn't reachable through the app's current navigation, since pending games don't route to the chat screen until #903/#906 land. It's covered by component tests instead.

Closes #902

---
_Generated by [Claude Code](https://claude.ai/code/session_016a4jiSeAknweYjEUy7Es4p)_


johnpooch:
Still active — CI is green (test/build/lint all pass) and this is labeled `workflow: needs-pr-review`, awaiting human review.

---

PR #1012, opened 2026-07-06
Trigger NMR extensions on unsubmitted orders, not unconfirmed ones

### LINKED ISSUE #1011: NMR extensions trigger on unconfirmed orders instead of unsubmitted orders
## Goal

Make the phase deadline/extension logic trigger NMR extensions based on whether a player actually submitted orders, not on whether they tapped the optional "Confirm orders" button.

## Current behaviour

At each deadline, `_check_and_apply_nmr_extensions` finds phase states with `has_possible_orders=True, orders_confirmed=False` and, for any such member with extensions remaining, burns an extension and pushes the whole phase's deadline forward. Submitting, editing, or deleting orders never sets `orders_confirmed`, so a player who entered a full, valid set of orders but did not confirm is treated identically to a no-show. In a fixed-deadline game where players submit but don't use the optional confirm button, every phase slides day after day until each unconfirmed player exhausts their extensions. The notification copy compounds the confusion by saying "did not submit orders" when the real condition is "did not confirm orders". This also contradicts the in-app `DeadlineSummary` copy, which tells fixed-time players the phase resolves "regardless of whether players confirmed their orders".

Observed in production game `new-imperialism-those-who-stand-against-it` (daily 6PM ET, 2 extensions/player): phase 1 had all 7 players submit but 0 confirm and took 2 daily deadlines to resolve; phase 2 took 3 daily deadlines; phase 3 is still stalling. Players received "some player(s) did not submit orders and used an extension" while holding screenshots of submitted orders.

## Expected behaviour

An extension should only be consumed for players who genuinely failed to submit orders (i.e. no orders / NMR), not for players who submitted but left orders unconfirmed. Confirmation should remain what it is designed to be — an optional early-resolution signal — and in fixed-time mode a phase with all orders submitted should resolve at its deadline. Notification copy should match the actual condition that triggered the extension.

## Where

`service/phase/models.py` — `PhaseManager._check_and_apply_nmr_extensions` (~line 202) and `PhaseQuerySet.filter_due_phases` (~line 74), both of which key off `orders_confirmed`. The submission signal already exists as `orders_outcome` (`received`/`nmr`) / order count via `_set_orders_outcome` (~line 383). Extension notification copy at ~lines 238 and 253.

johnpooch:
## What this PR does

`PhaseManager._check_and_apply_nmr_extensions` was granting NMR extensions and pushing the phase deadline forward for any player with `orders_confirmed=False`, regardless of whether they had actually submitted orders. Since confirming is an optional, independent toggle (`PhaseStateSerializer.update`), a player who submitted a full set of orders but never tapped "Confirm" was treated identically to a genuine no-show — causing fixed-deadline phases to slide day after day until extensions ran out.

The fix changes the check to look at whether a member has submitted zero orders (the same NMR definition already used by `_set_orders_outcome`), rather than the confirmation flag. `PhaseQuerySet.filter_due_phases`'s separate "all confirmed → resolve early" mechanism is untouched — that's a distinct, working feature (early resolution via the optional confirm button) and isn't part of this bug.

Self-review turned up a regression the first version of this fix would have introduced: civil-disorder members never submit orders, and the old `orders_confirmed=False` check happened to skip them because their phase states get `orders_confirmed` auto-set to `True` at phase creation. Switching to an order-count check would have made CD members match "not submitted" and silently burn through their extensions every deadline. Civil-disorder members are now explicitly excluded from the check, with a regression test.

Closes #1011

## Checklist

- [x] For PRs of any significant complexity: I self-reviewed this diff (the `/review-pr` skill needs the `gh` CLI, unavailable in this environment) and fixed a civil-disorder regression it surfaced
- [x] Tests cover the change
- [ ] Screenshots embedded in the PR description for any visual changes (see CLAUDE.md) — not applicable, backend-only change

### DIFF
--- a/service/phase/models.py
+++ b/service/phase/models.py
@@ -202,6 +202,8 @@ class PhaseManager(models.Manager):
     def _check_and_apply_nmr_extensions(self, phase):
         not_submitted = phase.phase_states.filter(
             has_possible_orders=True
+        ).exclude(
+            member__civil_disorder=True
         ).annotate(order_count=Count("orders")).filter(order_count=0).select_related('member')
 
         members_with_extensions = [
--- a/service/phase/models.py
+++ b/service/phase/models.py
@@ -200,12 +200,12 @@ class PhaseManager(models.Manager):
         return self.resolve_due_phases(canary=True)
 
     def _check_and_apply_nmr_extensions(self, phase):
-        unconfirmed = phase.phase_states.filter(
-            has_possible_orders=True, orders_confirmed=False
-        ).select_related('member')
+        not_submitted = phase.phase_states.filter(
+            has_possible_orders=True
+        ).annotate(order_count=Count("orders")).filter(order_count=0).select_related('member')
 
         members_with_extensions = [
-            ps.member for ps in unconfirmed
+            ps.member for ps in not_submitted
             if ps.member.nmr_extensions_remaining > 0
         ]
 
--- a/service/phase/tests.py
+++ b/service/phase/tests.py
@@ -4627,7 +4627,7 @@ class TestNMRExtensionsFixedTime:
         assert phase.scheduled_resolution > now + timedelta(hours=24)
 
     @pytest.mark.django_db
-    def test_fixed_time_confirmed_skips_extension(
+    def test_fixed_time_orders_submitted_unconfirmed_skips_extension(
         self,
         deadline_warning_game_factory,
         italy_vs_germany_italy_nation,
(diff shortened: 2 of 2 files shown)


johnpooch (commented):
### PR Review

Changes `_check_and_apply_nmr_extensions` to key off actual order submission (`Count("orders") == 0`) instead of the `orders_confirmed` flag, matching the definition already used by `_set_orders_outcome`. Civil-disorder members are explicitly excluded, since their phase states get `orders_confirmed=True` auto-set at phase creation (`phase/models.py:833`) but never submit orders — without the exclusion they'd wrongly match "not submitted" and burn extensions every deadline.

Verified by tracing `has_possible_orders`/`orders_confirmed` through phase-state creation, `_check_civil_disorder`, and `filter_due_phases` (confirmed the untouched early-resolution path is a genuinely separate mechanism, as the description claims). Checked out the PR branch and ran the full `phase/` suite: all 252 tests pass, including the 4 tests in `TestNMRExtensionsFixedTime` covering submitted-but-unconfirmed, confirmed-but-not-submitted, and the civil-disorder exclusion.

No issues found — correct fix, well-scoped, good regression coverage.

---

PR #1015, opened 2026-07-06
Cap rasterised base map to a pixel-area budget

### LINKED ISSUE #1013: Large-viewBox variants render a blank game map on some devices
## Goal

Stop variants with very large maps (e.g. Cold War 1961) from rendering a blank interactive game board on devices with tighter canvas limits.

## Current behaviour

The interactive game map rasterises the composed base SVG to a canvas at the SVG's native viewBox resolution (`rasterizeSvg(baseSvg, width, height)`, called from `GameMapCanvas.tsx`). Cold War 1961 (`cold-war-1961`) has a 4000×4000 viewBox = 16 MP — more than 2× the next-largest variant and ~8× classical. That exceeds the canvas area ceiling on mobile Safari/iOS WebKit and memory-constrained devices, so `drawImage`/`toBlob` fails, the error is silently swallowed by the `.catch(() => {})`, and with no prior base on first load the board stays blank. Smaller maps like classical (~2 MP) render normally, so affected users see the classic map work while Cold War 1961 does not.

## Expected behaviour

Every variant, regardless of map size, renders its base map on all supported devices.

## Where

`packages/web/src/components/GameMapCanvas/rasterizeSvg.ts` (primary) and its caller `packages/web/src/components/GameMapCanvas/GameMapCanvas.tsx`. Frontend only.

## Approach

Cap the canvas to a pixel-area budget (~4 MP) and scale the vector down into it via `drawImage`, preserving aspect ratio (`scale = min(1, sqrt(MAX_AREA / (width * height)))`). This is safe because the raster is only a backdrop image positioned by Leaflet from the true viewBox — independent of the PNG's pixel count — and the units and order arrows are a separate vector overlay, so downscaling does not affect alignment and only slightly softens the base landmass at maximum zoom (a trade the code already accepts). One localised change fixes all variants with no data migration. Add a unit test for the scaling maths, and consider surfacing the raster failure instead of silently swallowing it so the failure mode is observable in telemetry.

johnpooch:
## What this PR does

Large-viewBox variants such as Cold War 1961 (4000×4000 = 16 MP) can exceed the canvas area ceiling on mobile Safari/iOS WebKit and memory-constrained devices. When that happens, `drawImage`/`toBlob` fails, the error was previously swallowed silently, and the interactive board rendered blank on first load since there was no prior base to fall back to.

`rasterizeSvg` now caps the rasterised base to a ~4 MP pixel budget, scaling both the injected SVG dimensions and the destination canvas down while preserving aspect ratio (`scale = min(1, sqrt(maxPixels / (width * height)))`). This is safe because Leaflet positions the base image from the true viewBox via `imageOverlay(pngUrl, this.bounds, ...)`, independent of the PNG's own pixel dimensions — the vector unit/order overlay is unaffected, and the trade is a slightly softer landmass at maximum zoom on the largest maps, which the code already accepts for all variants.

Also swapped the silent `.catch(() => {})` for a `map.raster_failure` telemetry span, so a raster failure that does slip through is observable instead of invisible.

Closes #1013

---
_Generated by [Claude Code](https://claude.ai/code/session_01T3LZ5NoYyvRY66eYTxcptJ)_

### DIFF
--- a/packages/web/src/components/GameMapCanvas/GameMapCanvas.tsx
+++ b/packages/web/src/components/GameMapCanvas/GameMapCanvas.tsx
@@ -12,7 +12,11 @@ import type {
 import { parseDsvg } from "../InteractiveMap/dsvgParser";
 import { DiplicityMap } from "../InteractiveMap/mapRenderer";
 import { toRenderState } from "../InteractiveMap/toRenderState";
-import { recordGesture, recordInitialRender } from "../InteractiveMap/mapTelemetry";
+import {
+  recordGesture,
+  recordInitialRender,
+  recordRasterFailure,
+} from "../InteractiveMap/mapTelemetry";
 import { useDsvg } from "../../hooks/useDsvg";
 import { isNativePlatform } from "../../utils/platform";
 import {
@@ -216,8 +220,13 @@ const GameMapCanvas: React.FC<GameMapCanvasProps> = (props) => {
           containerRef.current?.setAttribute("data-map-ready", "true");
         }
       })
-      .catch(() => {
-        /* a failed raster leaves the previous base in place */
+      .catch((error) => {
+        // A failed raster leaves the previous base in place.
+        recordRasterFailure({
+          variantId: variantIdRef.current,
+          error,
+          implementation: "leaflet",
+        });
       });
 
     return () => {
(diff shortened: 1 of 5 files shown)


johnpooch (commented):
### PR Review

This PR caps the rasterised base map to a ~4MP pixel budget (`computeRasterDimensions`, preserving aspect ratio via `scale = min(1, sqrt(maxPixels / (width*height)))`), and replaces the silent `.catch(() => {})` with a `map.raster_failure` telemetry span. I verified `GameMapController.setBase` positions the image overlay from `this.bounds` (built from the true `viewBox`, not the raster's own pixel dimensions), so downscaling the raster doesn't affect alignment with the vector overlay — matching the PR's stated reasoning. `computeRasterDimensions` is well covered (unchanged-within-budget, scale-down, aspect-ratio preservation), and the new `recordRasterFailure` span is tested for its attributes. Ran the affected test files, `tsc -b --noEmit`, and `eslint` on the changed files — all clean.

---

PR #1023, opened 2026-07-06
Persist a queryable record of every notification sent

johnpooch:
## What this PR does

Adds a `Notification` model that records every push notification the backend sends, so notification history is queryable when a player reports a message that didn't match game state (e.g. "it said I submitted no orders, but my orders were in"). Today notifications are fire-and-forget FCM pushes with no server-side record, which makes those reports impossible to diagnose after the fact.

Every notification in the app already funnels through a single chokepoint — `send_notification_to_users` in `notification/utils.py` (both direct callers and the `send_notification` procrastinate task bottom out there). This persists one row per recipient at that chokepoint, capturing:

- `user`, `notification_type`, the rendered `title`/`body`, the `data` payload, and `game_id` (denormalised from the payload)
- a `delivery_status`: `sent`, `no_device` (user had no active device), `skipped` (Firebase unconfigured), or `failed` (FCM raised), plus the `error` text on failure

`game_id` is stored as a plain indexed string rather than a FK to `Game` — deliberately, so the audit record survives game deletion/pruning and can't fail on an insert for an already-removed game; it still joins to `game_game.id` in SQL. Recording is wrapped so a write failure logs and continues — the notification itself always takes priority. A daily periodic task (`prune_old_notifications`) deletes records past `NOTIFICATION_RETENTION_DAYS` (default 90, env-configurable) so the table doesn't grow unbounded.

This is intentionally a write-only audit log for now — no API/inbox surface. It's the diagnosability half of the deadline-notification work in #1011/#1017; a future in-app notification centre could read from the same table.

## Checklist

- [ ] For PRs of any significant complexity: I ran `/review-pr` against this PR in Claude Code and addressed (or responded to) its findings — not yet run; happy to run it on request
- [x] Tests cover the change
- [x] Screenshots embedded in the PR description for any visual changes (see CLAUDE.md) — N/A, backend-only, no visible surface

## Verification

- New `notification/tests_records.py` (11 tests): the `_delivery_status` matrix (all four statuses), one-row-per-recipient, content/`game_id` persistence, `skipped` status without Firebase, empty `user_ids` no-op, caller `data` not mutated, recording failure doesn't break the send, and prune deletes only past-retention rows.
- `cd service && python -m pytest notification/ phase/ game/ member/ channel/ draw_proposal/ victory/ integration/` → all green except 3 pre-existing `integration/test_bot.py` failures that also fail on clean `main` (bot LLM/options `403`, environmental — unrelated to this change).
- `manage.py check` clean; `makemigrations --check` reports no missing migrations for `notification`.

https://claude.ai/code/session_01TFgqgLfj5EVHNveWFbsXL4

---
_Generated by [Claude Code](https://claude.ai/code/session_01TFgqgLfj5EVHNveWFbsXL4)_


service/notification/tasks.py
```diff
@@ -17,3 +20,15 @@ def send_notification(user_ids, title, body, notification_type, data=None):
         notification_type=notification_type,
         data=data,
     )
+
+
+@app.periodic(cron="17 3 * * *")
+@app.task(name="notification.prune_old_notifications")
+def prune_old_notifications(timestamp):
+    from notification.models import Notification
```
johnpooch, line 28: Does this inline import need to be here or can it be moved to the top of the file?

service/notification/utils.py
```diff
@@ -1,33 +1,92 @@
...
+    if send_succeeded:
+        return NotificationDeliveryStatus.SENT
+    return NotificationDeliveryStatus.FAILED
+
+
+def _record_notifications(
+    user_ids, title, body, notification_type, data, firebase_configured, users_with_devices, send_succeeded, send_error
+):
+    try:
+        from notification.models import Notification
```
johnpooch, line 24: Does this import need to be inline?

service/notification/utils.py
```diff
@@ -1,33 +1,92 @@
...
+    if not firebase_configured:
+        return NotificationDeliveryStatus.SKIPPED
+    if not has_device:
+        return NotificationDeliveryStatus.NO_DEVICE
+    if send_succeeded:
+        return NotificationDeliveryStatus.SENT
+    return NotificationDeliveryStatus.FAILED
+
+
+def _record_notifications(
```
johnpooch, line 20: I feel like this should be a method that lives on a NotificationManager or similar.

service/notification/utils.py
```diff
@@ -1,33 +1,92 @@
...
+                    delivery_status=status,
+                    error=send_error if status == NotificationDeliveryStatus.FAILED else None,
+                )
+            )
+        Notification.objects.bulk_create(records)
+    except Exception as e:
+        logger.error(f"Failed to record {notification_type} notifications: {str(e)}")
+
+
 def send_notification_to_users(user_ids, title, body, notification_type, data=None):
```
johnpooch, line 48: I feel like having this utility function here feels like a hangover from the old approach. I wonder about a model-first approach. 

First, we create a manager and add a method for creating multiple notifications. e.g. `models.Notification.broadcast(users, **options)`

This creates all the notifcation records in a `Pending` state.

Whenever a notification is created, a signal or save method creates a task to send the notification via FCM. Once that task completes, the notification status is updated.

This is almost like an inversion of what is being done here.

johnpooch:
Closing this in favour of #1030, which rebuilds the feature with the **model-first** approach discussed in the review threads here (the `Notification` model as the source of truth via `Notification.objects.broadcast(...)`, records written `pending` up front, a batched delivery task that writes back per-recipient status). Superseded — not merging this one.

---

PR #1049, opened 2026-07-13
Guard Firefox users from crashing on non-array list data

### LINKED ISSUE #1046: Firefox users crash when list data arrives as a non-array
## Goal

Stop Firefox users from crashing when list-shaped API data (`members`, `orders`, `variants`, `games`, `devices`) arrives as something other than an array.

## Current behaviour

Seven Sentry issues (DIPLICITY-REACT-V, Y, D, X, W, J, S) all come from Firefox 152 users and share one signature: code that assumes an array receives a non-array value and throws (`.some`/`.find`/`.map is not a function`, `can't access property "id", t is undefined`). They span unrelated data lists across the app:

- `GameDetailLayout.tsx:66` — `game?.members.some(...)` (the `?.` guards `game`, not `members`)
- `OrdersScreen.tsx` — orders `.find(c => c.source.id === l.id)`
- `MyGames.tsx` — variants `new Map(j.map(...))` and games `.map(t => ... t.id)`
- `useMessaging` / `useNotificationPermissionPrompt` — devices `.some(x => x.active && ...)`

Because many different arrays are simultaneously non-arrays for the same users, this reads as one systemic cause rather than several independent missing guards.

## Expected behaviour

These screens render without crashing on Firefox even when a list value is missing or malformed, and the underlying reason Firefox responses aren't arrays is understood and addressed.

## Where

`packages/web/src/components/GameDetailLayout.tsx`, `packages/web/src/screens/GameDetail/OrdersScreen.tsx`, `packages/web/src/screens/Home/MyGames.tsx`, `packages/web/src/hooks/useMessaging.ts`, `packages/web/src/hooks/useNotificationPermissionPrompt.ts`.

## Approach

Two parts: (1) add defensive handling at the crash sites (`Array.isArray` checks / complete the optional chains) as immediate protection; (2) inspect the attached Sentry session replays to see what the network responses actually contain on Firefox (candidates: an extension / tracking-protection or a service worker returning HTML or an error object instead of JSON) and fix the upstream cause. Sentry: https://johnpooch.sentry.io/issues/?project=4511598200553552&query=is%3Aunresolved

johnpooch:
## What this PR does

Seven Sentry issues (DIPLICITY-REACT-V, Y, D, X, W, J, S), all Firefox 152 users, share one crash signature: code that assumes list-shaped API data (`members`, `orders`, `variants`, `games`, `devices`) is an array and calls `.find`/`.some`/`.map` on it without checking. This adds `Array.isArray` guards (or completes optional chains) at each crash site:

- `GameDetailLayout.tsx:66` — `game?.members.some(...)` only guarded `game`, not `members`.
- `OrdersScreen.tsx` — `buildNationGroups` and the `currentMember` lookup now guard `phaseStates`, `orders`, and `game.members`.
- `MyGames.tsx` — `games` is now filtered for non-array pages and null entries, and the eliminated-index check completes its optional chain on `game.members`.
- `useMessaging.ts` / `useNotificationPermissionPrompt.ts` — the `devices` list is guarded everywhere it's read (including the ref used by the async token-refresh listener).

One deviation from the issue's stated file list: the `new Map(variants.map(...))` crash frame it describes as "MyGames.tsx" actually lives in `FindGames.tsx` — the minified bundle's chunk name didn't line up with the source file. Fixed at the real location (`variants` and the `games` list there get the same guards), noted here so the diff makes sense against the issue text.

Root cause investigation (part 2 of the issue's approach) wasn't reachable from this session: the attached Sentry replays and spans for these issues aren't retrievable through the Sentry MCP tools available here (replay/breadcrumb lookups 404, no spans recorded on the affected traces), and every event is `handled:true` with no network-response context. This PR is the defensive-guard half only; the "why is Firefox returning non-arrays" question is still open.

Also worth flagging: the same `game.members.find(...)`/`.map(...)` pattern (unguarded) exists in several files outside this issue's stated scope — `GameCard.tsx`, `DrawProposalsScreen.tsx`, `ProposeDrawScreen.tsx`, `ChannelListScreen.tsx`, `ChannelScreen.tsx`, `PhaseGuidance.tsx`, `PlayerInfoContent.tsx`. None have a matching Sentry issue yet, so left out of scope here, but likely worth a follow-up.

Closes #1046

## Checklist

- [x] Tests cover the change
- [x] Screenshots embedded in the PR description for any visual changes — N/A, no visible UI change, only defensive guards against malformed data

---
_Generated by [Claude Code](https://claude.ai/code/session_016XxPPLdXkYoRwDszAzMgGb)_

### DIFF
--- a/packages/web/src/components/GameDetailLayout.tsx
+++ b/packages/web/src/components/GameDetailLayout.tsx
@@ -63,7 +63,7 @@ const GameDetailLayout: React.FC<GameDetailLayoutProps> = ({
           game?.totalUnreadMessageCount &&
           game.totalUnreadMessageCount > 0) ||
         (item.label === "Orders" &&
-          game?.members.some(m => m.isCurrentUser && m.civilDisorder))
+          game?.members?.some(m => m.isCurrentUser && m.civilDisorder))
           ? "•"
           : undefined;
       let path: string;
--- a/packages/web/src/hooks/useMessaging.ts
+++ b/packages/web/src/hooks/useMessaging.ts
@@ -16,6 +16,7 @@ import {
   getDevicesListQueryKey,
   useDevicesList,
   useDevicesCreate,
+  FCMDevice,
 } from "../api/generated/endpoints";
 import { isNativePlatform, isIosPlatform } from "../utils/platform";
 
@@ -43,6 +44,9 @@ const useMessaging = (): MessagingState => {
 
   const native = isNativePlatform();
   const deviceType = native ? getNativeDeviceType() : "web";
+  const devices: readonly FCMDevice[] = Array.isArray(devicesListQuery.data)
+    ? devicesListQuery.data
+    : [];
 
   const devicesRef = useRef(devicesListQuery.data);
   devicesRef.current = devicesListQuery.data;
@@ -69,7 +73,12 @@ const useMessaging = (): MessagingState => {
   useEffect(() => {
     if (!native) return;
     const listener = addTokenRefreshListener(async (newToken) => {
-      const activeDevice = devicesRef.current?.find(
+      const currentDevices: readonly FCMDevice[] = Array.isArray(
+        devicesRef.current
+      )
+        ? devicesRef.current
+        : [];
+      const activeDevice = currentDevices.find(
         d => d.active && d.type === getNativeDeviceType()
       );
       if (activeDevice) {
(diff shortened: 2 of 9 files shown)


johnpooch (commented):
### PR Review

Adds `Array.isArray` guards (or completes optional chains) at each of the crash sites named in #1046: `GameDetailLayout.tsx`, `useMessaging.ts`, `useNotificationPermissionPrompt.ts`, `OrdersScreen.tsx`, `FindGames.tsx`, and `MyGames.tsx`. I traced each guard back to its call site (`buildNationGroups`, the `devicesRef` token-refresh listener, the `variantMap`/`knownGames` filtering in `FindGames`/`MyGames`, etc.) and confirmed the fallback (`[]` or completed `?.`) is safe wherever the guarded value is subsequently read, including the one non-obvious spot — the async `addTokenRefreshListener` callback reading `devicesRef.current` instead of the memoized `devices`. New tests exercise `undefined`/non-array payloads at each site and assert against the actual fallback UI, and the imported types (`FCMDevice`, `Member`, `GameList`, `Variant`) all exist in `generated/endpoints`. The PR description accurately discloses its one scope deviation (fixing the `variants`/`games` crash in `FindGames.tsx` rather than `MyGames.tsx` as the issue text stated) and the unguarded `game.members` pattern left out of scope elsewhere.

---

PR #1062, opened 2026-07-18
Fix white screen on Android launch by wiring up Android 12 splash API

JorenC:
## What this PR does

On Android 12+ (our `targetSdkVersion` is 36, so every supported device) the launch splash is drawn by the system Splash Screen API from the `windowSplashScreenBackground`/`windowSplashScreenAnimatedIcon` theme attributes. Our launch theme only set `android:background`, which that API ignores, so users saw a default white system splash; it then auto-hid on a fixed 2.5s timer while the WebView was still white waiting for React's first paint — a long white screen instead of a splash.

Changes:

- **`styles.xml`**: set `windowSplashScreenBackground` (`#291b1b`, the color already configured for the splash in `capacitor.config.ts`), `windowSplashScreenAnimatedIcon` (the launcher icon, which also fixes the icon on pre-12 devices via the compat library), and `postSplashScreenTheme` so the activity drops back to the normal theme after launch. The legacy `android:background` line is kept for the pre-12 window background.
- **`capacitor.config.ts`**: top-level `backgroundColor: "#291b1b"` so the WebView behind the splash is no longer default white; `launchAutoHide: false` (and drop the now-unused `launchShowDuration`) so the splash covers the actual load time instead of a fixed timer.
- **`App.tsx`**: call `SplashScreen.hide()` on mount (native platforms only, matching the existing `isNativePlatform()` effect pattern). Verified against the plugin source that this releases the Android 12 `setKeepOnScreenCondition` hold and plays the configured `launchFadeOutDuration` fade.
- **`main.tsx`**: arm a 10s `setTimeout` fallback `SplashScreen.hide()` at module scope, before the init calls, so a boot failure before React mounts (chunk parse error, module-scope init exception) can't strand users on the splash now that the fixed-timer auto-hide is gone. After a normal boot the extra `hide()` is a no-op. (Added in response to a `/review-pr` finding.)

## Checklist

- [x] For PRs of any significant complexity: I ran `/review-pr` against this PR in Claude Code and addressed (or responded to) its findings — small config/theme change; behavior verified against the `@capacitor/splash-screen` Android source
- [x] Tests cover the change — native splash behavior isn't reachable from Vitest; `npm run lint` and `npx tsc -b --noEmit` pass
- [x] Screenshots embedded in the PR description for any visual changes (see CLAUDE.md) — the change is native Android launch behavior, which can't be captured in this environment; needs a quick check on the Pixel 8a (`npx cap run android`)

https://claude.ai/code/session_01VixvS5ejvYeCsEMjrwdGQb

---
_Generated by [Claude Code](https://claude.ai/code/session_01VixvS5ejvYeCsEMjrwdGQb)_

### DIFF
--- a/packages/web/capacitor.config.ts
+++ b/packages/web/capacitor.config.ts
@@ -4,6 +4,7 @@ const config: CapacitorConfig = {
   appId: "com.diplicity.app",
   appName: "Diplicity",
   webDir: "dist",
+  backgroundColor: "#291b1b",
   ios: {
     webContentsDebuggingEnabled: true,
   },
@@ -15,8 +16,7 @@ const config: CapacitorConfig = {
   },
   plugins: {
     SplashScreen: {
-      launchAutoHide: true,
-      launchShowDuration: 2000,
+      launchAutoHide: false,
       launchFadeOutDuration: 500,
       backgroundColor: "#291b1b",
       showSpinner: false,
--- a/packages/web/android/app/src/main/res/values/colors.xml
+++ b/packages/web/android/app/src/main/res/values/colors.xml
@@ -3,4 +3,5 @@
     <color name="colorPrimary">#3F51B5</color>
     <color name="colorPrimaryDark">#303F9F</color>
     <color name="colorAccent">#FF4081</color>
+    <color name="splashBackground">#291b1b</color>
 </resources>
--- a/packages/web/android/app/src/main/res/values/styles.xml
+++ b/packages/web/android/app/src/main/res/values/styles.xml
@@ -17,6 +17,9 @@
 
 
     <style name="AppTheme.NoActionBarLaunch" parent="Theme.SplashScreen">
+        <item name="windowSplashScreenBackground">@color/splashBackground</item>
+        <item name="windowSplashScreenAnimatedIcon">@mipmap/ic_launcher</item>
+        <item name="postSplashScreenTheme">@style/AppTheme.NoActionBar</item>
         <item name="android:background">@drawable/splash</item>
     </style>
 </resources>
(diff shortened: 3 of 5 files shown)


JorenC:
Draft until the splash behavior is verified on an Android device — still needs a check on the Pixel 8a (`npm run build && npx cap sync android && npx cap run android`).

---

JorenC:
@johnpooch confirmed working, please merge

packages/web/capacitor.config.ts
```diff
@@ -15,8 +16,7 @@ const config: CapacitorConfig = {
   },
   plugins: {
     SplashScreen: {
-      launchAutoHide: true,
-      launchShowDuration: 2000,
+      launchAutoHide: false,
```
JorenC, line 19: 💡 **suggestion**: With `launchAutoHide: false`, any startup failure before App's effects run (bundle parse error, an exception thrown in main.tsx module init like `initializeSentry()`/`themeStorage.initialize()`, a failed chunk load) now leaves the splash on screen forever on both platforms — the old 2.5s timer at least surfaced the broken state. Consider a native-side fallback timeout (e.g. a `setTimeout`-based `SplashScreen.hide()` in main.tsx before render, or wrapping the module-scope init calls) so a boot crash can't strand users on the splash.

JorenC (commented):
### PR Review

The PR wires the Android 12 Splash Screen API attributes into the launch theme, holds the splash until React mounts, and hides it from App.tsx. I verified against the @capacitor/splash-screen 8.0.1 Android source that installSplashScreen() is called (so postSplashScreenTheme applies), that launchAutoHide:false holds the splash via setKeepOnScreenCondition, and that hide() triggers the configured 500ms fade — the mechanism described in the PR is accurate and the theme/manifest references all resolve. The one issue flagged is that removing the fixed-duration fallback means any JS boot failure before App mounts now strands the user on the splash indefinitely instead of showing a (broken but visible) app; a fallback timeout would cap that risk.

**1** inline comment(s) (0 blocking, 1 suggestion).

packages/web/capacitor.config.ts
JorenC: Addressed in 01a282ab: a 10s `setTimeout` fallback `SplashScreen.hide()` is armed at module scope in `main.tsx`, before the init calls, so the splash always hides even if boot fails before React mounts. After a normal boot the extra `hide()` is a no-op.

PR #1065, opened 2026-07-19
Show a skeleton over the canvas map until the first base raster lands

JorenC:
## Summary

On Android the canvas map pipeline (dsvg fetch → double parse → base SVG serialisation → SVG rasterisation → PNG encode) runs sequentially on the WebView main thread, and until the first base raster lands the user sees a pure white area — reported as ~15 seconds on device. The units/orders overlay pane is deliberately hidden until that first raster (`GameMapController`), and nothing else occupies the space.

This PR adds the loading state prescribed by the `ux-patterns` skill (skeletons, never spinners, dimension-matched so the swap causes no layout shift):

- `GameMapCanvas` renders a `Skeleton` absolutely over the map container until the first base raster is applied. Phase-change re-rasters keep the old base visible and do not re-show the skeleton.
- `MapView`'s interactive/pannable `Suspense` fallback (lazy `GameMapCanvas` chunk) is now a dimension-matched `Skeleton` instead of an empty div, per the skill's standing instruction for screens being touched.

Static thumbnails (`StaticMap`) already had a skeleton; this brings the interactive game map and the expanded pannable preview dialog in line with them.

This does not make the map load faster — the underlying pipeline cost is tracked in the follow-up RFC discussion on Android map load time.

## Testing

- New `GameMapCanvas.test.tsx`: skeleton visible while the dsvg is loading; removed once the first base raster lands; not re-shown while a later phase re-rasterises.
- `MapView.test.tsx`: asserts the lazy-chunk fallback renders a skeleton.
- `npx vitest run GameMapCanvas.test.tsx MapView.test.tsx GameMap.test.tsx` — 11 passed.
- `npx eslint` on changed files — clean. `npx tsc -b --noEmit` — clean.

## Screenshots

Captured on the mocked dev server (mobile viewport, dsvg request stalled to hold the loading state):

_Screenshots to be attached: `/tmp/shots/map-loading-skeleton.png` (skeleton while the base rasterises) and `/tmp/shots/map-loaded.png` (loaded map)._

### DIFF
--- a/packages/web/src/components/GameMapCanvas/GameMapCanvas.test.tsx
+++ b/packages/web/src/components/GameMapCanvas/GameMapCanvas.test.tsx
@@ -1,0 +1,153 @@
+import { render, waitFor } from "@testing-library/react";
+import { describe, it, expect, vi, beforeAll, beforeEach } from "vitest";
+import { GameMapCanvas, type GameMapCanvasProps } from "./GameMapCanvas";
+
+const { mockUseDsvg, mockControllerInstances, rasterDeferreds } = vi.hoisted(
+  () => ({
+    mockUseDsvg: vi.fn(),
+    mockControllerInstances: [] as Array<Record<string, ReturnType<typeof vi.fn>>>,
+    rasterDeferreds: [] as Array<{
+      resolve: (url: string) => void;
+      reject: (error: Error) => void;
+    }>,
+  })
+);
+
+vi.mock("../../hooks/useDsvg", () => ({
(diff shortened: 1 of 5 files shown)


JorenC:
Closing — on-device testing showed Chrome mobile on the same phone renders the map in ~3s while the native WebView takes 10-15s, so the root cause is native-shell-specific rendering, not device speed. The skeleton/vector-base approach papered over the wrong problem; re-investigating from the top.

PR #1067, opened 2026-07-19
[ANDROID LOAD FIX] Skip the PNG encode/decode round-trip when mounting the base map raster

JorenC:
## Summary

Fixes the 10–15 second white map on Android (RFC #1066). The canvas map pipeline rasterised the board SVG onto a canvas, then encoded that ~2 MP canvas to a PNG blob (`canvas.toBlob("image/png")`) only for Leaflet's `ImageOverlay` to immediately decode it back into pixels. On an Android WebView that encode step alone takes ~5 seconds per map.

This PR removes the round-trip: `rasterizeSvg` now resolves with the canvas itself, and `GameMapController.setBase` mounts it through a small `L.ImageOverlay` subclass whose element is the supplied canvas instead of an ``. Leaflet positions and composites it identically, so pan/zoom behaviour is unchanged, and the object-URL lifecycle bookkeeping in `GameMapCanvas` disappears.

## Root cause evidence (measured on a Pixel 8a via WebView remote debugging)

Sub-step benchmark inside the live production app, classical-size board (2.4 MB SVG, 1524×1357):

| step | duration |
|---|---|
| SVG load + raster into `` | 436 ms |
| `drawImage` to canvas | 66 ms |
| `canvas.toBlob("image/png")` | **5,387 ms** |
| (alternatives: WebP 3,421 ms · JPEG 4,277 ms · `createImageBitmap` 5 ms) | |

`map.initial_render` telemetry read from the device, same variant:

- before: **10,647 ms**
- after: **79 ms**

Chrome on the same phone was always fast because its PNG encoder is quicker; desktop never noticed for the same reason. Web gets the same win for free.

## Testing

- `vitest run rasterizeSvg.test.ts MapView.test.tsx GameMap.test.tsx` — 13 passed
- `npx eslint` on changed files and `npx tsc -b --noEmit` — clean
- On-device verification on the Pixel 8a: map appears in a few seconds instead of 13 (remaining wait is API request waterfall, tracked separately in #1066)
- Desktop rendering verified against the mocked dev server: board, units, and order arrows aligned — output is pixel-identical to the previous path, so no screenshots (no visual change)

### DIFF
--- a/packages/web/src/components/GameMapCanvas/GameMapCanvas.tsx
+++ b/packages/web/src/components/GameMapCanvas/GameMapCanvas.tsx
@@ -48,7 +48,6 @@ type GameMapCanvasProps = {
 const GameMapCanvas: React.FC<GameMapCanvasProps> = (props) => {
   const containerRef = useRef<HTMLDivElement>(null);
   const controllerRef = useRef<GameMapController | null>(null);
-  const basePngRef = useRef<string | null>(null);
   const initialRecordedRef = useRef(false);
 
   const mode = props.mode ?? "interactive";
@@ -153,10 +152,6 @@ const GameMapCanvas: React.FC<GameMapCanvasProps> = (props) => {
       resizeObserver.disconnect();
       controller.destroy();
       controllerRef.current = null;
-      if (basePngRef.current) {
-        URL.revokeObjectURL(basePngRef.current);
-        basePngRef.current = null;
-      }
     };
     // eslint-disable-next-line react-hooks/exhaustive-deps -- create the map once per parsed dSVG; live values flow through refs and dedicated effects
   }, [parsed]);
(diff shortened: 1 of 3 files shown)


JorenC (commented):
### PR Review

This PR removes the PNG encode/decode round-trip by resolving `rasterizeSvg` with the canvas itself and mounting it via an `L.ImageOverlay` subclass whose `_initImage` uses the supplied canvas. I verified against Leaflet 1.9.4 source that every base-class method touching `_image` (`onAdd`, `_reset`, `_animateZoom`, `onRemove`) works identically on a canvas, that `_zoomAnimated` is set before `_initImage` runs, and that the features the override drops (`load` event, `setUrl`, `className`/`zIndex` options) have no consumers in this codebase. Object-URL cleanup for the intermediate SVG blob is retained, each overlay gets a fresh canvas, and the only test touching the file covers unchanged pure helpers (passing locally). No issues worth flagging.

Issue #1072, opened 2026-07-20
Notifications PR 3: consolidate all imperative triggers into notification/signals.py


johnpooch:
Part of #1069 (signal-driven, model-first notification redesign). Depends on PR 2's decorator infrastructure.

## Goal

Move every imperative notification trigger into signal receivers in `notification/signals.py`, so that file becomes the single, complete description of every notification trigger and no send is invoked from a view, serializer, or model method.

## Context

Notification sends are scattered across: `game_admin_reassigned` (`game/models.py:611`), `send_game_management_notification` (`game/serializers.py:36`), `game_deleted` (`game/views.py:160`), `kicked_from_staging` + `civil_disorder_recovery` (`member/views.py`), and `elimination` + `removed_from_staging` + `nmr_*` (`phase/models.py`). Two of these write via bulk operations (`_check_eliminations` uses `bulk_update`; `removed_from_staging` uses a bulk `.delete()`), and Django emits no `post_save`/`post_delete` for bulk writes — so a naive signal would never fire.

## Approach

- Move each imperative call into a signal receiver using the PR 2 decorators; still calls the existing send path (no delivery-mechanism change yet).
- Convert `_check_eliminations` and `removed_from_staging` to per-instance saves so signals fire (cardinality is low — one or two rows in practice).
- **Product decisions to call out in the PR:** (a) `removed_from_staging` changes from one grouped "removed from A, B" message to one-per-game; (b) confirm no other `Member` `post_save` receiver newly fires per elimination.
- Update test patch targets that reference old modules (e.g. `notification.signals.send_notification_to_users`, `phase.models.notification_utils`).
- Splittable by domain (game/member vs phase) if the diff is large.

johnpooch:
Closing as superseded. The consolidate-into-signals approach is replaced by the emit-driven plan in #1069: PR 3 (#1084) adds the emit dispatcher and PR 4 (#1085) routes all triggers through it and deletes the signals. PR #1079 is closed alongside this.

Issue #1071, opened 2026-07-20
Notifications PR 2: introduce the transition-decorator infrastructure


johnpooch:
Part of #1069 (signal-driven, model-first notification redesign). Establishes the pattern the rest of the work depends on.

## Goal

Add reusable `@on_<transition>` decorators to the notification app and convert one existing notification to use them, so subsequent consolidation is repetitive rather than inventive.

## Context

`notification/signals.py` currently detects state transitions with ad-hoc module-level dicts (`_game_status_cache`, `_phase_status_cache`) re-implemented per signal. The `bot` app already has the clean version: `capture_phase_status` (`pre_save`) paired with composable `@on_phase_activated` / `@on_orders_confirmed` receivers (`bot/signals.py`), where each handler reads as "on this transition, do X."

## Approach

- Add `pre_save` capture helpers + `@on_<transition>` decorators to the notification app, mirroring `bot/signals.py`.
- Convert one existing notification (`phase_resolved`) to the new decorators and retire its ad-hoc status-cache dict.
- Land the first cut of the notification convention in CLAUDE.md / the backend skill (signals-only triggering; the bulk-write footgun; the cron-vs-signal boundary).
- Behavior-preserving; still calls the existing `send_notification` / `send_notification_to_users`. Depends on nothing except optionally #1069's PR 1.

johnpooch:
Superseded by #1069's revised plan: notification signals are being abandoned, so the transition-decorator infrastructure this PR added is removed as part of the emit teardown in #1085. No action needed here — noting for traceability.

Issue #1069, opened 2026-07-20
Notification architecture: emit-driven, model-first redesign


johnpooch:
## Goal

Make the entire notifications story reasonable from one place. Route every notification trigger through a single `emit(event_type, …)` dispatcher, back it with a queryable per-recipient `Notification` record, and give system happenings a home in the timeline via a new `GameEvent` primitive. This is the structural fix for the recent wave of notification bugs (#1032–#1035) and the base the reliability epic's later commitment work builds on.

## What changed from the original plan

The original plan consolidated every trigger into `notification/signals.py` using transition decorators. Building it (#1072 / PR #1079) showed the signal approach is the wrong shape: several real triggers — GM game-delete, deadline extended, NMR extensions applied, staging kicks — are not model-field transitions, so they needed a second synthetic-signal mechanism bolted alongside the decorators, recreating the half-imperative / half-signal split this epic set out to remove.

Revised approach: **abandon notification signals entirely.** Centralize the notification *machinery* (recipient rules, copy, per-recipient records, delivery) behind a single `emit(...)` dispatcher and call it explicitly from each domain action site. This keeps "one place to look" without contorting real triggers into field-diff detection — the scattering that was the original problem was of the *logic*, not of a one-line declaration at the site where the thing happens.

- #1071 (transition-decorator infra) is superseded; the infra is removed as part of the `emit` teardown in PR 4.
- #1072 (consolidate into signals) and its PR #1079 are closed.

## Architecture

- **`emit(event_type, …)`** — one dispatcher. Looks up the event type's classification (`timeline-only` / `push+timeline` / `push-only`) and its recipient resolver, then fans out to the sibling primitives below. The classification table is the reliability doc's matrix, in code.
- **`Notification.objects.broadcast(...)`** — writes one durable per-recipient record inside the caller's transaction, then defers batched FCM delivery and writes status back. Source of truth for "did we notify, whom, did it land."
- **`GameEvent`** — system-generated timeline entry (`channel` FK, nullable `phase` FK, no sender). A plain DB row read on fetch (no realtime layer). Orthogonal to pushing: a happening may create a chip, a push, both, or neither.

`Notification` and `GameEvent` stay decoupled siblings — `emit` fans out to each per the classification. Neither drives the other: their audiences, copy, and failure modes differ (e.g. `game_deleted` pushes with no chip; `civil_disorder` gets a chip with no push; an elimination chip is game-wide but the push goes only to the eliminated player).

[...]

PR #1094, opened 2026-07-22
Route all notification triggers through emit and delete the signals

### LINKED ISSUE #1085: Notifications PR 4: route all triggers through emit and delete the signals
Part of #1069 (emit-driven, model-first notification redesign). Depends on PR 3 (the `emit` dispatcher).

## Goal

Replace every imperative notification call and every `notification/signals.py` receiver with an `emit(...)` call at the domain action site, so `emit` is the single, complete description of every trigger and no send is invoked from a signal, view, serializer, or model method.

## Context

Triggers are split today between:

- **`signals.py` receivers** — draw proposal, channel message, game start/end, phase resolved, via ad-hoc `_game_status_cache`/`_phase_status_cache` dicts and the #1071 transition decorators.
- **Imperative call sites** — `game_admin_reassigned` (`game/models.py`), `send_game_management_notification` (`game/serializers.py`), `game_deleted` (`game/views.py`), kicks + `civil_disorder_recovery` (`member/views.py`), elimination + `removed_from_staging` + `nmr_*` (`phase/models.py`).

#1069 abandons signals in favour of explicit `emit`.

## Approach

- Replace each send with `emit(...)` at its trigger site.
- **Colocate the elimination / civil-disorder writes** in `PhaseManager.resolve()` (bulk semantics kept — no per-instance conversion) and emit from that single site. This absorbs the former PR 2 (#1083), which was closed: doing the colocation blind — before `emit` existed to define the seam — produced only a cosmetic wrapper, so the restructuring lands here where the consuming `emit` calls make the right shape visible.
- Delete all `notification/signals.py` receivers, the `_game_status_cache` / `_phase_status_cache` dicts, and the #1071 transition-decorator infrastructure.
- `game_deleted` resolves recipients before the cascade delete; `game_deadline_extended`, kicks, and `nmr_extensions_applied` pass their context into `emit`.
- Still on the existing send path (PR 6 swaps it); behaviour-preserving otherwise.

**Product decision to call out in the PR:** `removed_from_staging` moves from one grouped "removed from A, B" message to one-per-game (flagged in #1072).

Depends on PR 3.

johnpooch:
## What this PR does

Closes #1085 (part of #1069, the emit-driven notification redesign; depends on the `emit` dispatcher from #1090).

Replaces every imperative notification call and every `notification/signals.py` receiver with an `emit(...)` call at the domain action site, so `emit` is now the single, complete description of every trigger — no notification is sent from a signal, view, serializer, or model save.

**Trigger sites moved onto `emit`:**

| Notification | New emit site |
| --- | --- |
| `channel_message` | `ChannelMessageSerializer.create` |
| `draw_proposal` | `DrawProposalManager.create_proposal` |
| `game_start` (+ email) | `Game.start` |
| `game_admin_reassigned` | `Game.reassign_admin` |
| `game_draw` / `game_solo_win` / `game_solo_loss` | new `Game.emit_game_ended`, called from phase resolution (victory) and draw acceptance |
| `game_paused` / `game_resumed` / `game_deadline_extended` | the game-management serializers |
| `game_deleted` | `GameDeleteView.perform_destroy` (recipients resolved before the cascade delete) |
| `kicked_from_staging` / `civil_disorder_recovery` | member views |
| `elimination` / `civil_disorder` (+ email) / `removed_from_staging` / `nmr_extension_used` / `nmr_extension_applied` / `phase_resolved` / `phase_resolved_early` | `PhaseManager.resolve` and the sites it calls |

**Colocation:** the elimination and civil-disorder writes stay where `resolve()` orchestrates them (bulk `bulk_update` semantics kept, no per-instance conversion), and the `emit(...)` calls now sit alongside those writes instead of behind signals/`on_commit`. `phase_resolved` is emitted where the resolving phase is marked complete, so it fires once per resolution (the old signal double-fired on game-ending resolutions).

**Deletions:** `notification/signals.py`, `notification/decorators.py`, the `_game_status_cache` / `_previous_phase_status` dicts, and the #1071 transition-decorator infra. `_truncate_body` moves into the emit `ChannelMessageSpec`.

**Email:** email sends are not an `emit` transport, so they remain explicit `send_email_notification.defer(...)` calls alongside the `emit(...)` at the game-start, phase-resolved, and civil-disorder sites (behaviour-preserving).

## Product decisions to call out

- **`removed_from_staging`** now sends **one message per game** (a player removed from games A and B gets two notifications) instead of one grouped "removed from A, B" message. Flagged in #1072.
- **`civil_disorder`** and **`civil_disorder_recovery`** adopt emit's `Active` / `ActiveExceptActor` resolvers (defined and tested in #1090), which **exclude members already in civil disorder** (and eliminated/kicked). The old imperative paths used `active_only`, which still included civil-disorder members — so a player who just entered CD no longer receives the "X entered civil disorder" push about themselves.

## Checklist

- [x] This PR does one thing — routing triggers through `emit` and deleting the signals
- [x] For PRs of any significant complexity: self-reviewed the diff (the `/review-pr` skill needs the `gh` CLI, which isn't wired up in this environment, so the review was manual rather than tool-driven)
- [x] Tests cover the change — trigger-site tests now observe deferred push jobs (or the routed test mock); full backend suite green locally (1928 passed, 8 skipped)
- [x] Screenshots embedded in the PR description for any visual changes — n/a, backend only


service/channel/serializers.py
```diff
@@ -47,13 +48,25 @@ def create(self, validated_data):
...
+        game = channel.game
+        sender_name = "Anonymous" if game.anonymity_active else member.name
+        emit(
+            "channel_message",
+            game=game,
+            phase=game.current_phase or game.phases.last(),
+            actor=member.user,
+            channel=channel,
+            context={"sender_name": sender_name, "body": message.body},
+        )
```
johnpooch: I think sender_name should be removed from context. Instead the EmitSpec class should be responsible for determining the sender name internally from the actor and the game.

Generally I think we should try to rely on context less. Context should be for fields that only this event needs (not all events)

service/draw_proposal/models.py
```diff
@@ -59,6 +60,15 @@ def create_proposal(self, game, created_by):
...
         )
 
+        proposer_name = "Anonymous" if game.anonymity_active else created_by.name
+        emit(
+            "draw_proposal",
+            game=game,
+            phase=phase,
+            actor=created_by.user,
+            context={"proposer_name": proposer_name},
+        )
```
johnpooch: Proposer name should be determined internally by the EmitSpec.

service/emit/link.py
```diff
@@ -18,6 +18,8 @@ def resolve(self, context):
 
 class ChannelLink(Link):
     def resolve(self, context):
+        if context.phase is None:
+            return f"{settings.FRONTEND_URL}/game/{context.game.id}"
```
johnpooch: I'm not clear why this would ever be the case. Phase always exists when a channel exists I think?

service/emit/specs.py
```diff
@@ -13,14 +13,24 @@
 from emit.transport import Push, Timeline
 
 
+def _truncate_body(text: str, max_lines: int = 3, max_chars: int = 200) -> str:
```
johnpooch: Make this a private method on the class rather than a standalone function. Or else define a general util in the `common/utils.py` file because truncation might need to be used elsewhere.

service/game/models.py
```diff
@@ -690,6 +686,38 @@ def start(self, current_phase=None, members=None):
...
+
+            user_ids = self.notification_user_ids()
+            if user_ids:
+                body = "The game has started. You can now chat with other players and submit your orders. Good luck!"
+                link = f"{settings.FRONTEND_URL}/game/{self.id}"
+                send_email_notification.defer(
+                    user_ids=user_ids,
+                    subject=f"{self.name} — Game Started",
+                    html=notification_email(title=self.name, body=body, link=link),
+                )
```
johnpooch: Why is this the only place that we use email notifications? I feel like email should be just another transport that emit handles.

service/game/models.py
```diff
@@ -690,6 +686,38 @@ def start(self, current_phase=None, members=None):
...
+            names = ", ".join(m.name for m in victory_members)
+            emit("game_draw", game=self, context={"winner_names": names})
+        elif len(victory_members) == 1:
+            winner = victory_members[0]
+            winner_ids = [winner.user_id] if winner.user_id is not None else []
+            emit("game_solo_win", game=self, context={"recipients": winner_ids})
+            emit(
+                "game_solo_loss",
+                game=self,
+                context={"winner_user_id": winner.user_id, "winner_name": winner.name},
```
johnpooch: This context can be inferred from the game. It should be handled internally by the EmitSpec

service/game/models.py
```diff
@@ -690,6 +686,38 @@ def start(self, current_phase=None, members=None):
...
+            return
+
+        victory_members = list(victory.members.all())
+        if len(victory_members) >= 2:
+            names = ", ".join(m.name for m in victory_members)
+            emit("game_draw", game=self, context={"winner_names": names})
+        elif len(victory_members) == 1:
+            winner = victory_members[0]
+            winner_ids = [winner.user_id] if winner.user_id is not None else []
+            emit("game_solo_win", game=self, context={"recipients": winner_ids})
```
johnpooch: This context can be inferred from the game. It should be handled internally by the EmitSpec. It should be another audience like `Winners`

service/game/models.py
```diff
@@ -690,6 +686,38 @@ def start(self, current_phase=None, members=None):
...
+    def emit_game_ended(self):
+        try:
+            victory = Victory.objects.prefetch_related("members").get(game=self)
+        except Victory.DoesNotExist:
+            return
+
+        victory_members = list(victory.members.all())
+        if len(victory_members) >= 2:
+            names = ", ".join(m.name for m in victory_members)
+            emit("game_draw", game=self, context={"winner_names": names})
```
johnpooch: This context can be inferred from the game. It should be handled internally by the EmitSpec

service/game/serializers.py
```diff
@@ -639,12 +625,11 @@ def update(self, instance, validated_data):
...
-            instance,
-            title=instance.name,
-            body=f"Game paused by {instance.manager_label}{actor_suffix}",
-            notification_type="game_paused",
-            exclude_user_id=actor.id,
+        emit(
+            "game_paused",
+            game=instance,
+            actor=actor,
+            context={"manager_label": f"{instance.manager_label}{actor_suffix}"},
```
johnpooch: This context can be inferred from the game. It should be handled internally by the EmitSpec

service/game/serializers.py
```diff
@@ -664,12 +649,14 @@ def update(self, instance, validated_data):
...
-            notification_type="game_resumed",
-            exclude_user_id=actor.id,
+        emit(
+            "game_resumed",
+            game=instance,
+            actor=actor,
+            context={
+                "manager_label": f"{instance.manager_label}{actor_suffix}",
+                "deadline": deadline_str,
+            },
```
johnpooch: This context can be inferred from the game. It should be handled internally by the EmitSpec

service/game/serializers.py
```diff
@@ -698,12 +685,14 @@ def update(self, instance, validated_data):
...
-            notification_type="game_deadline_extended",
-            exclude_user_id=actor.id,
+        emit(
+            "game_deadline_extended",
+            game=instance,
+            actor=actor,
+            context={
+                "manager_label": f"{instance.manager_label}{actor_suffix}",
+                "deadline": deadline_str,
+            },
```
johnpooch: This context can be inferred from the game. It should be handled internally by the EmitSpec

service/game/views.py
```diff
@@ -155,14 +154,10 @@ def perform_destroy(self, instance):
...
-                lambda: send_notification.defer(
-                    user_ids=user_ids,
-                    title=game_name,
-                    body="The game was deleted by the Game Master.",
-                    notification_type="game_deleted",
-                )
+        if is_game_master_delete:
+            emit(
+                "game_deleted",
+                context={"recipients": user_ids, "game_name": game_name},
```
johnpooch: I'm okay with this for now, but I feel like the better fix for this would be to not delete the entry at all. Instead we could have something like `Archived` which is effectively deleting a game but keeping it in the DB. Maybe something for a follow up effort.

service/member/views.py
```diff
@@ -96,19 +88,13 @@ def post(self, request, *args, **kwargs):
...
-                    notification_type="civil_disorder_recovery",
-                    data={"game_id": str(game.id)},
-                )
-
-            transaction.on_commit(send_notifications)
+            emit(
+                "civil_disorder_recovery",
+                game=game,
+                actor=request.user,
+                context={"nation_name": nation_name},
```
johnpooch: This context can be inferred from the game. It should be handled internally by the EmitSpec

service/phase/models.py
```diff
@@ -231,33 +231,25 @@ def _check_and_apply_nmr_extensions(self, phase):
...
+            if member.user_id is None:
+                continue
+            emit(
+                "nmr_extension_used",
+                game=phase.game,
+                context={
+                    "recipients": [member.user_id],
+                    "extensions_remaining": member.nmr_extensions_remaining,
+                    "deadline": deadline_str,
+                },
```
johnpooch: This context can be inferred from the game. It should be handled internally by the EmitSpec

Use actor.

service/phase/models.py
```diff
@@ -231,33 +231,25 @@ def _check_and_apply_nmr_extensions(self, phase):
...
+            )
 
-        transaction.on_commit(send_notifications)
+        extension_user_ids = [m.user_id for m in members_with_extensions if m.user_id is not None]
+        emit(
+            "nmr_extension_applied",
+            game=phase.game,
+            context={"extension_user_ids": extension_user_ids, "deadline": deadline_str},
```
johnpooch: I think we should remove the extension_user_ids from this entirely. The notification body shouldn't need to show who used the extension.

service/phase/models.py
```diff
@@ -487,35 +479,23 @@ def _notify_civil_disorder(self, phase, newly_cd_members):
...
-            if user_ids:
-                link = f"{settings.FRONTEND_URL}/game/{phase.game.id}"
-                body = f"{nation_names} entered civil disorder."
+        emit("civil_disorder", game=phase.game, context={"nation_names": nation_names})
 
-                notification_utils.send_notification_to_users(
-                    user_ids=user_ids,
+        user_ids = phase.game.notification_user_ids(active_only=True)
+        if user_ids:
+            send_email_notification.defer(
```
johnpooch: Again, this should be a transport

service/phase/models.py
```diff
@@ -533,21 +513,16 @@ def _remove_from_staging_games(self, user_ids):
...
-                user_ids=[user_id],
-                title="Removed from staging games",
-                body=f"You were removed from {game_names} because you entered civil disorder in an active game.",
-                notification_type="removed_from_staging",
+        for m in staging_members:
+            if m.user_id is None:
+                continue
+            emit(
+                "removed_from_staging",
+                context={"recipients": [m.user_id], "game_names": m.game.name},
```
johnpooch: Pass game into emit, and then get game_names from that.

service/phase/models.py
```diff
@@ -648,6 +619,36 @@ def resolve(self, phase):
...
+            return
+
+        if resolved_early:
+            body = f"{phase.name} resolved early — all players confirmed their orders."
+            subject = f"{phase.game.name} — {phase.name} Resolved Early"
+        else:
+            body = f"{phase.name} has been resolved"
+            subject = f"{phase.game.name} — {phase.name} Resolved"
+
+        send_email_notification.defer(
```
johnpooch: Use transport

johnpooch:
Addressed the review in `e7e97c2` (context-minimisation) on top of `c519fd4` (sender/proposer name, `_truncate`, `ChannelLink`). Specs are now self-sufficient — deriving manager label + deadline, winners (via `Victory`), phase name, returning nation, and NMR remaining/deadline from `game`/`actor`/`phase` — so context only carries genuinely event-specific data (`nation_names`, `body`, `game_name`/`recipients` for the post-delete cases). Email is now an `Email` transport rather than side-channel `defer`s.

Two consequences worth flagging, since they're recipient/behaviour changes that fall out of the unified design:

- **`civil_disorder` email** now goes to the spec's audience (`Active`) like the push, rather than the old `active_only`. So it no longer emails members already in civil disorder, and — because `civil_disorder` uses `NoLink` — the email has no "View Game" button (the old email had a game link). Happy to give `civil_disorder` a `GameLink` if you'd like the email CTA back, though that would also add a link to the push.
- **`nmr_extension_applied`** no longer excludes the members who used an extension (dropped `extension_user_ids`), so those members receive both the individual "you used an extension" push and the game-wide "some player(s) used an extension" push. Let me know if you'd rather they be excluded.

Full backend suite green (1929 passed, 8 skipped).

---

service/phase/tests.py
```diff
@@ -1191,7 +1196,9 @@ def test_create_from_adjudication_data_query_count_with_small_game(
         # +1 vs. the pre-cancellation-tracking count: the newly armed resolution
         # job's id is persisted onto the phase so a stale job can be cancelled
         # later if the phase resolves early or gets a new deadline.
-        assert query_count == 16
+        # +2 for the phase_resolved emit (recipient lookup plus the push and
+        # email transport defers) now that resolution notifies inline.
```
johnpooch: Remove this comment.

PR #1110, opened 2026-07-28
Add durable AgentTask record for bot orchestration

johnpooch:
## What this PR does

Bot work (plan, finalize, reply) previously went straight from a Django signal to a deferred Procrastinate job with no durable domain record — so once `retry=3` was exhausted the intent was gone, and there was no way to enqueue bot work by hand. This introduces an `AgentTask` model so `agent` gains the same "durable record of intent + status lifecycle" that its siblings `inference` (`Inference`) and `notification` (`Notification`/`NotificationDelivery`) already have.

This is **step 1** of an agreed multi-step direction. It deliberately does **not** touch `emit`, add a reconciler cron, or build a step machine — those are later steps.

### How it works

- `AgentTask` (`agent/models.py`) records `kind` (`plan`/`finalize`/`reply`), a `status` lifecycle (`pending → running → succeeded/failed`), `attempts`, `last_error`, timing, and FKs to `member`/`phase`/`channel`/`message` — mirroring `Inference`.
- Signals now create a pending row via `AgentTask.objects.enqueue(...)` and defer a single `agent.run` task keyed off the row id (`lock=f"agent-task-{id}"`).
- `agent.run` (`agent/tasks.py`) loads the row, calls `mark_running()`, dispatches to the existing `plan`/`finalize`/`reply` executors (unchanged raw-id signatures, so the mocked-API unit tests are untouched), then `mark_succeeded()` — or `mark_failed(e)` and re-raises so Procrastinate retries. Because the executors already catch their own `ApiClientError`/`InferenceError` and fall back, only genuinely-unexpected errors propagate and retry; all three executors are safe to re-run from the top.
- `enqueue` is idempotent via two partial unique constraints: `(kind, member, phase)` where `channel IS NULL` (plan/finalize) and `(kind, member, message)` where `message IS NOT NULL` (reply). Partial indexes so the NULL-distinct behaviour doesn't defeat dedup, and `get_or_create` stays race-safe under concurrent signal firings.
- `AgentTask` is registered in Django admin (add enabled), so the intent can be recorded by hand today. Automatically executing a hand-inserted pending row (e.g. a bot dropped into an abandoned game that never saw a phase-activated signal) lands with the step-2 reconciler; until then such a row is driven by deferring `agent.run(agent_task_id=…)` for it.

### Layering

`AgentTask` is the source of truth of intent; Procrastinate stays the executor; `Inference` rows remain the per-model-call records (one AgentTask may span several). Consolidating the three `agent.plan`/`agent.finalize`/`agent.reply` tasks into one `agent.run` dispatcher is why the trigger tests now assert `agent.run` + `{agent_task_id}` args.

**Deploy note:** job args changed from `{user_id, game_id[, channel_id]}` to `{agent_task_id}` and the three task names collapsed to `agent.run`, so any in-flight bot jobs at deploy time would not resolve. Low volume; drain the queue before deploying if that matters.

## Checklist

- [x] For PRs of any significant complexity: self-reviewed the diff against `main` (the `/review-pr` skill's `gh` dependency isn't wired in this environment, so the review was run manually) — no correctness issues found
- [x] Tests cover the change
- [x] Screenshots embedded in the PR description for any visual changes (see CLAUDE.md) — n/a, backend-only, no visual changes


service/agent/models.py
```diff
@@ -0,0 +1,89 @@
...
+        blank=True,
+        related_name="agent_tasks",
+    )
+    message = models.ForeignKey(
+        "channel.ChannelMessage",
+        on_delete=models.CASCADE,
+        null=True,
+        blank=True,
+        related_name="agent_tasks",
+    )
```
johnpooch: Do we need channel and message? Message already includes a link to channel.

service/agent/models.py
```diff
@@ -0,0 +1,89 @@
+from django.db import models
+from django.utils import timezone
+
+from agent.constants import AgentTaskKind, AgentTaskStatus
+from common.models import BaseModel
+
+
+class AgentTaskManager(models.Manager):
+    def enqueue(self, *, kind, member, phase=None, channel=None, message=None):
```
johnpooch: To match notification and channel event, this should be called "create_from_event". It will soon be created the exact same way that notification and channel event are (emit)

service/agent/signals.py
```diff
@@ -31,27 +31,21 @@
...
     for member in bot_members:
-        plan_task.configure(lock=f"plan-{instance.id}-{member.id}").defer(
-            user_id=member.user_id, game_id=instance.game_id
-        )
-        logger.info(
-            f"[agent.plan signal] queued plan for member {member.id} (phase {instance.id})"
-        )
+        task, _ = AgentTask.objects.enqueue(kind=AgentTaskKind.PLAN, member=member, phase=instance)
+        run_task.configure(lock=f"agent-task-{task.id}").defer(agent_task_id=task.id)
+        logger.info(f"[agent.plan signal] queued plan task {task.id} for member {member.id} (phase {instance.id})")
```
johnpooch: Remove these logs. Now that we have the record, we don't need the logs.

service/agent/signals.py
```diff
@@ -31,27 +31,21 @@
...
+    logger.info(f"[agent.finalize signal] all humans confirmed for phase {phase.id} (game {phase.game_id})")
     for phase_state in bot_phase_states:
-        finalize_task.configure(
-            lock=f"finalize-{phase.id}-{phase_state.member_id}"
-        ).defer(user_id=phase_state.member.user_id, game_id=phase.game_id)
+        task, _ = AgentTask.objects.enqueue(kind=AgentTaskKind.FINALIZE, member=phase_state.member, phase=phase)
+        run_task.configure(lock=f"agent-task-{task.id}").defer(agent_task_id=task.id)
         logger.info(
-            f"[agent.finalize signal] queued finalize for member {phase_state.member_id} "
+            f"[agent.finalize signal] queued finalize task {task.id} for member {phase_state.member_id} "
```
johnpooch: Remove these logs. Now that we have the record, we don't need the logs.

service/agent/signals.py
```diff
@@ -31,27 +31,21 @@
...
 
 
 @receiver(post_save, sender=PhaseState)
 @on_orders_confirmed
 @when_humans_confirmed
 def finalize(sender, instance, phase, bot_phase_states, **kwargs):
-    logger.info(
-        f"[agent.finalize signal] all humans confirmed for phase {phase.id} (game {phase.game_id})"
-    )
+    logger.info(f"[agent.finalize signal] all humans confirmed for phase {phase.id} (game {phase.game_id})")
```
johnpooch: Remove these logs. Now that we have the record, we don't need the logs.

service/agent/signals.py
```diff
@@ -65,11 +59,12 @@ def reply(sender, instance, bot_members, **kwargs):
...
-        logger.info(
-            f"[agent.reply signal] queued reply for member {member.id} (message {instance.id})"
+        task, _ = AgentTask.objects.enqueue(
+            kind=AgentTaskKind.REPLY,
+            member=member,
+            phase=instance.phase,
+            channel=instance.channel,
+            message=instance,
         )
+        run_task.configure(lock=f"agent-task-{task.id}").defer(agent_task_id=task.id)
```
johnpooch: Should we move the `run_task.configure` calls inside the `enqueue`/`create_from_event` calls?

johnpooch:
Addressed the review in 51041bf:

- **`create_from_event` + defer inside it** — renamed `enqueue` → `create_from_event` and moved the `agent.run` defer into the manager method (deferring only on create), so it mirrors `Notification`/`ChannelEvent`. Signals are now thin `create_from_event(...)` calls. This also covers the "move `run_task.configure` inside" question.
- **Dropped `channel`** — the reply channel is derived from `message.channel` now; `_dispatch` uses `task.message.channel_id` and the phase-task partial unique constraint keys on `message IS NULL`.
- **Removed the per-signal info logs** — the `AgentTask` row is the record now.

One deliberate call worth flagging: I kept `create_from_event`'s signature as explicit kwargs (`kind`, `member`, `phase`, `message`) rather than adopting `notification`'s `(event_type, context)` shape. That shape needs the `emit` context object, and wiring `agent` into `emit` is the next step we scoped out of this PR — the method will move to `(event_type, context)` when that lands. Happy to pull it forward here instead if you'd prefer.

---

PR #1134, opened 2026-08-03
Report convoy and support failures with their specific result codes

JorenC:
## What this PR does

Restores the specific failure codes for three order outcomes that were all being reported as a generic bounce or illegal move. This also un-breaks part of the full-game replay test suite, which had been failing on exactly these codes.

### The problem

When a phase resolves, every order gets a result code stored against it — `OK`, `ErrBounce`, `ErrIllegalMove` and so on. That code is what the UI shows the player to explain what happened to their order.

Internally the adjudicator is much coarser than that vocabulary. It classifies every order into one of four buckets: `OK`, `BOUNCE`, `ILLEGAL` or `CUT`. The translation layer that turns those into stored codes was a straight four-entry lookup, so **every** failed move became `ErrBounce` and **every** rejected order became `ErrIllegalMove`, regardless of why it failed.

Three failures lost real information that way:

| What happened | Was reported as | Should be |
|---|---|---|
| A convoyed army's convoy was broken, so it never sailed | `ErrBounce` | `ErrMissingConvoyPath` |
| A convoying fleet was dislodged, so its convoy order failed | `ErrBounce` | `ErrConvoyDislodged` |
| A support order named a unit that never ordered the supported action | `ErrIllegalMove` | `ErrInvalidSupporteeOrder` |

The first two matter most to players: "bounced" tells someone their army lost a fight for the destination, when actually the fleet carrying it was sunk and the army never left home. Those are quite different stories about the turn.

This was also breaking the replay test suite. `integration/test_replay.py` replays recorded real games move-by-move and asserts that today's engine reproduces the original outcomes — the strongest regression net in the repo. It is excluded from the default pytest run, and it had 4 failing fixtures, two of them purely because of these mismatched codes. A test suite that fails for known cosmetic reasons stops being useful as a regression net, because a genuine regression can't be distinguished from the existing noise.

### The fix

The engine was never actually missing the information. Alongside the coarse status it records a human-readable failure reason — "The convoy was disrupted.", "The convoying fleet was dislodged." — which pins down exactly which case occurred. Only the translation layer was throwing it away.

So the reason now narrows the code: if an order's failure reason is one of the three recognised ones, it selects the specific code; otherwise the existing status-based lookup applies unchanged.

To keep that from becoming a fragile match on prose, the three reason strings are lifted into named constants (`FailureReason`) shared by the code that produces them and the code that maps them, so the two cannot drift apart if someone rewords a message. Two result codes godip had but this vocabulary lacked — `ErrMissingConvoyPath` and `ErrConvoyDislodged` — are added, with a migration for the changed field choices.

No frontend change is needed: the API serialises the human-readable display label rather than the raw code, so the new codes surface as "Missing convoy path" and "Convoy dislodged" automatically, and they aren't enumerated in the OpenAPI schema, so no codegen churn.

## Testing

- Two new tests in `adjudication/tests.py` covering all three codes: a convoy broken by a dislodged fleet (asserting both convoy codes at once), and a support whose supportee was ordered to hold instead.
- `adjudicator`, `adjudication`, `order`, `phase`, `game` and `integration` suites pass — 1301 tests.
- Full-game replay suite (`integration/test_replay.py`, excluded from the default run): **was 4 failed / 1 passed, now 3 failed / 2 passed.** `03_hundred_solo` passes outright, and `02_classical_solo` gets past the codes and much further into the game. Every resolution-code mismatch in the suite is gone — grepping the run output for resolution mismatches returns nothing.

### What's still failing, and why it's not in scope

The 3 remaining replay failures are not resolution mismatches at all — they're the replay harness itself hitting API errors in `integration/dsl.py`, from two separate causes:

- `Cannot perform action for eliminated players.` — the harness submits orders and confirmations for every nation, including ones eliminated partway through the recorded game.
- `nation has reached maximum of 1 orders for this adjustment phase` — the harness submits more build/disband orders than the backend allows for that phase.

Both look like harness gaps rather than adjudication bugs, but each needs its own investigation and fix, so they're left out to keep this PR to one thing. Happy to take them next.

One thing deliberately left alone: bounces also record a `by` province naming the opposing unit, and that is still computed from the coarse status, so a convoy-disrupted move can carry a `by` that doesn't mean much. That behaviour is unchanged from `main`, the replay fixtures don't record `by`, and I had no ground truth for what godip did there — so I didn't guess.

## Checklist

- [ ] For PRs of any significant complexity: I ran `/review-pr` against this PR in Claude Code and addressed (or responded to) its findings — the skill could not run in this environment (it shells out to `gh pr diff`, and the session's git remote is a local proxy rather than a GitHub host). I self-reviewed the diff manually instead.
- [x] Tests cover the change
- [x] Screenshots embedded in the PR description for any visual changes (see CLAUDE.md) — n/a, no UI changes; the displayed text changes only via existing status labels

---
_Generated by [Claude Code](https://claude.ai/code/session_012T2kc7ZtUudcbCTrsrkmBo)_


johnpooch:
Closing this because I don't think this is the right fix for this technically. Making two PRs to do this properly.

PR #1137, opened 2026-08-03
Finalize bot orders at phase start when no human can order

### LINKED ISSUE #1136: Bot never confirms orders in a phase where no human has possible orders
## Goal

Make bots confirm their orders in phases where no human member has orderable units, instead of stalling until the deadline.

## Current behaviour

The `FINALIZE` AgentTask — the only path that calls `api.confirm_phase(...)` — is created solely in reaction to the `phase_state_confirmed` emit event, which is emitted only when a player toggles their own confirmation on. When every member with `has_possible_orders=True` is a bot, no human ever hits the confirm endpoint, the event never fires, and `PhaseStateConfirmedSpec` is never asked the question whose answer would be "all humans are done, finalize now".

The bot's orders are still submitted, because `phase_started` creates a `PLAN` task unconditionally, but `plan()` deliberately does not confirm. So the orders sit submitted-but-unconfirmed and the phase only moves when the deadline sweep picks it up.

Observed in game `scholars-diplomats-scoundrels-and-stabbers`, phase 68359: a retreat phase where Turkey (a bot) is the only dislodged nation. The same hole applies to adjustment phases where only bots have builds or disbands, and to games where every remaining orderable seat is a bot (kicked members get `has_possible_orders=False`, so a fully replaced table hits this too).

`agent.reconcile` does not rescue it — it re-drives stalled rows, but the `FINALIZE` row was never created.

## Expected behaviour

When a phase starts and no human member has possible orders, the bots holding the orderable seats should plan and confirm without waiting for a human signal, so the phase resolves early via the `all_confirmed` branch of `filter_due_phases`.

## Where

- `service/agent/registry.py` — `PhaseStartedSpec` and `PhaseStateConfirmedSpec`
- `service/agent/tasks.py` — `plan()` and `finalize()`
- `service/phase/serializers.py` — the sole `phase_state_confirmed` emit site
- `service/phase/models.py` — `filter_due_phases`, and `has_possible_orders` assignment on phase creation

## Approach

Make the finalize decision level-triggered rather than edge-triggered: evaluate the "are all humans done?" condition at `phase_started` as well as on `phase_state_confirmed`, so a phase that already satisfies it at creation time queues the bot's finalize work instead of waiting for an event that will never arrive. The humans-pending check itself is correct and should be shared rather than duplicated, and the interaction with the `PLAN` task queued for the same member and phase needs handling so a seat does not plan and finalize concurrently.

johnpooch:
## What this PR does

A bot never confirmed its orders in a phase where no human member had orderable units, so the phase stalled until the deadline instead of resolving early. `PhaseStartedSpec` now evaluates the same humans-pending condition that `PhaseStateConfirmedSpec` uses and queues `FINALIZE` for the orderable bot seats instead of `PLAN`. Closes #1136.

The `FINALIZE` AgentTask — the only path that calls `api.confirm_phase(...)` — was created solely in reaction to the `phase_state_confirmed` emit event, which fires only when a player confirms their own orders. When every member with `has_possible_orders=True` is a bot, no human ever hits the confirm endpoint, so the event never fires and the condition that would have queued the finalize is never evaluated. The bot's orders were still submitted by the `PLAN` task, which deliberately does not confirm, so they sat unconfirmed until the deadline sweep picked the phase up.

That covers a retreat phase only a bot nation must answer (the reported case), an adjustment phase where only bots have builds or disbands, an all-bot game, and a table whose remaining humans are all in civil disorder — civil-disorder members are created pre-confirmed, so they are not "pending" either.

Queuing finalize *instead of* plan matters rather than alongside it: `unique_phase_agent_task` is on `(kind, member, phase)`, so both rows can coexist and each defers under its own `agent-task-{id}` lock. Creating both would let one seat plan and finalize concurrently and re-submit orders after confirming. Bots without possible orders still get their `PLAN` task, unchanged.

The humans-pending and finalize-descriptor queries move to `_humans_pending` and `_finalize_tasks` so both specs share one definition instead of diverging. Behaviour when a human *is* still pending is untouched, which the existing plan/finalize trigger tests cover.

Also updates the AI player architecture notes in `CLAUDE.md`, including a stale reference to a `when_humans_confirmed` trigger that does not exist in the code.

## Checklist

- [x] Tests cover the change

On the review box: `/review-pr` shells out to the `gh` CLI, which is not available in the Claude Code on the web environment (GitHub access there is via MCP tools), so the skill could not run and I have left the box unticked rather than claim otherwise. I self-reviewed the diff manually; the points that came out of it — why finalize replaces plan rather than joining it, and that kicked bots stay excluded via `has_possible_orders=False` — are covered above and in the tests. Worth a human running `/review-pr` locally.

Backend-only change, so no screenshots apply.

New tests: a `TestPhaseStartedSpec` class covering plan-only while a human is pending, finalize when no human has possible orders, finalize when every human is already confirmed, a non-orderable bot still planning alongside a finalizing bot, and an already-confirmed bot not being asked to finalize again; plus an end-to-end `emit("phase_started", ...)` trigger test asserting the queued `FINALIZE` row and its deferred job. The four tests targeting the new behaviour fail on `main` and pass here; the regression guards pass both ways.

`agent`, `phase`, `game`, `emit`, `notification` and `bot_profile` — 864 passed, 8 skipped. Integration suite — 23 passed.

### DIFF
--- a/service/agent/registry.py
+++ b/service/agent/registry.py
@@ -30,6 +30,26 @@ def _bot_user_ids_for_phase(phase_id):
     )
 
 
+def _humans_pending(phase, bot_user_ids):
+    return (
+        phase.phase_states.filter(has_possible_orders=True, orders_confirmed=False)
+        .exclude(member__user_id__in=bot_user_ids)
+        .exists()
+    )
+
+
+def _finalize_tasks(phase, bot_user_ids):
+    bot_phase_states = phase.phase_states.filter(
+        has_possible_orders=True,
+        orders_confirmed=False,
+        member__user_id__in=bot_user_ids,
+    ).select_related("member")
+    return [
+        {"kind": AgentTaskKind.FINALIZE, "member": phase_state.member, "phase": phase}
+        for phase_state in bot_phase_states
+    ]
+
+
 class AgentTaskSpec:
     event_type = None
 
@@ -44,10 +64,20 @@ class AgentTaskSpec:
 class PhaseStartedSpec(AgentTaskSpec):
     def get_tasks(self):
         phase = self.context.phase
-        bot_members = Member.objects.filter(game_id=phase.game_id, user__bot_profile__isnull=False).select_related(
-            "user"
+        bot_members = list(
+            Member.objects.filter(game_id=phase.game_id, user__bot_profile__isnull=False).select_related("user")
         )
-        return [{"kind": AgentTaskKind.PLAN, "member": member, "phase": phase} for member in bot_members]
+        if not bot_members:
+            return []
+
+        bot_user_ids = {member.user_id for member in bot_members}
+        finalize = [] if _humans_pending(phase, bot_user_ids) else _finalize_tasks(phase, bot_user_ids)
+        finalizing = {task["member"].id for task in finalize}
+        return finalize + [
+            {"kind": AgentTaskKind.PLAN, "member": member, "phase": phase}
+            for member in bot_members
+            if member.id not in finalizing
+        ]
 
 
 @register("phase_state_confirmed")
@@ -59,24 +89,9 @@ class PhaseStateConfirmedSpec(AgentTaskSpec):
             return []
         if phase.status != PhaseStatus.ACTIVE:
             return []
-
-        humans_pending = (
-            phase.phase_states.filter(has_possible_orders=True, orders_confirmed=False)
-            .exclude(member__user_id__in=bot_user_ids)
-            .exists()
-        )
-        if humans_pending:
+        if _humans_pending(phase, bot_user_ids):
             return []
-
-        bot_phase_states = phase.phase_states.filter(
-            has_possible_orders=True,
-            orders_confirmed=False,
-            member__user_id__in=bot_user_ids,
-        ).select_related("member")
-        return [
-            {"kind": AgentTaskKind.FINALIZE, "member": phase_state.member, "phase": phase}
-            for phase_state in bot_phase_states
-        ]
+        return _finalize_tasks(phase, bot_user_ids)
 
 
 @register("channel_message")
--- a/service/agent/registry.py
+++ b/service/agent/registry.py
@@ -1,7 +1,6 @@
 from bot_profile.constants import BotKind
 from bot_profile.models import BotProfile
 from common.constants import PhaseStatus
-from member.models import Member
 
 from agent.constants import AgentTaskKind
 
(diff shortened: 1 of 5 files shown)


service/agent/registry.py
```diff
@@ -30,6 +30,26 @@ def _bot_user_ids_for_phase(phase_id):
     )
 
 
+def _humans_pending(phase, bot_user_ids):
```
johnpooch: I feel like this should be a method on the phase model. It probably doesn't even need to receive bot_user_ids because they can be derived from the phase.game.

I think it should be named: `has_unconfirmed_human`

service/agent/registry.py
```diff
@@ -44,10 +64,20 @@ def get_tasks(self):
 class PhaseStartedSpec(AgentTaskSpec):
     def get_tasks(self):
         phase = self.context.phase
-        bot_members = Member.objects.filter(game_id=phase.game_id, user__bot_profile__isnull=False).select_related(
-            "user"
+        bot_members = list(
```
johnpooch: Could this be a method on Game?

@property
def bot_members(self):
  # ...

service/agent/registry.py
```diff
@@ -44,10 +64,20 @@ def get_tasks(self):
...
+            Member.objects.filter(game_id=phase.game_id, user__bot_profile__isnull=False).select_related("user")
         )
-        return [{"kind": AgentTaskKind.PLAN, "member": member, "phase": phase} for member in bot_members]
+        if not bot_members:
+            return []
+
+        bot_user_ids = {member.user_id for member in bot_members}
+        finalize = [] if _humans_pending(phase, bot_user_ids) else _finalize_tasks(phase, bot_user_ids)
```
johnpooch: I don't like this ternary operator. Maybe just:
```
if phase.has_unconfirmed_human:
  return []
```

service/agent/registry.py
johnpooch: Done — `Phase.has_unconfirmed_human` in `phase/models.py`, no arguments.

Deriving it from the phase rather than a passed-in id set let it identify humans directly as "no bot profile" instead of "not in this set of bot user ids":

```python
@property
def has_unconfirmed_human(self):
    return self.phase_states.filter(
        has_possible_orders=True,
        orders_confirmed=False,
        member__user__bot_profile__isnull=True,
    ).exists()
```

Same predicate as the old helper, including for a member with no user at all — a null user gives a null bot profile, so it still counts as human, exactly as `exclude(member__user_id__in=...)` did.

That also let `_finalize_tasks` filter on the bot profile, so `PhaseStateConfirmedSpec` no longer needs its own "are there any bots?" guard — an empty result falls out of the query. `_bot_user_ids_for_phase` is gone entirely.


---

service/agent/registry.py
johnpooch: Done:

```python
@property
def bot_members(self):
    return self.members.filter(user__bot_profile__isnull=False).select_related("user")
```

It replaces both `Member` queries the registry was doing, so `registry.py` no longer imports `Member` at all.

The two `TestBotIdentificationByProfile` tests asserted against the deleted `_bot_user_ids_for_phase`, so they now assert against `phase.game.bot_members` — same intent (bots are identified by profile, not by the magic email, and roster bots count).


---

service/agent/registry.py
johnpooch: Ternary's gone. `PhaseStateConfirmedSpec` now reads exactly as you wrote it:

```python
if phase.status != PhaseStatus.ACTIVE:
    return []
if phase.has_unconfirmed_human:
    return []
return _finalize_tasks(phase)
```

`PhaseStartedSpec` can't literally `return []` there, though — when a human is still pending the bots must still get their `PLAN` tasks, so an empty return would stop bots planning in every ordinary phase. Same early-return shape, returning the plan list instead:

```python
plan = [{"kind": AgentTaskKind.PLAN, "member": member, "phase": phase} for member in phase.game.bot_members]
if not plan:
    return []
if phase.has_unconfirmed_human:
    return plan

finalize = _finalize_tasks(phase)
finalizing = {task["member"].id for task in finalize}
return finalize + [task for task in plan if task["member"].id not in finalizing]
```

The `if not plan` guard is load-bearing rather than defensive: without it a game with no bots pays for the `has_unconfirmed_human` query, which `TestSandboxGameCreateViewPerformance` caught as sandbox creation going 55 → 56 queries.


---

PR #1183, opened 2026-08-10
Finalise every phase when a game ends and forbid duplicate active phases

### LINKED ISSUE #1181: Finished games can be left with a stale active phase and an armed resolution job
## Goal

Stop finished games from retaining a phase that is still `active` with a live `scheduled_resolution` and a scheduled `phase.resolve_phase` job.

## Current behaviour

29 games in production have `game.status` in (`completed`, `abandoned`) while one of their phases is still `status = 'active'` with a non-null `scheduled_resolution`. Four are recent: `three` (finished 2026-08-06), `feed` (08-07), `the-free-for-all-of-the-execrable-example` (08-08, two such phases) and `the-controversy-of-the-concomitant-coat` (08-08). None are sandbox games.

Each of these stale phases is a duplicate that shares its `ordinal` with a second phase that did complete. In `the-controversy-of-the-concomitant-coat`, phase 79264 and phase 79267 are both ordinal 6 (Fall 1902, Adjustment), created three seconds apart at 17:14:00.989Z and 17:14:03.806Z. 79267 completed and the game went on to finish at 17:58:02Z; 79264 was never touched again — its `updated_at` still equals its `created_at`, it holds `scheduled_resolution = 2026-08-10T01:00:00Z`, and `resolution_job_id = 564056` is still armed against a deadline that is in the future as of today. `Phase.Meta` declares only `ordering = ["ordinal"]`, with no uniqueness on `(game, ordinal)`.

Because `Game.current_phase` is `self.phases.order_by("ordinal").last()`, the tie between the two ordinal-6 rows makes the returned phase arbitrary, so a finished game can report an active current phase with a live deadline. `Phase.objects.filter_due_phases` and `PhaseManager.send_deadline_warnings` both happen to skip these rows, but only via their `game__status` exclusion — the phase-level state itself is wrong, so anything keyed off the phase rather than the game (the deadline countdown, `formatted_deadline`) sees a running deadline on a game that is over.

## Expected behaviour

A game should never have two phases at the same ordinal, and when a game reaches `completed` or `abandoned` every one of its phases should be `completed` with `scheduled_resolution` cleared and any armed resolution job cancelled, so `current_phase` is unambiguous and no finished game presents a live deadline.

## Where

`service/phase/models.py` — `PhaseManager.resolve_if_due`, `create_from_adjudication_data` and the game-ending branches of `resolve`; the `Phase.Meta` class, which has no `(game, ordinal)` uniqueness; `service/phase/signals.py` `arm_deadline_resolution`; and `Game.current_phase` in `service/game/models.py`.

johnpooch:
## What this PR does

Fixes #1181. A finished game could keep a phase that was still `active` with a live `scheduled_resolution` and an armed `phase.resolve_phase` job, because the game-ending branches only finalised the phase they had in hand. Ending a game now finalises *every* phase of that game, and a unique constraint stops a second live phase from being created at an ordinal a game already has.

### Changes

- **`Game.finish(status)`** (`service/game/models.py`) sets the status and `finished_at`, then completes every phase that is not already completed, clearing `scheduled_resolution` — which makes the existing `arm_deadline_resolution` post_save signal cancel the armed resolution job. Both game-ending branches of `PhaseManager.resolve` (victory, abandonment) and `DrawProposal.process_acceptance` now go through it, replacing three copies of the same six lines.
- **Unique constraint** `unique_live_phase_ordinal_per_game` on `(game, ordinal)` for every phase that is not `completed`. In both production incidents the duplicate was inserted while its twin was still live, so this is the insert that gets rejected.
- **Deterministic current phase**: `Phase.Meta.ordering` is now `["ordinal", "id"]`, `Game.current_phase` orders by `("ordinal", "id")`, and the two `DISTINCT ON (game_id)` subqueries that pick the current/latest-completed phase per game break their tie on `-id`. Legacy duplicates no longer make `current_phase` arbitrary.
- **Data migration** `0018_complete_phases_of_finished_games` completes the phases of games that are already `completed`/`abandoned` and clears their deadlines, so the constraint in `0019` applies cleanly.

### Why the constraint is scoped to live phases

The issue asks for uniqueness on `(game, ordinal)` outright. I checked production before writing the migration: five duplicate groups match the pattern in the issue (one live stale row, one completed row the game actually followed), but a sixth game — `the-battle-of-the-execrable-experience`, from October 2025 — carries **27 duplicate pairs where both rows are `completed`**, two interleaved chains of real phases with units, supply centres and phase states on both sides. Which chain the game followed cannot be determined automatically (the low-id chain holds the final phase the game ended on; the high-id chain holds the fuller board), so an unconditional constraint would need a migration that deletes 27 phases of a real 7-player game on a guess. That is a call for a human, not for a data migration.

Excluding completed rows keeps the invariant that actually bites — a game cannot gain a second live phase at an ordinal it already has — with no destructive migration. I confirmed no `(game, ordinal)` group in production currently holds two non-completed rows, so `0019` applies cleanly. If you want full `(game, ordinal)` uniqueness, say which chain of that 2025 game to keep and I will follow up with the dedupe.

### Two things I deliberately did not do

- **The five armed jobs in production are left to expire** rather than cancelled by the migration. Cancelling from a migration means reaching into `procrastinate_jobs` and depending on a third-party app's migration graph; the jobs are already no-ops, since `resolve_if_due` re-checks `filter_due_phases`, which now sees a completed phase in a finished game. Going forward the signal cancels them properly.
- **I did not find the root-cause race.** `resolve_if_due` takes `SELECT ... FOR UPDATE` on the phase and re-checks `filter_due_phases` inside the transaction, and every deferral shares a `resolve-game-<id>` procrastinate lock, so the double resolution should not be reachable from the code on `main`. The constraint turns whatever the path is into a loud `IntegrityError` that rolls the second resolution back (and procrastinate's retry will then skip it), rather than silent corruption.

### Rebased on #1176

`main` picked up the `processing` phase status while this was open, which conflicted in `PhaseManager.resolve`. Merge resolved in favour of `Game.finish(...)` inside `_resolve_claimed`'s try block, and the constraint condition widened from "active" to "not completed" so the new `processing` status cannot host a second live phase either. `claim_for_processing`'s compare-and-swap and this constraint are complementary: that one stops a second resolution starting, this one stops a second phase row landing whatever the path.

## Checklist

- [x] Tests cover the change

### Tests

`service/phase/tests.py`
- a second `active`/`processing`/`pending` phase at an ordinal the game already has is rejected; a completed one is allowed (documents the scope) and `current_phase` resolves deterministically
- resolving into a victory, and into abandonment, leaves every phase of the game completed with no `scheduled_resolution`, including a leftover live phase at another ordinal
- `Game.finish` cancels the armed resolution job and clears `resolution_job_id`

`service/draw_proposal/tests.py`
- accepting a draw completes a leftover live phase, not just the proposal's phase

`service/conftest.py` — the `retreat_phase` fixture hardcoded `ordinal=1` on games that already had a live phase at ordinal 1, which is the state the constraint forbids; it now appends at the next ordinal.

Full backend suite: 2126 passed, 8 skipped.

### DIFF
--- a/service/conftest.py
+++ b/service/conftest.py
@@ -1250,7 +1250,7 @@ def retreat_phase(db):
             year=1901,
             type=PhaseType.RETREAT,
             status=PhaseStatus.ACTIVE,
-            ordinal=1,
+            ordinal=game.phases.count() + 1,
         )
         primary_member = game.members.first()
         phase.phase_states.create(member=primary_member)
--- a/service/draw_proposal/models.py
+++ b/service/draw_proposal/models.py
@@ -1,7 +1,6 @@
 from django.db import models, transaction
-from django.utils import timezone
 from common.models import BaseModel
-from common.constants import GameStatus, PhaseStatus
+from common.constants import GameStatus
 from draw_proposal.constants import DrawProposalStatus
 from emit import emit
 from victory.models import Victory
@@ -142,14 +141,7 @@ class DrawProposal(BaseModel):
                 member.drew = True
                 member.save()
 
-            self.game.status = GameStatus.COMPLETED
-            self.game.finished_at = timezone.now()
-            self.game.save()
-
-            self.phase.status = PhaseStatus.COMPLETED
-            self.phase.scheduled_resolution = None
-            self.phase.save()
-
+            self.game.finish(GameStatus.COMPLETED)
             self.game.emit_game_ended()
 
             return victory
--- a/service/draw_proposal/tests.py
+++ b/service/draw_proposal/tests.py
@@ -245,6 +245,30 @@ class TestDrawProposalProcessAcceptance:
         assert phase.status == PhaseStatus.COMPLETED
         assert phase.scheduled_resolution is None
 
+    def test_completes_every_phase_of_the_game(
+        self, game_factory, phase_factory, member_factory, draw_proposal_factory
+    ):
+        game = game_factory(variant__solo_victory_sc_count=18)
+        future = timezone.now() + timezone.timedelta(hours=20)
+        stale_phase = phase_factory(game=game, scheduled_resolution=future)
+        phase = phase_factory(game=game, ordinal=stale_phase.ordinal + 1, scheduled_resolution=future)
+        m1 = member_factory(game=game)
+        m2 = member_factory(game=game)
+
+        proposal = draw_proposal_factory(
+            game=game, created_by=m1, phase=phase,
+            included_member_ids=[m1.id, m2.id],
+        )
+        for vote in proposal.votes.all():
+            vote.accepted = True
+            vote.save()
+
+        assert proposal.process_acceptance() is not None
+
+        stale_phase.refresh_from_db()
+        assert stale_phase.status == PhaseStatus.COMPLETED
+        assert stale_phase.scheduled_resolution is None
+
     def test_cd_member_not_in_victory(
         self, game_factory, phase_factory, member_factory
     ):
--- a/service/game/models.py
+++ b/service/game/models.py
@@ -90,13 +90,13 @@ class GameQuerySet(models.QuerySet):
         )
 
         current_phase_ids = (
-            Phase.objects.order_by("game_id", "-ordinal")
+            Phase.objects.order_by("game_id", "-ordinal", "-id")
             .distinct("game_id")
             .values("id")
         )
         latest_completed_phase_ids = (
             Phase.objects.filter(status=PhaseStatus.COMPLETED)
-            .order_by("game_id", "-ordinal")
+            .order_by("game_id", "-ordinal", "-id")
             .distinct("game_id")
             .values("id")
         )
@@ -155,13 +155,13 @@ class GameQuerySet(models.QuerySet):
         )
 
         current_phase_ids = (
-            Phase.objects.order_by("game_id", "-ordinal")
+            Phase.objects.order_by("game_id", "-ordinal", "-id")
             .distinct("game_id")
             .values("id")
         )
         latest_completed_phase_ids = (
             Phase.objects.filter(status=PhaseStatus.COMPLETED)
-            .order_by("game_id", "-ordinal")
+            .order_by("game_id", "-ordinal", "-id")
             .distinct("game_id")
             .values("id")
         )
@@ -500,7 +500,7 @@ class Game(BaseModel):
                 phases = list(self.phases.all())
                 return phases[-1] if phases else None
 
-            return self.phases.order_by("ordinal").last()
+            return self.phases.order_by("ordinal", "id").last()
 
     @property
     def movement_phase_duration_seconds(self):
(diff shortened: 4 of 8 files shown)


johnpooch:
CI status on `48ccb0a`: everything owned by this repo is green — `lint` ×2, `build` ×3, `test` ×3, `check-pr-limit`, and the `diplicity-react` deploy preview all pass.

The only red is the **`diplicity-design-playground` Netlify site** (`Pages changed` / `Header rules` / `Redirect rules`, all "Deploy failed"). I'm not fixing it, because it isn't this PR's failure and I don't think it's fixable from a diff:

- There is no `design-playground` package in the repo. `packages/` holds `map`, `map-parse`, `maps`, `variant-creator`, `web`, and the only `netlify.toml` files are `packages/web/netlify.toml` and `packages/variant-creator/netlify.toml` — nothing configures a design-playground build.
- No PR before today shows that check at all (#1178 from 08-08 and #1153 from 08-04 have Netlify checks for `diplicity-react` and `diplicity-variant-creator` only), so the site was connected to this repo very recently and its first deploys are failing on setup.
- This PR is backend-only: `service/**` Python plus two migrations. It touches nothing any frontend build reads.

It looks like the Netlify site needs a base directory / build command pointing at real source, or needs unlinking from the repo. Happy to pick it up if you'd rather it were handled here, but it needs a decision on what that site is meant to build.

---

johnpooch:
Two updates, both superseding my earlier comment.

**The design-playground failure is gone — ignore that comment above.** Merging `main` pulled in the workflow that skips design playground builds when nothing under it changes, so those three checks now come back `neutral` (canceled) rather than failed. Nothing needs configuring or unlinking. CI on `bc8cdfe` is green across the board: `lint` ×2, `build` ×3, `test` ×4, `approve`, `deploy-staging`.

**The migrations were verified against a copy of production data.** The staging environment for this PR runs the 03:50 GMT snapshot with `0018` and `0019` applied, so `AddConstraint` — the only operation that could fail on real data — succeeded against the full production dataset. Checking the five stale phases from #1181 through the public phase endpoint on staging:

| phase | game | before | after |
|---|---|---|---|
| 79264 | the-controversy-of-the-concomitant-coat | `active`, deadline 2026-08-10T01:00Z, job 564056 | `completed`, no deadline |
| 71577 | three | `active`, deadline 2026-08-07T16:00Z | `completed`, no deadline |
| 73607 | feed | `active`, deadline 2026-08-07T16:00Z | `completed`, no deadline |
| 78499 | the-free-for-all-of-the-execrable-example | `active`, deadline 2026-08-12T01:00Z | `completed`, no deadline |
| 78570 | the-free-for-all-of-the-execrable-example | `active`, deadline 2026-08-10T01:00Z | `completed`, no deadline |

`remaining_time` is null on all five. Phase 79267 — the ordinal-6 twin the game actually followed — still carries its original `scheduled_resolution` of `2026-09-21T01:00:00Z`, which is the `status <> 'completed'` guard working as intended: the migration repairs the broken rows without rewriting the deadline history of every phase in every finished game.

One note for the production run: `0019` will abort if a duplicate live phase appears between now and then, since the race stays live until this deploys. Worth re-running immediately before shipping:

```sql
SELECT game_id, ordinal, COUNT(*)
FROM phase_phase
WHERE status <> 'completed' AND game_id IS NOT NULL
GROUP BY game_id, ordinal HAVING COUNT(*) > 1;
```

It returned zero rows against live production as of this comment.

---

PR #1198, opened 2026-08-10
Preserve commitment rating across account deletion and re-registration

### LINKED ISSUE #1186: Preserve commitment rating across account deletion and re-registration
## Goal

Persist a player's commitment rating when they delete their account, and restore it if they later register again with the same email.

## Context

`commitment` lives on `UserProfile`, which is cascade-deleted along with the `User`. A player whose rating has fallen to `LOW` — and who is therefore rejected from every game by `commitment_allows_requirement` — can delete their account and re-register with the same email to return to the default `UNDEFINED`, which passes every requirement except `COMMITTED`. Account deletion is currently a reset button on the only reliability mechanic we have.

## Approach

Record the rating against a hash of the email at deletion time in `UserAccountDeleteView.perform_destroy`, and rehydrate it wherever the `UserProfile` is created during registration — the email, Google and Apple paths all reach it through `get_or_create`.

johnpooch:
## What this PR does

`commitment` lives on `UserProfile`, which is cascade-deleted with the `User`, so a player rated `LOW` could delete their account, re-register with the same email and come back as `UNDEFINED`. Deletion was a reset button on the only reliability mechanic we have. Closes #1186.

`UserAccountDeleteView.perform_destroy` now records the rating in a new `RetainedCommitment` row keyed by a hash of the email (inside the existing transaction, before the user is deleted), and the three registration paths seed the new profile's `commitment` from that record:

- `RegisterSerializer` (email) — `UserProfile.objects.create(..., commitment=...)`
- `AuthSerializer` (Google) and `AppleAuthSerializer` — added to the `get_or_create` defaults, so an existing profile is untouched

Details worth knowing:

- The email is never stored. The key is a `salted_hmac` digest rather than a bare SHA-256, so the stored value can't be reversed from a list of candidate addresses by anyone holding the database alone.
- The address is stripped and lowercased before hashing, so `Player@Example.com` and `player@example.com` resolve to the same record. That also means switching provider on the way back in — deleting an email account and re-registering with Google on the same address — still restores the rating.
- A blank email is not retained, so users without an address can't collide on a shared empty-string hash. Retaining again for the same address overwrites the previous value.
- Deleting an account that has no `UserProfile` (a `createsuperuser` account) is guarded rather than dereferencing `instance.profile`, which would 500.
- The model is registered in admin so a rating inherited via a recycled address can be cleared.

A restored `LOW` rating is not recomputed away in practice: the only games a `LOW` player can join are private ones, and `_recompute_commitment` returns early for private and sandbox games (`phase/models.py:497`). The one thing that would clear a restored rating is `manage.py backfill_commitment`, which recomputes every profile from its own phase history — worth knowing before running it, but out of scope here.

## Checklist

- [x] For PRs of any significant complexity: I ran a review against this PR in Claude Code and addressed (or responded to) its findings
- [x] Tests cover the change
- [x] Screenshots embedded in the PR description for any visual changes (see CLAUDE.md) — backend only, nothing visual changes

Tests: retention on delete, deletion of a profile-less account, and the manager's matching rules in `user_profile/tests.py`; restoration on each of the three registration paths in `login/tests_register.py`, `login/tests.py` and `login/tests_apple_auth.py`, plus an end-to-end delete → re-register test. The restoration tests fail without the serializer changes, and the profile-less deletion test fails without the guard (both verified by reverting). Full backend suite on the merged tree: 2149 passed, 8 skipped.

Includes a merge of `main` to resolve a conflict in `perform_destroy` with `626ff17` ("Stop expecting orders from a deleted account on the live phase"); both sides are kept.


johnpooch:
Closing. I've decided that this actually isn't a real problem, especially when bots and forgiveness are in place.

Issue #1163, opened 2026-08-06
Creating a channel 500s when the user holds more than one member in the game


johnpooch:
## Goal

Stop `POST /games/{game_id}/channels/create/` raising `Member.MultipleObjectsReturned` when the requesting user has more than one `Member` row in the game.

## Current behaviour

`ChannelManager.create_from_member_ids` resolves the caller's own member with a bare `get()` (`service/channel/models.py:59`):

```python
member_ids = member_ids + [game.members.get(user=user).id]
```

Nothing guarantees one member per (game, user). There is no unique constraint on `Member` — `service/member/models.py:33` declares indexes on `game`/`user` but no `constraints`. And the sandbox path deliberately creates one member per nation for the same user (`service/game/models.py:328`, and the same pattern in the clone-to-sandbox serializer at `service/game/serializers.py:643`):

```python
members_to_create = [Member(game=game, user=user) for _ in nations_list]
```

Sentry `DIPLICITY-API-9K`, 13 events since 2026-08-03, escalating, most recent 2026-08-06:

```
Member.MultipleObjectsReturned: get() returned more than one Member -- it returned 3!
  channel/models.py, line 59, in create_from_member_ids
  game: <Game: Game object (the-fisticuffs-of-the-insidious-scissors)>
  user: <User: animewave>
```

The user gets a 500 and cannot open a private channel at all.

## Expected behaviour

Channel creation resolves the caller's member unambiguously and succeeds, whatever number of member rows the user holds in that game.

## Where

- `service/channel/models.py` — `ChannelManager.create_from_member_ids` (58)
- `service/channel/serializers.py` — `ChannelSerializer.create` (76)
- `service/member/models.py` — `Member.Meta` (33), no uniqueness on (game, user)
- `service/game/models.py` — `create_sandbox` (319)

## Approach

Decide what "the caller's member" means when several exist and encode it — for a normal game that is the seat that has not been kicked or replaced, matching the `IsNotKickedGameMember` rule the chat endpoints already use. Worth checking at the same time whether the failing game is genuinely multi-seat or whether members are being duplicated by some other path, since a non-sandbox game with three seats for one user may be a second defect underneath this one.

johnpooch:
Scoping note, so the "escalating" substatus in the description is not read as breadth: aggregating the events by URL (14d) shows all **13 events are the same game**, `the-fisticuffs-of-the-insidious-scissors`, and the rate is ~2/day.

[...]

PR #1202, opened 2026-08-10
Forbid a user holding two members in one game

johnpooch:
## What this PR does

Adds a database unique constraint on `(game, user)` for `Member`, so a user can never hold two seats in the same game. Relates to #1163.

The join view's "is this user already a member?" test lived in a permission class, which reads the game before anything is written — two overlapping joins both passed and both took a seat. b1b6475 closed that race in the application with `SeatClaimMixin`, but nothing in the database stopped a duplicate. Two production games had already been left with one user holding several nations, which is what broke `POST /games/{id}/channels/create/`: `ChannelManager.create_from_member_ids` resolves the caller with a bare `get()` and raised `Member.MultipleObjectsReturned` (Sentry `DIPLICITY-API-9K`). Both games have since been deleted, so production currently holds zero non-sandbox duplicates.

**Why the constraint is conditional.** Sandbox games deliberately seat one user in every nation (`Game.objects.create_sandbox`, and the same pattern in `GameCloneToSandboxSerializer`), so an unconditional constraint is not available. Postgres cannot reference `game.sandbox` from a partial index predicate, so `Member` carries its own `sandbox` flag and the constraint keys off it:

```python
models.UniqueConstraint(
    fields=["game", "user"],
    condition=models.Q(sandbox=False),
    name="unique_member_per_user_per_game",
)
```

`Game.sandbox` is fixed at creation and never mutated, so the flag cannot drift, and the two sandbox seeding paths are the only writers. A future path that forgets to set it fails loudly on its second member rather than silently duplicating.

`user` stays nullable — account deletion nulls it — and Postgres treats NULLs as distinct, so a game with several deleted-account members is still legal.

**Migrations.** `0008` adds the field, `0009` backfills `sandbox=True` for members of sandbox games (378 game/user pairs in production), `0010` adds the constraint. `0009` must run before `0010` or the constraint rejects every existing sandbox game. Verified end to end: the staging deploy applied all three against a production snapshot and the backend served (`/variants/` → 200).

**Test fixtures.** Several fixtures and tests seated one user in many nations of a *non-sandbox* game as a shortcut. Those now give each seat its own user, which is what the tests meant; `phase_factory` keeps `primary_user` for the first seat so tests that authenticate as them are unaffected. Sandbox fixtures pass `sandbox=True`.

No serializer exposes the new field (`BaseMemberSerializer` is an explicit `Serializer`), so the OpenAPI schemas are unchanged.

**Open question on test style.** `.claude/rules/backend/tests.md` gained "do not test models, managers, or querysets directly — assert behaviour through HTTP endpoints" after this branch was written. Four of the five tests in `TestUniqueMemberPerUserPerGame` create `Member` rows directly and assert `IntegrityError`, because the constraint guards a race the HTTP layer rejects first — `SeatClaimMixin` returns 403 before the insert, so an endpoint cannot reach the constraint. The HTTP-observable behaviour is already covered by the existing join-race and sandbox-creation tests. Happy to drop the direct-model tests if the rule should win here.

## Checklist

- [x] Tests cover the change — `TestUniqueMemberPerUserPerGame` covers the rejected duplicate, the same user across two games, several members without a user, a full sandbox seating, and that a join creates a non-sandbox member. Full backend suite on a clean checkout: 2171 passed, 8 skipped.

### DIFF
--- a/service/adjudicator/test_service.py
+++ b/service/adjudicator/test_service.py
@@ -671,20 +671,18 @@ class TestAdjudicationService:
         classical_austria_nation,
         classical_russia_nation,
         classical_turkey_nation,
-        primary_user,
-        secondary_user,
-        tertiary_user,
+        user_factory,
     ):
         game = Game.objects.create(variant=classical_variant, name="Classical Test Game", status=GameStatus.ACTIVE)
 
         members_by_nation = {
-            "England": Member.objects.create(nation=classical_england_nation, user=primary_user, game=game),
-            "France": Member.objects.create(nation=classical_france_nation, user=secondary_user, game=game),
-            "Germany": Member.objects.create(nation=classical_germany_nation, user=tertiary_user, game=game),
-            "Italy": Member.objects.create(nation=classical_italy_nation, user=primary_user, game=game),
-            "Austria": Member.objects.create(nation=classical_austria_nation, user=secondary_user, game=game),
-            "Russia": Member.objects.create(nation=classical_russia_nation, user=tertiary_user, game=game),
-            "Turkey": Member.objects.create(nation=classical_turkey_nation, user=primary_user, game=game),
+            "England": Member.objects.create(nation=classical_england_nation, user=user_factory(), game=game),
+            "France": Member.objects.create(nation=classical_france_nation, user=user_factory(), game=game),
+            "Germany": Member.objects.create(nation=classical_germany_nation, user=user_factory(), game=game),
+            "Italy": Member.objects.create(nation=classical_italy_nation, user=user_factory(), game=game),
+            "Austria": Member.objects.create(nation=classical_austria_nation, user=user_factory(), game=game),
+            "Russia": Member.objects.create(nation=classical_russia_nation, user=user_factory(), game=game),
+            "Turkey": Member.objects.create(nation=classical_turkey_nation, user=user_factory(), game=game),
         }
 
         phase = Phase.objects.create(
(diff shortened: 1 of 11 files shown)


johnpooch:
CI is red on `test`, but the failure is on the base branch, not this PR:

```
FAILED game/tests.py::TestGameCloneToSandbox::test_clone_to_sandbox_twice_creates_two_games
  ModuleNotFoundError: No module named 'adjudication'
1 failed, 2137 passed, 8 skipped
```

That test arrived in `06ecdd38` (#1200) and patches `"adjudication.service.start"` — the module path from before `3d5c757` renamed the app to `adjudicator`. Other tests in the same file patch the imported `adjudication_service` object, so they survived the rename; this one uses a string target, so it didn't.

I reproduced it on a clean checkout of `main` at `a2e0c78c` with this branch nowhere involved — it fails there too. Every open PR will pick it up on merge until it's fixed. The fix is the one-line target change to `"adjudicator.service.start"`, which doesn't belong in this PR.

Everything else on `6014985e` is green, including `deploy-staging`: the staging backend is serving (`/variants/` → 200), so `0009`/`0010` applied cleanly against the Aug 12 production snapshot — which was the migration risk worth checking.

Separately, and not blocking here: the `deploy-staging` check reported success on the previous run even though the container crashed on `migrate` seconds later. Its readiness poll read `SUCCESS` 15s after seeing `BUILDING`, which was the prior deployment's status. A green `deploy-staging` isn't currently evidence that migrations applied.

---

PR #1210, opened 2026-08-12
Refactor channel and draw_proposal apps

johnpooch:
## What this PR does

Executes the channel / draw_proposal refactor plan end to end: dead queryset helpers deleted, draw-proposal cancel and channel mark-read converted to updates, unread counts moved off the game serializers onto two dedicated single-query endpoints, the channel list/retrieve split with a latest-message preview and cursor-paginated messages, and both apps' test suites consolidated to one class per view.

**Backend**
- **A** — `for_game` / `pending_for_phase` / `active` deleted from draw_proposal (callers inlined to `.filter(cancelled=False)`), channel's zero-caller helpers deleted, the two `DrawVote(...)` constructions merged, the duplicated sandbox 400 removed (the `IsNotSandboxGame` 403 stands alone), and the vote lookup game-scoped (cross-game proposal ids now 404, with a test).
- **B** — `DrawProposalVoteView` → `DrawProposalVoteUpdateView`; cancel is now `PATCH …/cancel/` via an `UpdateAPIView` + `DrawProposalCancelSerializer.update()`, returning 200 with the serialized proposal (`to_representation` delegates to `DrawProposalSerializer`; `@extend_schema_view` keeps the response schema honest).
- **C** — `GameQuerySet.with_total_unread_counts()` replaced by the `ChannelMember`-rooted GROUP BY (`ChannelMemberQuerySet.unread_counts_by_game`). `GET /game/<id>/channel/unread/` and `GET /game/unread/` are each asserted to run exactly **one query**; the batched endpoint filters zero rows in the database. `total_unread_message_count` is gone from both game serializers. `channel.urls` is now included before `game.urls` so `/game/unread/` isn't shadowed by `/game/<str:game_id>/`.
- **D** — `ChannelPreviewSerializer` (id, name, private, latest message) with `with_preview_data()`: a `Subquery` annotation for ordering (the `Max()` in `order_for_list()` is gone) plus a **sliced `Prefetch`** for the preview row. The sliced-Prefetch variant was chosen over body/sender-id annotations because `ChannelListScreen` renders the full sender object (`isCurrentUser` for the "You:" label, nation name, sender-name fallback), so `ChannelMessageSerializer` is reused verbatim. New `ChannelRetrieveView` serves cursor-paginated messages (`page_size` 20, `ordering = ("-created_at", "-id")`, supporting `(channel, created_at)` index). `with_sender_data()` now covers `sender__user__profile`, `sender__nation`, `sender__nation__flag`, and the retrieve query count is pinned constant in a test. `with_related_data()` is deleted.
- **E** — mark-read is an `UpdateAPIView`; the `ChannelMember` lookup lives in `get_object()`, the serializer keeps only `update()`, and the response is DRF's default 200 (fixing the schema's phantom 201).
- **F** — `service/channel/tests/` collapsed into one `tests.py`, one class per view; the model/manager/queryset test classes are folded into endpoint assertions; `TestChannelMessageBotChannel` deleted; the one surviving `ChannelEvent` test moved to `phase/tests.py`; draw_proposal tests now use `reverse()` and the shared client fixtures (`_auth` deleted).

**Frontend**
- In-game nav badge polls `GET /game/<id>/channel/unread/` (5 s, active games, members only) instead of polling the whole game retrieve.
- `GameCard` renders an `UnreadBadge` inside `<Suspense fallback={null}>` calling `useGameUnreadListSuspense` with a 30 s `refetchInterval` — every card shares one request via query-key dedup. `FindGames` compiles and passes untouched (non-member games are absent from the response → 0).
- `ChannelListScreen` renders the served `latestMessage`; `ChannelScreen` consumes the retrieve endpoint through a `useSuspenseInfiniteQuery` over the cursor, with a "Load older messages" affordance.
- MSW fixtures/handlers reworked: fixture channels keep full message arrays, handlers derive the preview list, retrieve envelope, and both unread endpoints; mark-read and cancel handlers switched to PATCH.

**Decisions & deviations from the plan (called out deliberately)**
- `ChannelQuerySet.with_retrieve_data()` was **not** created: with messages cursor-paginated per request through `ChannelMessageQuerySet.with_sender_data()`, a channel-level retrieve queryset would encapsulate nothing — a `return self` pass-through is exactly the rot Part A deletes. If a retrieve-specific prefetch appears later, it can earn the method then.
- `with_sender_data()` omits the plan's `sender__user__bot_profile`: no such relation exists (bot identity lives on `UserProfile.kind`), and `select_related` on it errors.
- `test_create_for_channels_empty_is_noop` is **deleted without replacement** — it covered a defensive branch not reachable through any endpoint. This is a deliberate loss of coverage, not an accident.
- `TestDrawProposalSecretVoting`'s three list-shaped tests folded into `TestDrawProposalListView` as planned; its fourth test exercises the **vote** endpoint, so it folded into `TestDrawProposalVoteUpdateView` instead (one class per view).
- The "New messages" divider and per-channel unread badges are removed on the frontend: per-channel unread left the API (Part D3, and reinstating it is explicitly out of scope), so they have no data source.
- `IsChannelMember` now returns `False` for anonymous users on private channels instead of crashing on an `AnonymousUser` filter — needed because the new retrieve endpoint is `AllowAny` (public channels stay anonymously viewable, matching the list).
- The agent harness consumed the channel list's nested messages; `agent/api_client.get_channels` now follows the list with per-channel retrieve calls (in-process, bot games have few channels) and feeds the newest page ascending. `harness/types.py` defines the merged shape locally.
- `game/tests/test_games_list_unread.py` is deleted wholesale — it tested only the retired list field; the behaviours are re-asserted against the unread endpoints.

**Codegen** — `docker compose` is unusable here (no docker daemon; the known `postgres_data` 13-vs-17 issue also stands), so the three generators were run natively with the config `.claude/rules/backend/codegen.md` requires: `DJANGO_DEBUG` off and `FIREBASE_PROJECT_ID` set (with a throwaway service-account key so firebase init succeeds and `/devices/` stays in the schema), then `manage.py spectacular` (both variants), `datamodel-codegen` with the compose file's exact flags, and `npx orval`. No generated file was hand-edited. `openapi-schema.internal.yaml` is untracked on `main` and stays untracked.

**Verification**
- Backend: `pytest -n auto --create-db` → **2141 passed, 8 skipped** (the pre-existing schema-generation warnings/`UserAccountDeleteView` error predate this branch).
- Frontend: `npx tsc -b --noEmit`, `npm run lint`, `npm run test` → **562 passed** across 61 files.
- Screenshots were taken against `npm run dev:mocks` (channel list with previews, channel screen on the retrieve endpoint, My Games with the "1 new" badge from the batched endpoint). This environment has no way to upload GitHub attachments, so they are delivered in the Claude session alongside this PR — please drag them into this description. The `/tmp/shots/` files are not committed.

## Checklist

- [x] Tests cover the change
- [ ] Screenshots embedded in the PR description for any visual changes (see CLAUDE.md) — taken, but this environment cannot upload attachments; delivered via the session for drag-in

https://claude.ai/code/session_01NJKpaP6FBAZs1xUcNmta7z

---
_Generated by [Claude Code](https://claude.ai/code/session_01NJKpaP6FBAZs1xUcNmta7z)_


johnpooch:
Merged `main` into this branch to clear the conflict (`8ca39f3`). The only conflicts were the four create-game query-count assertions in `service/game/tests.py` — this branch drops one query, `main`'s game-id retry adds two, so they resolve to 46/46/56/56. Verified by running them, plus the frontend typecheck/lint/test suite.

The remaining red check is **not from this branch**: `test` fails on

```
FAILED game/tests.py::TestGameCloneToSandbox::test_clone_to_sandbox_twice_creates_two_games - ModuleNotFoundError: No module named 'adjudication'
```

`test_clone_to_sandbox_twice_creates_two_games` patches `"adjudication.service.start"`, but the module is `adjudicator.service` — every other test in the file patches that. The test arrived with a2e0c78 on `main`, and [`main`'s own test-service run for that commit is red for the same reason](https://github.com/johnpooch/diplicity-react/actions/runs/31651598454/job/94296916217). Nothing else fails: 2121 passed, 8 skipped.

Leaving the one-word fix out of this branch so the PR stays scoped to the refactor — happy to send it as its own PR against `main` if you'd prefer that over fixing it in the commit that introduced it.

---

packages/web/src/screens/GameDetail/ChannelListScreen.tsx
```diff
@@ -103,14 +102,9 @@ const ChannelListScreen: React.FC = () => {
                             {!channel.private && (
                               <Badge variant="outline">Public</Badge>
                             )}
-                            {channel.unreadMessageCount > 0 && (
-                              <Badge variant="default">
-                                {channel.unreadMessageCount}
-                              </Badge>
-                            )}
```
johnpooch, line 110: We removed the unread message count here. Why?

service/channel/views.py
```diff
@@ -1,43 +1,105 @@
...
+
+    permission_classes = [permissions.AllowAny, IsChannelMember]
+    serializer_class = ChannelRetrieveSerializer
+
+    def get_object(self):
+        return self.get_channel()
+
+
+class ChannelUnreadRetrieveView(generics.RetrieveAPIView):
+    """Retrieve the total number of unread channel messages for the current user in a game."""
```
johnpooch: Why are these docstrings here? The guidance clearly asks not to include docstrings. Does the guidance need updating?

service/channel/views.py
```diff
@@ -1,43 +1,105 @@
...
+        row = (
+            ChannelMember.objects.unread_counts_by_game(self.request.user)
+            .filter(channel__game_id=self.kwargs.get("game_id"))
+            .first()
+        )
+        return row or {"total_unread_message_count": 0}
+
+
+class GameUnreadListView(generics.ListAPIView):
+    """List the current user's games that have unread channel messages, with their unread counts."""
```
johnpooch: Why are these docstrings here? The guidance clearly asks not to include docstrings. Does the guidance need updating?

service/channel/views.py
```diff
@@ -1,43 +1,105 @@
...
         )
+
+
+@extend_schema_view(
+    get=extend_schema(
+        parameters=[OpenApiParameter("cursor", str, required=False)],
+    ),
+)
+class ChannelRetrieveView(SelectedGameMixin, SelectedChannelMixin, generics.RetrieveAPIView):
+    """Retrieve a channel with a cursor-paginated page of its messages."""
```
johnpooch: Why are these docstrings here? The guidance clearly asks not to include docstrings. Does the guidance need updating?

service/channel/views.py
```diff
@@ -1,43 +1,105 @@
...
+    def get_object(self):
+        return get_object_or_404(
+            ChannelMember,
+            member=self.get_current_game_member(),
+            channel=self.get_channel(),
+        )
 
 
 class ChannelListView(SelectedGameMixin, generics.ListAPIView):
+    """List the channels of a game visible to the current user, with a preview of the latest message."""
```
johnpooch: Why are these docstrings here? The guidance clearly asks not to include docstrings. Does the guidance need updating?

johnpooch: Answering here for all four threads (lines 48, 69, 79, 94) — they're the same question.

These are the documented exception, so I don't think the guidance needs updating. Both copies carve out DRF view docstrings specifically:

- `CLAUDE.md:85` — "Do not add comments or docstrings, including in tests. … **The only exception is DRF view docstrings, which are extracted for OpenAPI.**"
- `.claude/rules/backend/index.md:8` — "**The one exception is DRF view docstrings**, which are extracted into the OpenAPI schema — write those."

The reason is load-bearing rather than decorative: drf-spectacular lifts the docstring into the operation `description`, so deleting them silently empties the generated client docs. You can see the current ones in the committed schema:

- `service/openapi-schema.yaml:443` — `description: Retrieve the total number of unread channel messages for the current…` (`ChannelUnreadRetrieveView`)
- `service/openapi-schema.yaml:1019` — `description: List the current user's games that have unread channel messages,…` (`GameUnreadListView`)
- `service/openapi-schema.yaml:1147` — `description: Retrieve a channel with a cursor-paginated page of its messages.` (`ChannelRetrieveView`)

So removing them would be a visible regression in the schema diff. Happy to drop them if you'd rather the schema carry no descriptions, but that's a guidance change rather than a fix to this PR.

---

packages/web/src/screens/GameDetail/ChannelListScreen.tsx
johnpooch: It was removed because per-channel unread left the API in Part D — `ChannelPreviewSerializer` had no count to render, and the two new endpoints aggregated per *game*, not per channel. That was a scope call from the plan, and it was the wrong one: it's a UX regression and the data was one `values()` grouping away. Restored in `46b0ce5`.

The breakdown lives on `ChannelUnreadRetrieveView` rather than the channel list, for three reasons:

- **It's already polled.** `GameDetailLayout.tsx:48-51` polls that endpoint every 5 s on active games and `ChannelScreen.tsx:187` invalidates its query key on mark-read, so the badges refresh and clear with no new request — react-query dedups on the shared key. On the list endpoint they'd inherit `useGamesChannelsListSuspense`, which has no `refetchInterval`, and sit stale.
- **Total and parts can't drift.** `unread_counts_by_channel` groups by `channel_id`, and the game total is the sum of the rows returned, from that one query. Two endpoints computing unread separately can disagree mid-poll.
- **The list stays user-agnostic.** `ChannelListView` is `AllowAny` and serves public channels anonymously; unread is inherently per-user (`ChannelMember.last_read_at`). Folding it in would make an identical-for-everyone payload per-user and force an anonymous branch in the annotation.

`GameUnreadListView` keeps its game-level grouping — it spans games, so per-channel rows would be the wrong shape there. Channels with zero unread are filtered out in the database and the client defaults missing to 0. Query count for the endpoint is still pinned at 1 by `test_unread_count_single_query`, and four new tests cover the breakdown, omission of read channels, and cross-game isolation.

Not included: the in-channel "New messages" divider, which needs the requester's `last_read_at` on the retrieve endpoint. Say the word and I'll add it.

---

service/channel/views.py
johnpooch: Superseded — you called it: the guidance did need updating, and it's changed in `b000436`. The schema is internal, so operation descriptions earn nothing and the exception is gone from `CLAUDE.md:85`, `.claude/rules/backend/index.md:8`, and `.claude/rules/backend/views.md:42` (that last one actively required a docstring on every view).

One thing my earlier answer missed: removing the view docstrings *alone* would have made the schema worse, not description-free. drf-spectacular falls back to the mixin's docstring, which is why 27 operations described themselves as "Used by views that have a game parameter in the URL. Provides a get_game method…". So `common/views.py` had to be stripped too. After regenerating, that count is 0 and `openapi-schema.yaml` is 95 lines shorter.

---

PR #1212, opened 2026-08-12
Stop phase creation failing when one unit dislodges two units

### LINKED ISSUE #1165: Phase creation fails when two units are dislodged by the same unit (dislodged_by is a OneToOneField)
## Goal

Stop `create_from_adjudication_data` raising `UniqueViolation` on `unit_unit_dislodged_by_id_key` when the adjudication result names the same dislodging unit for more than one dislodged unit.

## Current behaviour

`Unit.dislodged_by` is a `OneToOneField` (`service/unit/models.py:13`):

```python
dislodged_by = models.OneToOneField(
    "unit.Unit", on_delete=models.CASCADE, related_name="dislodges", null=True, blank=True
)
```

Postgres therefore puts a global `UNIQUE` on `dislodged_by_id`. `create_from_adjudication_data` resolves each dislodger from the previous phase and bulk-creates the new phase's units (`service/phase/models.py:788-817`), with no check that two units in the batch resolved to the same previous-phase unit. When they do, `bulk_create` aborts and the new phase is never created.

Sentry `DIPLICITY-API-9J`, 2 events (2026-08-01 and 2026-08-02):

```
Failed to create phase from adjudication data for phase 67728: duplicate key value
violates unique constraint "unit_unit_dislodged_by_id_key"
DETAIL:  Key (dislodged_by_id)=(1849961) already exists.
  phase/models.py, line 817, in create_from_adjudication_data
  previous_phase: <Phase: Fall 1903, Movement (The Scrimmage of the Sclerotic Scissors (Sandbox))>
```

The game is stuck: adjudication succeeded, the new phase did not get written.

## Expected behaviour

The phase is created regardless of how the adjudicator attributes dislodgements, and the schema does not encode a uniqueness rule the adjudicator does not guarantee.

## Where

- `service/unit/models.py` — `dislodged_by` (13), `dislodged_by_requires_dislodged` check constraint (19)
- `service/phase/models.py` — `create_from_adjudication_data` unit block (770-817)
- `service/unit/serializers.py` — `get_dislodged_by` (15), reads it as a single object

## Approach

Establish first whether one unit dislodging two units is legitimate adjudicator output or a bug in the adjudicator, because the fix differs. If it is legitimate, `dislodged_by` should be a `ForeignKey` (a migration plus the `related_name` becoming a reverse set); if it is not, the adjudicator needs fixing and this becomes a guard with a logged warning rather than a crash. Note that the `OneToOne` is unique across *all* phases, not per phase, which is stricter than any game rule requires either way.

johnpooch:
## What this PR does

`Unit.dislodged_by` was a `OneToOneField`, so Postgres put a global `UNIQUE` on `dislodged_by_id`. When an adjudication result named the same dislodger for two units, `bulk_create` aborted and the new phase was never written — the game stuck with adjudication succeeded and no phase to show for it (Sentry `DIPLICITY-API-9J`). Closes #1165.

The investigation ([comment](https://github.com/johnpooch/diplicity-react/issues/1165#issuecomment-5247076772)) established that one unit dislodging two is **not** legitimate adjudicator output: `ApplyMovementOutcomesReducer._compute_next_units` only produces it when its input state already holds two units in one parent province, which is an impossible board. So this PR does two things:

- **`dislodged_by` → `ForeignKey`** (`unit/models.py` + migration `0010`). The `OneToOne` was unique across *all* phases, far stricter than any game rule, and it turned a corrupt board into a hard failure. `related_name="dislodges"` has no readers anywhere, and `UnitSerializer.get_dislodged_by` reads the forward side, so the migration is the whole change — no OpenAPI or frontend impact.
- **Report the real fault where it originates.** `PhaseManager._report_co_occupied_provinces` logs and captures to Sentry when a phase is written with two units sharing a parent province in the same state — two standing, or two dislodged. A clean board holds at most one of each per province, so a Retreat phase's legitimate dislodged-plus-occupant pair still passes. A bare FK on its own would silently persist the impossible board; this keeps it visible in the phase that introduces it rather than crashing a later resolution that happens to attack it.

Verified against the reported failure with a full-stack reproduction — a Movement phase holding Army `stp` and Fleet `stp/sc`, attacked from `mos` with support from `lvn`. On `main` that raises the exact production error (`duplicate key value violates unique constraint "unit_unit_dislodged_by_id_key"`, same `bulk_create` frame); on this branch it resolves, both dislodged units record `mos` as their dislodger, and the co-occupied province is reported.

Production currently has 32 `unit_unit` rows across 6 games where two standing units share a parent province — mostly a unit at a parent alongside a fleet at one of its named coasts (`stp` + `stp/sc`), plus one case of two fleets at an identical `spa/sc`. All of them are in old games, so nothing proves the current engine still creates the state; the new report exists to catch it if it does. The upstream cause is deliberately left unfixed here — the reported game was a `clone-to-sandbox` that has since been purged (7-day sandbox retention), so there is no evidence left to fix against.

## Checklist

- [x] Tests cover the change

Self-review raised one finding — the report skipped dislodged units, so the very board this PR fixes went unreported. Fixed in `bcd2f5e9`, with the first test now asserting the report fires.

Two tests in `phase/tests.py::TestDislodgedByAttribution`: one resolves the co-occupied board end to end and asserts both dislodged units share a dislodger and that the co-occupancy is reported, one asserts the standing-pair report fires. CI is green.

https://claude.ai/code/session_01KUhnhuDoRJM1Vp8Ci4DMuK


johnpooch:
The `test` check failed on `game/tests.py::TestGameCloneToSandbox::test_clone_to_sandbox_twice_creates_two_games` with `ModuleNotFoundError: No module named 'adjudication'`. **It is not from this branch** — it fails on `main` on its own:

```
$ git worktree add … origin/main && pytest game/tests.py -k TestGameCloneToSandbox
FAILED game/tests.py::TestGameCloneToSandbox::test_clone_to_sandbox_twice_creates_two_games
1 failed, 13 passed
```

The `adjudication` app was folded into `adjudicator`, and that one test still patches the old path. On `origin/main`, `game/tests.py` has 11 uses of `patch("adjudicator.service.start")` and exactly 1 of `patch("adjudication.service.start")` — the line above, in the test added by `a2e0c78c`. It's a one-line rename, but it belongs on `main` rather than bundled in here, so I've left it alone.

I've merged `main` into this branch and re-run the whole backend suite: **2157 passed, 8 skipped**, with that inherited failure the only red. This PR will go green once `main` does.

---

johnpooch:
`main` has been fixed, so I've merged it in and pushed `071cc279`. **CI is green** — all checks pass, including `test`. Ready for review.

---

PR #1213, opened 2026-08-13
Refactor nation app: shared draft permission, own urls, single tests module

johnpooch:
## What this PR does

Aligns `service/nation` with the backend conventions in `.claude/rules/backend/`. No wire changes — URL names (`nation-flag`, `nation-flag-svg`), paths, and response shapes are all unchanged.

- **Shared permission.** `IsNationVariantOwnedDraft` (`nation/views.py`) and `IsOwnedDraftForWrite` (`variant/views.py`) checked the same rule on different objects. Both are replaced by one `IsOwnedDraftForWrite` in `common/permissions.py`, which resolves the variant via `getattr(obj, "variant", obj)` so it works on a `Variant` or a `Nation`.
- **`Nation.flag_path`.** URL path construction moves from the serializer onto a model property; the serializer only adds the host via `request.build_absolute_uri`. No queryset change needed — `variant/models.py:61,86` already prefetches `Nation.objects.select_related("flag")`.
- **Serializer.** `color` was a `SerializerMethodField` doing a bare attribute pass-through; it is now `serializers.CharField(read_only=True)`. `flag_url` stays a method field because it needs request context.
- **`FLAG_MAX_BYTES` → `nation/utils.py`.** `admin.py` was importing it from `views.py`.
- **URL ownership.** The two flag routes move into `nation/urls.py`, included from `variant/urls.py` with `path("", include("nation.urls"))`, per the "routes belong in the owning app" rule in `index.md`. Public paths are byte-identical.
- **Tests consolidated** into a single `nation/tests.py` with one class per view (`TestNationFlagUploadView`, `TestNationFlagSvgView`) plus `TestNationFlagAdminForm` for the custom `ModelForm` validation. The `primary_flag` fixture moves to `nation/conftest.py`. Direct model tests and the backfill-migration test are dropped per `tests.md`; the sanitization assertion they covered is now made through the upload endpoint. Backfill tests in `variant/` and `province/` are untouched.
- **Guidance.** Following review, the permission-class table is removed from `.claude/rules/backend/permissions.md` — it duplicated `common/permissions.py` and went stale on every addition. The opening paragraph now points at that file instead. `CLAUDE.md` gains a rule against writing codebase inventories into guidance documents.

Skipped the optional `nation/forms.py` extraction — `architecture.md` lists `admin.py` but not `forms.py` as an app's canonical modules, so `NationFlagAdminForm` stays in `admin.py`.

## Checklist

- [x] Tests cover the change

## Verification

- `pytest -n auto --create-db` — 2153 passed, 8 skipped. (`--reuse-db` errors on this machine from a stale test database committing `primaryuser`; unrelated to this branch.)
- Regenerated `manage.py spectacular` output diffs against committed `openapi-schema.yaml` only by the `/devices/` + `FCMDevice` removal documented in `codegen.md` as environmental. The `Nation` component is byte-identical, so no codegen commit is needed.


.claude/rules/backend/permissions.md
```diff
@@ -37,6 +37,7 @@ Check the existing classes before writing a new one:
 | `IsCurrentPhaseActive` | Current phase status is `ACTIVE` |
 | `IsUserPhaseStateExists` | User has a phase state for the current phase |
 | `IsSandboxGame` / `IsNotSandboxGame` | Game is / is not a sandbox |
+| `IsOwnedDraftForWrite` | Draft variant owned by request user (works on `Variant` or `Nation`) |
```
johnpooch: Claude guidance shouldn't include lists like this. Claude can study the code itself.

Can you remove this list? Can you update CLAUDE.md to tell Claude not to create lists like this when writing guidance documents?

johnpooch: Done in 2cd25ea.

Removed the table. The opening paragraph already said to check for an existing class before adding one, so I folded "read that file" into it rather than leaving a standalone pointer.

Added to `CLAUDE.md` under **Maintaining this document**:

> **Guidance states rules, never inventories.** Do not write out a list or table of what exists in the codebase — classes, fixtures, endpoints, components, settings. Claude reads the code; a copy in a guidance document only goes stale and is wrong the moment someone adds a class without updating it. Say which file to read instead. Illustrative examples of a pattern are fine; an enumeration meant to be exhaustive is not.

I checked the other rules files for the same shape. The two remaining tables are `api-design.md`'s method/path table and the codegen-artefact table in `CLAUDE.md` — both state a rule rather than enumerate what exists, so I left them. Say the word if you want either reworked.

---

Issue #1272, opened 2026-08-28
Frontend: mustering in the staging list and on the game card


johnpooch:
## Goal

Surface mustering games in the My Games staging list and on the game card, including the inline confirm action.

## Context

Step 5 of the mustering series, and the first user-visible one. Implementation on `claude/game-mustering-state-ah1fw2` (PR #1235). Depends on step 4 for the wire fields.

## Approach

`packages/web/src/screens/Home/MyGames.tsx`: the Staging tab's `statusFilter` becomes `pending,mustering`, and mustering games are pulled out into a "Ready to start" section above a "Waiting for players" heading, with a muted line explaining that these games begin once every player confirms.

`packages/web/src/components/GameCard.tsx`: a `MUSTER_STATUS_CONFIG` badge (confirmation required / ready), a countdown to `musterDeadline` via `RemainingTimeDisplay` in the place the phase deadline line occupies, and an inline Confirm button in the join button's position, calling `useGameMusterCreate`.

Tests: the `mustering games` block added to `packages/web/src/components/GameCard.test.tsx` and the `MyGames mustering section` block added to `packages/web/src/screens/Home/MyGames.test.tsx`.

Screenshots are required in the PR description — see CLAUDE.md and the `screenshots` skill.

One thing to decide rather than copy: the branch changes the card's badge cluster gate from `(game.sandbox || isActive || isFinished) && badgeCluster` to an unconditional `badgeCluster` (`GameCard.tsx`), which affects pending cards too, not just mustering ones. Either justify that as an intended change or narrow the gate to include `isMustering` only.

Issue #1270, opened 2026-08-28
Add the muster reminder


johnpooch:
## Goal

Nudge players who have not confirmed, ahead of the muster deadline.

## Context

Step 3 of the mustering series. Separable from step 2's closed loop — the lifecycle works without it, players just get no warning before their seat is vacated. Implementation on `claude/game-mustering-state-ah1fw2` (PR #1235).

## Approach

First, a pure refactor with no behaviour change: extract `WARNING_THRESHOLDS` and `get_warning_threshold` from inside `PhaseManager.send_deadline_warnings` (`service/phase/models.py`) to module scope in `service/phase/utils.py`, and import it back.

Then the reminder half of `GameManager.arm_muster`: schedule `send_muster_reminder` at the deadline less the same lead time the phase deadline warning uses, and never re-arm once the scheduled time has passed, so a game save after it has run cannot send a duplicate. Plus `GameManager.send_muster_reminder`, its task registration in `service/game/tasks.py`, and the `muster_reminder` notification spec in `service/notification/registry.py`.

Tests: `test_entering_mustering_arms_reminder_before_deadline` and `test_reminder_is_not_rearmed_after_it_has_run` from `TestArming` in `service/game/tests/test_mustering.py`.

Roughly 110 lines plus tests.

Issue #1267, opened 2026-08-28
Add mustering stage


johnpooch:
## Goal

Land the mustering state: a `mustering` status between `pending` and `active` in which every human player confirms their seat before a full lobby starts.

## Context

When a lobby fills, mustering holds the game for a window while every human player confirms their seat. At the deadline the game starts anyway, with unconfirmed seats vacated as replaceable. Mustering is coerced on for public games and optional, default off, on private ones.

PR #1235 (branch `claude/game-mustering-state-ah1fw2`) is a complete working implementation. Every step below is a slice of it, and each Approach names the files to lift. #1235 was branched from #1232 and contains all of it, so the diff GitHub shows on that PR is not the mustering work. The mustering-only diff is the head of #1235 against a synthetic merge of `main` and #1232:

```
TREE=$(git merge-tree --write-tree origin/main origin/claude/armed-phase-resolution)
git diff $TREE origin/claude/game-mustering-state-ah1fw2
```

That is +1876/−84 across 46 files. Nothing from #1232 belongs to any of these steps. #1235 will be closed unmerged once this series is underway.

Two constraints shape the order.

**Every test drives the real API.** `muster_required` becomes settable on `GameCreateSerializer` in step 1, so a test that wants a mustering game creates a private game with `muster_required=true`, fills it through the join endpoint, and confirms through the muster endpoint. Nothing forces `Game.status` or `Member.mustered_at` directly. What is deferred to the last step is not the write path but the default — the coercion that turns mustering on for every public game, and the backfill of existing ones.

**A game must never be able to enter mustering without a way out.** That rules out landing the fill transition, the expiry job and the confirm endpoint separately: any such ordering leaves a mustering game with no confirm endpoint, which expires with every seat vacated. They are one mechanism and land together in step 2, which is the largest step and where review effort should concentrate. The reminder job is separable and follows in step 3.

### Carry-over notes

- The branch reverts `resend`, `anthropic` and `lxml` in `service/requirements.txt`, and `inspect-ai` and `datamodel-code-generator` in `service/dev_requirements.txt`, to older floors. Those are stale-branch artifacts and must not be carried over.
[...]

PR #1288, opened 2026-08-28
The mustering lifecycle: entry, exit, expiry and confirmation

### LINKED ISSUE #1269: The mustering lifecycle: entry, exit, expiry and confirmation
## Goal

Entry, exit, expiry and confirmation, as one mechanism, so the loop is closed the moment it lands.

## Context

Step 2 of the mustering series, and the largest — this is where review effort should concentrate. These pieces cannot be split: any ordering that lands the fill transition before the confirm endpoint leaves a mustering game with no way out, which then expires with every seat vacated. Implementation on `claude/game-mustering-state-ah1fw2` (PR #1235).

## Approach

`Game.enter_mustering(window_seconds)` and `Game.return_to_pending()` in `service/game/models.py`. The `start_if_full` branch: a full game with `muster_required` and a non-null effective movement duration enters mustering with a deadline of now plus that duration (`get_effective_phase_duration_seconds` — a fixed *length*, not the fixed-time mode's snapped instant) and assigns no nations. The null-duration path starts immediately, as today.

`GameQuerySet.filter_musterable`, and on `GameManager`: `arm_muster`, `_arm_muster_job`, `_is_mustered` and `start_if_mustered`, with the reminder half of `arm_muster` (`muster_reminder_job_id`, `MusterJob.REMINDER_TASK_NAME`) stubbed until step 3. `arm_muster` is idempotent and modelled directly on #1232's `arm_resolution`. The expiry task registration goes in `service/game/tasks.py`, the `post_save` receiver in the new `service/game/signals.py`, and the `import game.signals` in `GameConfig.ready` (`service/game/apps.py`).

`Game.start(notify_user_ids=...)` and the `recipients` override in `GameStartSpec.get_audience` (`service/notification/registry.py`), so players about to be vacated do not get a start notification. `start_if_mustered` starts the game *first* and removes unconfirmed members afterwards, so every seat has a nation and a phase state before it is emptied.

`POST /game/{game_id}/muster/` → `MemberMusterView` in `service/member/views.py`, permissioned on `IsMusteringGame` and `IsGameMember`, setting `mustered_at` once (idempotent, not reversible) and calling `arm_muster` so the final confirmation starts the game immediately. Bot members are exempt from confirmation, mirroring `Phase.has_unconfirmed_human`.

The `mustering_started` and `removed_from_muster` notification specs. `return_to_pending` wired into leave (`MemberDeleteView`), kick (`MemberManager.remove`), account deletion (`service/user_profile/views.py`) and the civil-disorder staging sweep (`PhaseManager._remove_from_staging_games`), keeping confirmations already given.

Regenerated schemas and clients: `docker compose up codegen`, then `npx tsc -b --noEmit` in `packages/web`.

Tests: `TestFillTransition`, `TestMusterDeadline`, `TestArming` minus the two reminder cases, `TestStartIfMustered`, `TestMusterEndpoint`, `TestRosterChange` and `TestStatusAudit` from `service/game/tests/test_mustering.py`, along with the `muster_game_factory` fixture. Roughly 330 lines plus 450 lines of tests.

Two things are worth the most review: idempotent re-arming, and the start-then-vacate ordering that gives every seat a nation and a phase state before it is emptied.

johnpooch:
## What this PR does

Closes the mustering loop: a full lobby can now enter mustering, players can confirm, the deadline can fire, and any roster change gets the game back out. Step 2 of #1267, closes #1269. Branched fresh off `main` now that #1278 (step 1) has merged.

**Entry.** `Game.start_if_full` gains one branch: a full game with `muster_required` and a non-null `get_effective_phase_duration_seconds(PhaseType.MOVEMENT)` calls the new `Game.enter_mustering(window_seconds)` — status `mustering`, deadline now + that window, no nations assigned, phase left `pending`. The null-duration path (fixed-time mode with no frequency) starts immediately, as today; `test_null_effective_duration_starts_immediately` covers it. Note this uses the effective *length*, not `get_scheduled_resolution` — a fixed-time game musters for one frequency interval rather than until the next snapped instant.

**Arming.** `GameQuerySet.filter_musterable` (mustering, non-sandbox, deadline set), and on `GameManager`: `arm_muster`, `_arm_muster_job`, `_is_mustered`. `arm_muster` reads `(muster_job_id, muster_deadline, musterable)` in one query, decides whether the game should have a job and when it should run, reconciles that against what is actually armed, and writes back only when the job id changes. It is idempotent — re-arming to the same schedule returns the existing job untouched — and it is driven by a `post_save` receiver in the new `service/game/signals.py`, wired up via `GameConfig.ready`. So `enter_mustering` arms, `return_to_pending` and `start` disarm, all without an explicit call at each site.

`_arm_muster_job` takes `task_name` because step 3 (#1270) adds a second call for the reminder; per the issue's Approach the reminder half (`muster_reminder_job_id`, `MusterJob.REMINDER_TASK_NAME`, `send_muster_reminder`, the `muster_reminder` spec and the `get_warning_threshold` move to `phase/utils.py`) is not in this diff.

**Expiry.** `GameManager.start_if_mustered`, registered as `MusterJob.TASK_NAME` in `service/game/tasks.py`. It takes a `select_for_update` on the game filtered to `mustering`, so a duplicate fire is a no-op, then **starts the game first and vacates afterwards** — every seat, confirmed or not, gets a nation and a phase state before the unconfirmed ones are emptied. That is what makes a vacated seat replaceable: `test_vacated_seat_can_be_taken_over` walks a third user through `game-member-replace` onto it.

**Notifications.** `mustering_started` (on entry, to all seated members) and `removed_from_muster` (on expiry, to the vacated). `Game.start` gains `notify_user_ids`, and `GameStartSpec.get_audience` honours a `recipients` payload when one is present, so a player about to lose their seat does not get a "the game has started" push. Every other `game_start` emitter passes nothing and keeps the default audience. Both specs follow `.claude/rules/backend/notification-copy.md` — the game name is the title, the event is named in the body's first clause, and neither body carries a prose call to action.

**Confirmation.** `POST /game/{game_id}/muster/` → `MemberMusterView`, permissioned on `IsAuthenticated` + `IsMusteringGame` + `IsGamePlayer` + `IsUnmusteredMember`. Sets `mustered_at` once — idempotent, not reversible — and calls `arm_muster`, which is what pulls the job forward: once nobody is left unconfirmed, `schedule_at` becomes `None` and the game starts immediately rather than waiting out the window. Bot members are exempt from confirming, mirroring `Phase.has_unconfirmed_human`.

**Exit.** `Game.return_to_pending()` wired into leave (`MemberDeleteView`), kick (`MemberManager.remove`), account deletion (`user_profile/views.py`) and the civil-disorder staging sweep (`PhaseManager._remove_from_staging_games`). It clears the deadline and, through the same `post_save`, cancels the armed job. Confirmations already given are kept, so a lobby that empties and refills does not make players confirm twice — `test_confirmations_survive_a_roster_change`.

## Worth a look

Both of the things #1269 flags:

1. **Idempotent re-arming.** `_arm_muster_job` keeps the existing job when `should_arm` and the armed job is still `todo` at the same `scheduled_at`; otherwise it cancels and re-defers. `test_partial_confirmation_keeps_deadline_schedule` asserts the job id is unchanged after a non-final confirmation, and `test_leaving_returns_the_game_to_pending` asserts the old job ends up `cancelled` with `muster_job_id` cleared.
2. **Start-then-vacate ordering.** `test_expiry_starts_game_and_vacates_unconfirmed_seats` asserts the vacated member has a `nation_id`, `kicked=True`, `replaceable=True`, and a phase state with `has_possible_orders=False`.

## Merges with `main`

`main` has moved under this branch several times since it opened. Two merges needed more than a clean auto-merge:

- **#1287** removed `Game.member_user_ids` in favour of `Game.seated_member_user_ids` (which filters `kicked=False` and always adds the GM). Three call sites here referenced the removed method, so the resolution ports all three rather than just the marked hunk — `GameStartSpec.get_audience` (keeping the `recipients` override, falling back to the new helper), `MusteringStartedSpec.get_audience`, and the recipient set in `start_if_mustered`. No behavioural change for mustering, since nobody is kicked in a mustering game.
- **The notification copy rewrite** made the game name the title for every spec and added a registry-wide test enforcing it. Git merged it cleanly, but the change was semantically conflicting: both mustering specs overrode `get_title()`. Both overrides are gone and the meaning now sits in the bodies, per the style guide.

Query-count assertions have been re-measured on every merge that touched them, never re-derived by arithmetic.

## Deviations from the issue's Approach

- **`muster_game_factory` defaults to `private=True`.** The branch's factory defaults to public and then passes `private=True` in the two cases that must not muster. The factory calls `create_from_template` directly, so step 8's serializer-level coercion will never reach it either way; defaulting to private keeps each test's mustering behaviour stated by its own `muster_required` argument. `test_cannot_join_a_mustering_game` asks for a public game explicitly.
- **Three tests added beyond the issue's list**, all covering stated behaviour the branch left untested: `test_null_effective_duration_starts_immediately` (the null-duration path) and the bot exemption (`test_bots_do_not_block_the_final_confirmation`, `test_expiry_keeps_unconfirmed_bots`). I checked both bot tests fail if the `BOT_KINDS` exclusions are removed.
- **`service/phase/utils.py` is untouched by this branch.** The stale branch moved `WARNING_THRESHOLDS` / `get_warning_threshold` out of `PhaseManager.send_deadline_warnings` for the reminder's benefit. Nothing here reads it, so it goes with step 3. (`main` has since changed that file on its own account, via #1339.)

None of the stale-branch artefacts #1267 lists (the requirements downgrades, the dropped `arm_resolution` calls, the merge migrations, the `legacy.ts` indentation) are in this diff, and nothing from #1232 is carried over.

## Query counts

The `post_save` receiver costs one indexed PK lookup per `Game.save()`. Deliberate: the alternative is trusting the in-memory instance's `status`/`muster_job_id`, which a `save(update_fields=…)` on a stale object would get wrong in exactly the direction that leaves a job armed on a game that has left mustering. Three assertions move: game create, sandbox create (which saves twice) and `test_start_query_count_with_preferences`. The deltas this PR is responsible for are **+1, +2 and +1** and have not changed; the absolute numbers in the assertions track `main`, which has moved them several times since this PR opened, so read them from the diff rather than from this description.

## Codegen

One new path and no new components, so the delta is the `/game/{gameId}/muster/` block in both schemas plus `gameMusterCreate`/`useGameMusterCreate` in `endpoints.ts`; `service/harness/generated/api.py` is unchanged. Generated with `FIREBASE_PROJECT_ID` satisfied by a throwaway service account per `.claude/rules/backend/codegen.md`, so `/devices/` and `FCMDevice` are untouched. Re-running codegen after the merge produced no further diff.

No migrations — the model fields all landed in step 1.

## Tests

The full backend suite (`pytest -n auto --create-db`) and the full frontend suite (`vitest run`) are green on the current head, with `npx tsc -b --noEmit` and `npm run lint` clean. Both totals climb with every `main` merge, so the useful figure is this PR's own: `service/game/tests/test_mustering.py` holds 30 tests across `TestFillTransition`, `TestMusterDeadline`, `TestArming`, `TestStartIfMustered`, `TestMusterEndpoint`, `TestRosterChange` and `TestStatusAudit`, alongside step 1's `TestGameCreateMusterRequired`.

No screenshots: nothing visible in the web app changes. The only frontend diff is the generated mutation hook, which nothing calls until step 5 (#1272).

## Checklist

- [x] Tests cover the change
- [x] Screenshots embedded in the PR description for any visual changes (see CLAUDE.md) — n/a, no visual changes

https://claude.ai/code/session_01Cbt5ynJUeZXJ6MPuAnQ9vR


johnpooch:
Merged `main` again (#1232 has now landed, along with five other commits). Two files conflicted — `service/member/views.py` and `service/game/tests.py` — but the interesting part is a conflict git *didn't* flag.

**`service/member/views.py`** — a genuine both-sides import conflict. `main` added `from phase.models import Phase`, this branch added `from game.models import Game`. Both are used in the merged file (`Phase.objects.arm_resolution` at the seat-replacement path, `Game.objects.arm_muster` in `MemberMusterView`), so the resolution keeps both.

**Query counts — three assertions were silently wrong after a clean merge.** This is worth flagging because git reported no conflict on two of them. `#1232` and this branch each independently added one query to the same code paths, and each side had written the *same new number* into the assertion, so git auto-merged them as agreeing when in fact the effects stack:

| Test | base | `main` | this branch | merged (measured) |
| --- | --- | --- | --- | --- |
| `test_create_game_query_count_*` | 48 | 49 | 49 | **50** |
| `test_create_sandbox_game_query_count_*` | 54 | 55 | 56 | **57** |
| `test_start_query_count_with_preferences` | 19 | 20 | 20 | **21** |

Only the sandbox pair conflicted textually (55 vs 56). The other two merged clean at 49 and 20 and were wrong — caught by running the suite, not by the merge. Corrected to the measured values.

This supersedes the "Query counts" section in the description above, which states the pre-merge deltas.

Re-verified on the merge commit:

- Backend: **2348 passed, 8 skipped** (`pytest -n auto --create-db`)
- Frontend: **657 passed** across 72 files, `npx tsc -b --noEmit` and `npm run lint` clean

No behavioural change in the resolution — the only edits are the import line and the three assertion values.

---

service/member/views.py
```diff
@@ -108,6 +111,30 @@ def post(self, request, *args, **kwargs):
         return super().post(request, *args, **kwargs)
 
 
+class MemberMusterView(SelectedGameMixin, generics.GenericAPIView):
```
johnpooch, line 114: Should this not be an update view?

service/member/views.py
```diff
@@ -108,6 +111,30 @@ def post(self, request, *args, **kwargs):
...
+    serializer_class = EmptySerializer
```
johnpooch, line 115: Why empty serializer?

service/member/views.py
```diff
@@ -108,6 +111,30 @@ def post(self, request, *args, **kwargs):
...
+        IsGameMember,
+    ]
+
+    @extend_schema(request=EmptySerializer, responses={200: MemberSerializer})
+    def post(self, request, *args, **kwargs):
+        """Confirm the requesting user's seat in a mustering game."""
+        game = self.get_game()
+        member = get_object_or_404(Member, game=game, user=request.user)
+
+        if member.mustered_at is None:
```
johnpooch: Shouldn't we just return an error if this happens? It should only be callable by unmustered players

service/game/models.py
```diff
@@ -808,9 +916,31 @@ def start_if_full(self):
...
         return True
 
+    def enter_mustering(self, window_seconds):
+        self.status = GameStatus.MUSTERING
+        self.muster_deadline = timezone.now() + timedelta(seconds=window_seconds)
+        self.save()
+
+        emit("mustering_started", game=self)
+
+    def return_to_pending(self):
```
johnpooch, line 933: We don't need to kill any pending tasks here?

service/game/models.py
```diff
@@ -281,6 +291,100 @@ def with_retrieve_data(self):
...
+        if not should_arm:
+            return None
+
+        return procrastinate_app.configure_task(
+            task_name,
+            schedule_at=schedule_at,
+            lock=MusterJob.lock_for_game(game.pk),
+        ).defer(game_id=game.pk)
+
+    def _is_mustered(self, game):
```
johnpooch: Should this not be a method on the Game model instead? It is a fact about the game

service/game/models.py
```diff
@@ -281,6 +291,100 @@ def with_retrieve_data(self):
...
+        return not (
+            Member.objects.filter(game=game, mustered_at__isnull=True)
+            .exclude(user__profile__kind__in=UserKind.BOT_KINDS)
+            .exists()
+        )
+
+    def start_if_mustered(self, game_id):
```
johnpooch, line 345: If any players aren't mustered, shouldn't we go back to pending?

service/game/models.py
```diff
@@ -774,9 +878,13 @@ def start(self, current_phase=None, members=None):
...
             self.status = GameStatus.ACTIVE
             self.started_at = timezone.now()
+            self.muster_deadline = None
             self.save()
 
-            emit("game_start", game=self)
+            if notify_user_ids is None:
+                emit("game_start", game=self)
+            else:
+                emit("game_start", game=self, recipients=notify_user_ids)
```
johnpooch, line 876: Why is this needed?

service/game/models.py
```diff
@@ -808,9 +916,31 @@ def start_if_full(self):
         playable_nations = self.variant.nations.filter(non_playable=False).count()
         if self.members.count() != playable_nations:
             return False
+
+        if self.muster_required:
+            window_seconds = self.get_effective_phase_duration_seconds(PhaseType.MOVEMENT)
```
johnpooch, line 910: Can window_seconds ever be none?

service/game/models.py
```diff
@@ -281,6 +291,100 @@ def with_retrieve_data(self):
...
+
+    def start_if_mustered(self, game_id):
+        with transaction.atomic():
+            game = (
+                self.select_for_update()
+                .filter(pk=game_id, status=GameStatus.MUSTERING)
+                .first()
+            )
+            if game is None:
+                return None
```
johnpooch, line 353: Should we send an error to sentry? This shouldn't happen.

service/game/models.py
```diff
@@ -281,6 +291,100 @@ def with_retrieve_data(self):
...
+
+            unmustered = list(
+                game.members.filter(mustered_at__isnull=True)
+                .exclude(user__profile__kind__in=UserKind.BOT_KINDS)
+                .select_related("user")
+            )
```
johnpooch: A method on the game would be good here, e.g. `unmustered_members`

service/game/models.py
johnpooch: Agreed — moved to `Game.is_mustered()` in 2a46121, with the manager calling `game.is_mustered()`.

---

service/game/models.py
johnpooch: Added `Game.unmustered_members()` in 2a46121. `is_mustered()` is now defined in terms of it, so the "non-bot member with no `mustered_at`" filter exists once instead of twice.

---

service/member/views.py
johnpooch: Done in 2a46121 — new `IsUnmusteredMember` permission on the view, so a second confirm is a 403 rather than a silent no-op. That matches how `IsInCivilDisorder` gates `CivilDisorderRecoveryView`.

`test_confirming_twice_keeps_the_original_timestamp` becomes `test_confirming_twice_is_rejected` (403, timestamp unchanged). No frontend impact — the confirm button is already hidden once `musterStatus` is `confirmed`.

---

service/member/views.py
johnpooch: It's shaped after `CivilDisorderRecoveryView` (`member/views.py:185`), which is the closest existing analogue: `GenericAPIView` + `EmptySerializer` + `post`, mutating one field on the requester's own member and returning `MemberSerializer`. `MemberDeleteView` and `MemberKickView` follow the same "action endpoint" shape with `DestroyAPIView`.

An `UpdateAPIView` would fit the REST semantics better in isolation, but it would make this the only member action on `PUT`/`PATCH` and would need a real serializer with a writable field (see the next thread). Happy to switch if you'd rather the convention move that way — it's cheap right now, since nothing calls `useGameMusterCreate` until step 5 (#1272) — but as it stands I kept it consistent with the neighbours.

---

service/member/views.py
johnpooch: The request carries no body — confirming is a bare "yes, I'm here"; the game and the member both come from the URL and the authenticated user. `EmptySerializer` is what the other bodiless member actions use (`MemberDeleteView:47`, `MemberKickView:66`, `CivilDisorderRecoveryView:186`), and it gives drf-spectacular something concrete to emit instead of an untyped request body.

The response is typed separately via `@extend_schema(responses={200: MemberSerializer})`, so the generated client still returns a `Member`.

---

service/game/models.py
johnpooch: Yes. `get_effective_phase_duration_seconds` (`game/models.py:680`) returns `None` for a fixed-time game whose frequency for that phase type is unset:

```python
if self.deadline_mode == DeadlineMode.FIXED_TIME:
    frequency = self.get_phase_frequency(phase_type)
    interval = FREQUENCY_INTERVALS.get(frequency) if frequency else None
    return int(interval.total_seconds()) if interval else None
```

There'd be no sensible muster window in that case, so the guard falls through to `self.start()` and the game starts immediately as it does today. `test_null_effective_duration_starts_immediately` covers it.

Inside `enter_mustering` itself the parameter can't be `None` — the caller is the only entry point and it checks first.

---

service/game/models.py
johnpooch: This is the branch that stops a player who is about to lose their seat from getting a "your game has started" push.

`start_if_mustered` starts the game *before* it vacates the unconfirmed seats, so at the moment `start()` emits, the unconfirmed players are still members and would land in the default audience. It passes `notify_user_ids` = seated members minus the unconfirmed, and those players get `removed_from_muster` instead.

`notify_user_ids=None` (every other caller) keeps the default audience, so no other start path changes. `test_expiry_notifications_split_by_confirmation` pins both halves.

---

service/game/models.py
johnpooch: It is killed, just not from here. `return_to_pending` calls `self.save()`, and the `post_save` receiver in `game/signals.py` runs `Game.objects.arm_muster(instance)` on every `Game` save. Once the status is back to `pending` the game no longer matches `filter_musterable`, so `should_arm` is `False` and `_arm_muster_job` cancels the armed job and clears `muster_job_id`.

That indirection is the reason for the receiver — `enter_mustering` arms, `return_to_pending` and `start` disarm, and no caller has to remember. `test_leaving_returns_the_game_to_pending` asserts the old job ends up `cancelled` with `muster_job_id` cleared, so the wiring is pinned rather than assumed.

Happy to add an explicit `arm_muster` call here if you'd prefer it visible at the site rather than implied by the save.

---

service/game/models.py
johnpooch: I'd argue not — it's reachable without anything being wrong, so it would be recurring noise rather than a signal.

The job is armed for the deadline, and every exit from mustering (leave, kick, account deletion, the civil-disorder sweep) cancels it through the `post_save` → `arm_muster` path. Cancellation and firing aren't atomic with respect to each other: if a player leaves in the moment the worker has already picked the job up, the task runs against a game that is legitimately back in `pending`, and the `status=MUSTERING` filter is what makes that a no-op. The same guard absorbs a duplicate delivery, which procrastinate allows.

So this is the concurrency guard doing its job, not an invariant breaking. If you'd like it observable I'd rather add a `logger.info` alongside the other task logging in `game/tasks.py` than a Sentry event — say the word and I'll put one in.

---

service/game/models.py
johnpooch: This is the one place I don't want to change unilaterally, because it's the feature's central product decision rather than a local implementation choice — so flagging it rather than pushing.

What's implemented is what #1267 and #1269 specify: at the deadline the game **starts anyway**, with the unconfirmed seats vacated as replaceable. Going back to `pending` instead is a coherent alternative, but it changes the feature rather than the code:

- **Nothing would ever expire.** Returning to pending frees the seats and re-opens the lobby, so a single player who never confirms holds up a full lobby indefinitely — and the next time it fills, mustering starts over. That's the failure mode mustering exists to fix.
- **The replaceable-seat mechanic loses its purpose.** `start_if_mustered` deliberately starts *before* vacating so each seat has a nation and a phase state and can be taken over mid-game (`test_vacated_seat_can_be_taken_over`). If we never start, that machinery is dead code.
- **The copy already promises the other behaviour.** `removed_from_muster` says "The seat is open for another player to take", and the mustering alert says the game "starts once every player confirms … confirm before the deadline or you'll lose your seat".

Worth being clear about what we already *do* return to pending for: any roster change during mustering (leave, kick, account deletion, civil-disorder sweep) calls `return_to_pending`, because the lobby genuinely isn't full any more. The deadline case is different — the lobby is still full, the players just haven't all answered.

If you do want expiry to return to pending, that's fine but it should be a decision on #1267, since steps 5–8 and the notification copy all assume start-and-vacate. Tell me which way and I'll follow: leave as is, or switch and update the copy, the tests and the issue.

---

johnpooch:
Merged `main` again for #1311 (game master seated as a `Member`). It conflicted, and resolving it surfaced a behavioural break worth calling out.

**The conflict** was just the permissions import in `service/member/views.py` — #1311 renamed `IsGameMember` → `IsGamePlayer`, `IsNotKickedGameMember` → `IsNotKickedGamePlayer`, `IsNotGameMember` → `IsNotGamePlayer`, while this branch had added `IsMusteringGame` and `IsUnmusteredMember` to the same line. Resolution keeps both sides.

**The part git did not flag.** Now that the GM has a `Member` row, `Game.unmustered_members()` — which filtered `self.members` — counted the game master as an unconfirmed seat. Two consequences, neither of which any existing test caught:

- `is_mustered()` could never be true for a GM-run game, so the final player's confirmation would not pull the start forward; the lobby would sit until the deadline.
- At the deadline, `start_if_mustered` would have **vacated the game master's seat** — `kicked=True`, marked replaceable.

Fixed by scoping to players, which is what the rest of the model already does (`start_if_full:912` and `delete_if_empty_pending:980` both count `members.players()`):

- `Game.unmustered_members()` → `self.members.players().filter(...)`
- `IsUnmusteredMember` resolves the requester via `game.members.players()`
- `MemberMusterView` follows the rename to `IsGamePlayer`

Added `test_game_master_does_not_block_or_lose_a_seat_in_mustering`, which seats a GM alongside two players, confirms both, and asserts the GM is neither in `unmustered_members()` nor kicked once the game starts. I checked it fails if the `players()` filter is removed.

Re-verified on the merge commit: backend **2385 passed, 8 skipped**; `npx tsc -b --noEmit` and `npm run lint` clean.

The seven threads above are still open on your side; nothing in this merge changes any of them.

---

johnpooch:
The `.github/dependabot.yml` check is red on this head, and it is not this PR's.

`.github/dependabot.yml` here is byte-identical to `main`'s (`git diff origin/main HEAD -- .github/dependabot.yml` is empty) — this PR does not touch it. The invalid properties predate the file's latest change: `cooldown.semver-major-days` / `semver-minor-days` / `semver-patch-days` are not supported for the `github-actions` and `docker` ecosystems, and those keys came in with `65df6e8`. What changed is that `ccbe030` on `main` (the boto3 ignore) touched the file, so Dependabot re-parses it on every branch that now carries it, and the pre-existing errors surface as a check.

I have no way to re-run a Dependabot config check, and there is no open PR fixing it, so I am not touching it from here rather than widening this PR with an unrelated repo-config change. It does not gate merge — the required checks (`test`, `lint`, `build`, `approve`) are green or running on this head.

The fix, whenever someone picks it up, is to drop the three `semver-*-days` keys from the `cooldown` blocks of the four affected entries — `github-actions` at `/`, and `docker` at `/`, `/service` and `/packages/web` — keeping `default-days: 7`:

```yaml
    cooldown:
      default-days: 7
```

The `npm`, `pip` and `bundler` entries support the per-severity keys and can keep them as they are.

---

PR #1290, opened 2026-08-28
Stop burning an NMR extension for a player with nothing to order

### LINKED ISSUE #1280: An extension is burned for players who have nothing to order
## Goal

Stop consuming an automatic extension for a player who has no orderable units.

## Current behaviour

`_check_and_apply_nmr_extensions` selects phase states on `has_possible_orders=True` and zero orders. `has_possible_orders` is not the same as having something to do — an adjustment phase where supply centres equal units leaves a nation with `has_possible_orders=True` and no actionable unit.

`send_deadline_warnings` already guards against this, computing an actionable count per nation and skipping when it is zero, which is why no warning is sent in that state. The extension path has no equivalent guard, so the same player silently burns an extension and pushes the deadline out for the whole game.

Reproduced with an adjustment phase carrying no units and no supply centres: the member's `nmr_extensions_remaining` dropped from 1 to 0.

## Expected behaviour

A player with nothing to order does not burn an extension and does not delay the phase.

## Where

`service/phase/models.py` — `_check_and_apply_nmr_extensions`; the guard to mirror is in `send_deadline_warnings`.

## Approach

Apply the same actionable-unit count used by `send_deadline_warnings` before treating a phase state as a missed deadline.

johnpooch:
## What this PR does

Fixes https://github.com/johnpooch/diplicity-react/issues/1280. `_apply_nmr_extensions` treated `has_possible_orders=True` with zero orders as a missed deadline, but that is not the same as having something to do — an adjustment phase where a nation's supply centres equal its units leaves `has_possible_orders=True` with no actionable unit, so the player silently burned an extension and pushed the deadline out for the whole game.

`send_deadline_warnings` already computed an actionable count per nation and skipped a nation whose count was zero. That derivation now lives on `Phase` as two properties, used by both call sites:

- `Phase.actionable_units` — a per-nation count keyed on phase type: dislodged units for a retreat phase, `abs(supply_centres - units)` for an adjustment phase, and unit count otherwise.
- `Phase.members_that_require_nmr_extension` — the phase states that missed the deadline and still have an extension to spend, filtered to those with something to order. This is a single query; the `nmr_extensions_remaining > 0` check that used to be a Python list comprehension is now `member__nmr_extensions_remaining__gt=0` in the queryset.

`_apply_nmr_extensions` is two lines as a result, and `phase/utils.py` is unchanged from main.

`send_deadline_warnings` now prefetches units and supply centers into the default related caches rather than `to_attr`, so `actionable_units` reads them for free. Measured with `CaptureQueriesContext`: 13 queries for 1 phase and 28 for 4, identical before and after the refactor — the per-phase growth is the existing `emit` path, not this code. `members_that_require_nmr_extension` on a phase loaded via `with_related_data()` is 1 query.

## Checklist

- [x] Tests cover the change
- [x] Screenshots embedded in the PR description for any visual changes (see CLAUDE.md) — backend only, no visual change

Tests in `service/phase/tests.py::TestNMRExtensionsNothingToOrder` cover a movement phase with no units, an adjustment phase with balanced supply centres, an adjustment phase with neither units nor supply centres, and a retreat phase with no dislodged unit (all: no extension consumed, deadline unmoved); plus the positive cases that must keep working — an adjustment phase with a build available and a retreat phase with a dislodged unit — and a mixed phase where only the member with something to order consumes an extension.

Known gap, raised in review and not yet addressed: these drive `Phase.objects._apply_nmr_extensions` directly rather than going through the API, which is against the guidance in `.claude/skills/testing`. Converting them raises a reachability question — `has_possible_orders` derives from godip's options via `nations_with_possible_orders`, so some of the cases above may describe states that cannot arise in real play. That is being settled separately.

## Merge with main

This branch was written before #1232 landed, which renamed `_check_and_apply_nmr_extensions` to `_apply_nmr_extensions`, moved it behind `resolve_if_due`, and removed the tracing and logging from `send_deadline_warnings`. Merging main in produced one conflict, in `send_deadline_warnings`: resolved by taking main's structure (no span, no logging) with the actionable-unit change applied on top.

One test from #1232, `TestNoOpRearming::test_an_nmr_extension_rearms_at_the_new_deadline`, began failing after the merge: its `due_phase` fixture creates a phase state with `has_possible_orders=True` and no units, so under this change no extension applies and the phase resolves instead of re-arming. The fixture now gives England a unit, so the test exercises the re-arming path it is named for rather than reaching it through the bug being fixed.

Full backend suite: 2328 passed, 8 skipped.

https://claude.ai/code/session_01FTSUJGBPgDacmsi1kyi1ze

---
_Generated by [Claude Code](https://claude.ai/code/session_01FTSUJGBPgDacmsi1kyi1ze)_


service/phase/models.py
```diff
@@ -187,9 +187,17 @@ def _apply_nmr_extensions(self, phase):
...
+        actionable_units = count_actionable_units(
+            phase.type, phase.units.all(), phase.supply_centers.all()
+        )
+
         members_with_extensions = [
-            ps.member for ps in not_submitted
-            if ps.member.nmr_extensions_remaining > 0
+            ps.member for ps in candidates
+            if actionable_units.get(ps.member.nation_id, 0) > 0
         ]
```
johnpooch: Could we not convert this into a single query? Do we need the `count_actionable_units` helper? 

Perhaps simply `members_that_require_nmr_extension`? This could be a method on the phase, given that it is really just derived information about the phase and its related models. 

I see that we also use `count_actionable_units` in a later part of the file. Maybe another method on the phase called `actionable_units` would be good. Again, just a query.

johnpooch: Done in 85e5535 — both are now properties on `Phase`, and `count_actionable_units` is gone, so `phase/utils.py` is back to matching main exactly.

`members_that_require_nmr_extension` is a single query: the `nmr_extensions_remaining > 0` filter that was a Python list comprehension is now `member__nmr_extensions_remaining__gt=0` in the queryset. `_apply_nmr_extensions` is two lines as a result.

On making the whole thing one query — the actionable-unit count can't fold into that same query without a regression, for two reasons:

- `send_deadline_warnings` needs the count as a *value*, not just a filter: `total_units` goes into the notification body. So it has to produce numbers either way.
- That method loops over every active phase. It was reading `prefetched_units` / `prefetched_supply_centers`, which existed only for this one call site. I dropped the `to_attr` so the same `Prefetch` populates the default `units` / `supply_centers` caches, which lets `actionable_units` read them for free. If it were SQL aggregation instead, that becomes a query per phase on a job that runs every minute.

Measured both, with a scratch `CaptureQueriesContext` test:

- `send_deadline_warnings`: 13 queries for 1 phase, 28 for 4 — **identical before and after this commit**. The ~5/phase growth is the existing `emit` path, not this code.
- `members_that_require_nmr_extension` on a phase loaded via `with_related_data()`: **1 query**. `actionable_units` hits the prefetch cache and adds none.

I dropped the early-return short-circuit I had in the previous version. It only avoided computing `actionable_units` when nobody was missing orders, and that is free now on both call sites, so it was buying nothing and read badly as `if not not_submitted`.

Worth flagging separately: you're right that these tests drive the manager directly rather than the API. I have not changed that here — happy to convert them, but there's a reachability question attached that I'd rather settle first, so I've left it in chat rather than bundling it into this commit.

---

PR #1298, opened 2026-08-29
Make notification.deliver idempotent on retry

### LINKED ISSUE #1281: notification.deliver re-sends deliveries that were already sent
## Goal

Make `notification.deliver` idempotent so a retry does not duplicate pushes.

## Current behaviour

`deliver` is registered with `retry=3` and re-reads its batch by id with no `status` filter. Any retry — worker restart, connection blip, a failure between the transport call and the status update — re-sends every delivery in the batch, including ones already marked `sent`.

Reproduced by invoking `deliver` on a delivery already at `status=sent`: the transport was called again with the original body.

Note that procrastinate's `retry=3` resolves to `RetryStrategy(max_attempts=3)` with `wait`, `linear_wait` and `exponential_wait` all defaulting to `0`, so a retry is rescheduled immediately rather than backing off. The duplicate therefore arrives seconds later, not hours later.

## Expected behaviour

A delivery already marked `sent` is skipped on retry, and a delivery old enough to be misleading is dropped rather than sent.

## Where

`service/notification/tasks.py` — `deliver`.

## Approach

Filter on `status=PENDING` when loading the batch, and skip deliveries older than a sanity window. `PUSH_TTL` in `service/notification/utils.py` caps how long FCM buffers a message for an offline device, but it is computed at send time, so it does not bound the enqueue-to-send lag.

johnpooch:
## What this PR does

`notification.deliver` is registered with `retry=3` and re-read its batch by id with no `status` filter (`service/notification/tasks.py:19`), so any retry re-sent every delivery in the batch — including ones already marked `sent`. It now loads only `PENDING` rows, so a retry picks up exactly the deliveries that have not gone out yet.

It also drops deliveries that are too old to be worth sending: rows older than `DELIVER_MAX_AGE_HOURS` are marked with a new `expired` status instead of being sent. `PUSH_TTL` (`service/notification/utils.py:11`) caps how long FCM buffers a message for an offline device, but it is computed at send time, so it does not bound the enqueue-to-send lag.

Both filters are per-delivery, not per-batch: a batch that mixes sent and pending rows, or stale and fresh ones, still delivers the rows that qualify.

Closes https://github.com/johnpooch/diplicity-react/issues/1281

## Checklist

- [x] Tests cover the change

Four tests in `service/notification/tests.py::TestNotificationDeliver` cover the sent-skip, the partly-sent batch, the expiry, and the partly-stale batch. Full backend suite: 2309 passed, 8 skipped. No visual changes, so no screenshots. `/review-pr` has not been run against this branch.


service/notification/tasks.py
```diff
@@ -11,17 +11,38 @@
...
-    _deliver_channel(deliveries, NotificationDelivery.Channel.PUSH, _send_push)
-    _deliver_channel(deliveries, NotificationDelivery.Channel.EMAIL, _send_email)
+    fresh = _expire_stale(deliveries)
+    if not fresh:
+        return
+    _deliver_channel(fresh, NotificationDelivery.Channel.PUSH, _send_push)
+    _deliver_channel(fresh, NotificationDelivery.Channel.EMAIL, _send_email)
+
+
+def _expire_stale(deliveries):
```
johnpooch: I feel like this should be a method on the manager

service/notification/tasks.py
```diff
@@ -11,17 +11,38 @@
...
+    stale = [d for d in deliveries if d.created_at < cutoff]
+    if not stale:
+        return deliveries
+    NotificationDelivery.objects.filter(id__in=[d.id for d in stale]).update(
+        status=NotificationDelivery.Status.EXPIRED
+    )
+    logger.warning(
+        f"Expired {len(stale)} notification delivery(s) older than {DELIVER_MAX_AGE_HOURS}h "
+        f"instead of sending them"
+    )
```
johnpooch: Remove this log

service/notification/tasks.py
johnpooch: Moved to `NotificationDeliveryManager.expire_stale(deliveries, max_age)` in `service/notification/models.py`. The task passes the age in, so `DELIVER_MAX_AGE_HOURS` stays next to `PRUNE_AFTER_DAYS` in `tasks.py`. The `PENDING` filter stayed in `deliver` — it's a plain `.filter()`, which the models rule says shouldn't get a wrapper.

---

service/notification/tasks.py
johnpooch: Removed.

---

Issue #1263, opened 2026-08-28
Add release tooling to build, upload and register a bundle


johnpooch:
## Goal

Turning a merged commit into an available OTA bundle should be one command or one CI job, not a manual sequence.

## Approach

Build `packages/web`, zip `dist`, compute the sha256 of the zip (that hash is what the endpoint returns as `checksum`), upload it to the bucket, and create the `Bundle` row.

Whoever cuts the release has to state the minimum native version the bundle requires. Make that an explicit input rather than a default — getting it wrong is the failure mode this whole design exists to prevent.

Decide and record whether this runs as a Django management command, a GitHub Actions workflow alongside `.github/workflows/ios-release.yml` and `android-release.yml`, or both.

PR #1304, opened 2026-08-31
Add release tooling to build, upload and register an OTA bundle

johnpooch:
## What this PR does

Turns a merged commit into an available OTA bundle in one command and one CI job, closing https://github.com/johnpooch/diplicity-react/issues/1263.

**The decision the issue asked for: both, split where the work changes machine.**

- `release_bundle` (`service/update/management/commands/release_bundle.py`) owns everything the service knows about — zip a built `dist`, take the sha256 of the zip as the `checksum` the update endpoint serves, upload to R2, write the `Bundle` rows. It never builds the web app; `--dist` points it at an already-built directory, which is what keeps it testable and keeps Node out of the service.
- `.github/workflows/bundle-release.yml` owns the rest — Node, the web build, and one invocation of the command.

Recorded in `.claude/rules/backend/ota-releases.md`, scoped to `service/update/**` and the workflow so it loads when either is edited.

### The minimum native version is stated, never inferred

`--minimum-native-version` is required on the command and a required `workflow_dispatch` input, with no default. That is why the workflow is dispatch-only rather than firing on push to `main`: serving a bundle that calls native APIs the installed binary lacks is the failure this whole design exists to prevent, and a default is how that failure happens quietly.

`--bundle-version` is required for the same reason — the version is what orders bundles for every installed client. It is `--bundle-version` and not `--version` because Django's `BaseCommand` already owns `--version`.

### Two details worth a reviewer's eye

- **The build is shared between platforms, so the workflow passes every `VITE_*` variable *either* store release needs** — including `VITE_GOOGLE_IOS_CLIENT_ID`, which `android-release.yml` omits. An OTA bundle built without it would break Google sign-in on iOS. The zip is then uploaded once per platform under `bundles/{platform}/{version}.zip`, so each `Bundle` row owns its object and replacing one platform's bundle can't disturb the other.
- **Railway's `DATABASE_URL` is an internal hostname a GitHub runner cannot resolve** (the same constraint `.claude/skills/production/SKILL.md` documents for cloud sessions). The publish step sets `DATABASE_CONNECTION_STRING` from a `DATABASE_PUBLIC_URL` secret, which `service/project/settings.py:166` already checks ahead of `DATABASE_URL`; `railway run` supplies the R2 credentials from the service's own variables. The step fails fast when the secret is missing rather than letting Django parse an empty connection string. **This needs a `DATABASE_PUBLIC_URL` repo secret before the workflow can run** — `pr-staging.yml:230` already reads that variable off the Postgres service, so the value exists.

## Review fixes (61e3342)

`/code-review` found four issues, all fixed here with three regression tests:

- **A version could collapse onto a published one.** `parse_version` strips trailing zero components (added in #1299), so `1.6` and `1.6.0` are distinct rows under the `(platform, version)` unique constraint but compare equal. The already-published check matched the string exactly, so both could be published, and `latest_for`'s `max()` then picked between them arbitrarily. It now compares parsed versions.
- **An overlong version orphaned uploads.** `is_numeric_version` bounds no length, but both version columns are `varchar(32)`. A 42-character numeric version passed every guard, uploaded both objects, then died on the insert — leaving unregistered zips in the bucket. Both fields are now checked against their column widths before anything uploads.
- **The workflow could publish from any branch.** `workflow_dispatch` accepts any ref carrying the workflow file, so a feature branch could publish itself to production storage and the production database with no store review in the way. The job is guarded on `github.ref == 'refs/heads/main'`.
- **boto3 sat on the Django boot path.** `update/models.py` imports `parse_version` from `update/utils.py`, which imported boto3 at module scope — loading it in every web and worker process for a release-only dependency. The client and upload helper moved to `update/storage.py`.

The ref guard and the storage split are recorded in `.claude/rules/backend/ota-releases.md`. Details in [this comment](https://github.com/johnpooch/diplicity-react/pull/1304#issuecomment-5482733506).

## Checklist

- [x] Tests cover the change — 16 new tests in `service/update/tests.py`: both-platform and single-platform publishes, the zip's contents and root layout, the checksum being the sha256 of the uploaded bytes, the `application/zip` content type, both platforms getting byte-identical archives, and every refusal path (already-published version, a version that collapses onto a published one, versions longer than their column, a directory that isn't a built bundle, non-numeric versions, an unstated minimum native version, unconfigured storage) each asserting nothing was uploaded. Two of them drive the published bundle back out through the update endpoint. Full backend suite: 2365 passed, 8 skipped.
- [x] Screenshots embedded in the PR description for any visual changes — none; nothing in the web app changes.

<sub>Correction to the original description: it claimed #1299 was closed unmerged and that its commits were still carried here. That was wrong — #1299 landed on `main` as `299b898` while this branch was in flight, which is what produced the add/add conflict. Main's copy is byte-identical to this branch's, so `0832547` merges `main` in and the diff above is this issue's work alone.</sub>


johnpooch:
Ran `/code-review` against this PR (the unticked checklist item). Four findings, all verified against the code and all fixed in 61e3342.

**A version can collapse onto a published one.** `parse_version` strips trailing zero components (added in #1299), so `1.6` and `1.6.0` are distinct rows under the `(platform, version)` unique constraint but compare equal. The already-published check matched the version string exactly, so both could be published; `latest_for`'s `max()` over an unordered queryset then picked between them arbitrarily, and a client that installed one would be offered the other on every subsequent check. The check now compares parsed versions.

This one is mine — the trailing-zero stripping came from my review fixes on #1299, and it made two publishable version strings indistinguishable to the selector.

**An overlong version orphans uploads.** `is_numeric_version` puts no bound on length, but `Bundle.version` and `Bundle.minimum_native_version` are `varchar(32)`. A 42-character numeric version passed every guard, uploaded both platform objects, then died on the insert — leaving published-but-unregistered zips in the bucket, since the uploads happen before the transaction and are never undone. Both fields are now checked against their column widths before anything is uploaded.

**The workflow could publish from any branch.** `bundle-release.yml` is `workflow_dispatch`, and a dispatch can name any ref that carries the workflow file, so a feature branch could build itself and publish straight to production R2 and the production database — reaching every installed client with no store review in the way. `build-and-deploy-service.yml` is push-to-`main` only; this had no equivalent gate. The job is now guarded on `github.ref == 'refs/heads/main'`.

**boto3 was on the Django boot path.** `update/models.py` imports `parse_version` from `update/utils.py`, which imported boto3 at module scope, so every web and worker process loaded it at startup for a dependency only the release command uses — and a boto3 import failure would stop the service booting. The client and upload helper moved to `update/storage.py`; verified boto3 is absent from `sys.modules` after `django.setup()` and after importing `update.models`. The lazy-import alternative is ruled out by the top-level-import rule in `.claude/rules/backend/index.md`, hence the module split.

Three regression tests added (collapsing version, overlong version, overlong minimum native version), each asserting nothing was uploaded. The ref guard and the storage split are recorded in `.claude/rules/backend/ota-releases.md`.

Full backend suite: 2365 passed, 8 skipped. `black --check` clean.

One thing I did **not** change: the `--dist` directory is trusted to be a real build beyond the `index.html` check. That seems right for a command whose caller is the workflow, but say the word if you want it stricter.

---

PR #1349, opened 2026-09-15
MEGAPR - FRONT END

JorenC:
## What this PR does

Lets a private chat channel carry a name of its own, and uses that as the seed for a broader pass at bringing the game-detail screens in packages/web in line with the design-playground shell (chat, Game Info, Player Info, Player Profile, orders, and the map), plus a few adjacent fixes picked up while touching those screens.

This branch grew across several sessions without ever being opened as a PR, so it bundles more than one change. See the checklist at the bottom for why.

### Channel naming (service/, packages/web/)

- `Channel.title` is a new optional field, separate from `Channel.name` (the backend-formatted nation list). `name` stays as-is because it's load-bearing elsewhere: the channel-uniqueness key, the frontend's derivation of membership, and the fallback label for an unnamed channel.
- `PATCH /games/<id>/channels/<id>/` lets members rename an existing private channel. Public Press is excluded (a new `IsPrivateChannel` permission) since it's one shared channel per game. A blank title is rejected rather than treated as "clear the name."
- A rename now posts an event into the channel's own thread ("France renamed the channel to The great alliance"), so it isn't silently confusing for everyone else in it.
- The channel header shows name-over-nations for a named channel, or nations-over-players for an unnamed one, with a tooltip for anything that truncates.
- Both OpenAPI schemas and both generated API clients were regenerated for the new field and endpoint.

channel-thread
channel-rename

### Chat, Game Info, Player Info, and Player Profile redesign (packages/web/)

Rebuilt against `packages/design-playground`'s shell and prototypes:

- Chat channel rows combine into a single divided card instead of one Card each, and the unread badge is recentered.
- Game Info restructures into labelled sections (Variant, Phase deadlines, Game settings) and becomes a top-level tab rather than a "back"-reachable screen; pause / extend-deadline / clone-to-sandbox moved out of the overflow menu onto that screen.
- Player Info groups members into Players / Eliminated / a collapsible Former Players section, and gives each row a photo-with-flag-badge (or flag-with-photo-badge once a nation is assigned).
- Player Profile replaces the old Commitment/Games card pair with a 2x2 stat grid plus "Most played" and "Recent results" cards, backed by two new fields on `GET /users/{id}/` (`favouriteNation`, `recentResults`), and a fix so an in-progress game counts toward a player's total games and favourite nation, not just completed ones.
- Avatar/flag fallback initials use luminance-based contrast against the nation color instead of always rendering white, and are vertically centered.

chat-list
create-channel
channel-1to1
game-info
player-info
player-profile

### Orders, phase navigation, and map (packages/web/)

- `PhaseStepper` replaces the separate phase-select and phase-guidance components; the orders list moves onto a new `ListItem`/`ListSection` primitive, grouped by nation, including rows for units that haven't been given an order yet.
- An unfilled orders-list row now jumps straight into order creation. The map keeps its zoom level when focusing a province, re-pans to the order source even on a repeated selection, clears a stale mouseover highlight when it pans, and gets spacing below its header.
- Every chat now shows a "phase resolved" chip on a turn change — the same chip treatment the channel-rename event already had, just wired up for phase transitions too (service/channel/registry.py; no new frontend component).

orders
map-mobile

### This session's fix

Game Info and Player Info hid their back button unconditionally (added while making Game Info a top-level tab on desktop, where the sidebar nav makes a back button redundant). On mobile, though, that left no way off either screen except the bottom tab bar — there was no way back to the game list. Wired `onNavigateBack` to `navigate("/")`, the same target the other tabs use, so the mobile close button reappears.

game-info-mobile
player-info-mobile

## Approach

Channel naming was built backend-first — model field, then endpoint, then event, then codegen — before any frontend work touched it. The screen redesigns then worked outward from `packages/design-playground`, which prototypes each screen's target layout before it gets rebuilt into the app (see CLAUDE.md's design-playground boundary). The map/orders polish commits accumulated afterward, picked up while continuing to exercise the game-detail screens that had just been rebuilt.

## Checklist

- [ ] This PR does one thing — no unrelated fixes, refactors, or drive-by cleanups bundled in. **Not checked**: this is 42 commits from several sessions that never got opened as a PR along the way. Splitting it in hindsight would mean re-deriving history split points with more risk than value at this point.
- [ ] For PRs of any significant complexity: I ran `/review-pr` against this PR in Claude Code and addressed its findings. **Not yet run** — flagging for the reviewer given the size, rather than skipping it silently.
- [x] Tests cover the change — the `service/` commits each carry their own test changes (e.g. `service/channel/tests/`); frontend changes rely on existing screen/component tests plus the fixtures added for the named channel.
- [x] Screenshots embedded above for every visible change, except the two follow-up fixes below (not yet captured — text description only for now).

### Side-panel title alignment fix

The side-panel titles (e.g. "Chats") sat slightly right of the panel content below them. `GameDetailAppBar` always rendered an empty left-section `div`, and the outer flex's `gap-3` still added spacing before the title for that empty slot, offsetting it from the `px-3` content padding. Only render the left section when there's an actual back button or `leftButton`.

screenshot-1789560530188-1

### Two more follow-up fixes

- Orders screen: a unit that hasn't received an order in the active phase now shows a transparent red X on its avatar instead of a plain dashed circle, so an unfilled order stands out in the list. A resolved historical phase's unfilled units are left as before (unfilled there doesn't mean "still needs attention").
- Create-channel: member rows previously rendered as their own individual Card each; they're now combined into a single divided card, matching the pattern already used for the chat list and Player Info.

(No screenshots yet for these two — will follow up.)


packages/web/src/components/GameMap.tsx
```diff
@@ -179,6 +193,29 @@ const GameMap: React.FC = () => {
     }
   };
 
+  useEffect(() => {
+    if (!sourceParam) return;
```
JorenC, line 197: 🛑 **blocking**: The `source` search param is only cleared when `wizard.nextField === "source"`. If a player clicks an unfilled orders-list row (setting `?source=X`) while already mid-flow on a different order (e.g. `nextField` is `target`), this effect no-ops and leaves `source=X` in the URL. Once the in-progress order completes and `wizard.reset()` returns `nextField` to `source`, the stale param fires and silently auto-selects province X as the source for whatever order the player starts next. Move the param-clearing `setSearchParams` call outside the `if (wizard.nextField === "source")` block so it always clears once consumed or on mismatch.

packages/web/src/components/PlayerInfoContent.tsx
```diff
@@ -128,217 +184,250 @@ export const PlayerInfoContent: React.FC = () => {
...
+        <div
+          className={cn(
+            "flex flex-1 min-w-0 items-center gap-3",
+            member.userId && "cursor-pointer"
+          )}
+          role={member.userId ? "link" : undefined}
+          tabIndex={member.userId ? 0 : undefined}
+          aria-label={member.userId ? `View profile for ${member.name}` : undefined}
+          onClick={member.userId ? () => navigate(profilePath(member)) : undefined}
+          onKeyDown={
```
JorenC, line 210: 💡 **suggestion**: The row's `onKeyDown` fires `navigate(profilePath(member))` on any Enter press that bubbles up, but the nested "Replace"/"Invite replacement"/nation-preference buttons only stop propagation on `onClick`, not on the keydown itself. Pressing Enter on one of those nested controls therefore also triggers the outer profile navigation alongside the button's own action.

JorenC (commented):
### PR Review

This PR adds channel naming end-to-end (backend model/serializer/permissions/event + frontend rename screen) and redesigns the game-detail screens to match the design-playground shell; the backend changes and the bulk of the frontend restructuring look correct and are well covered by new tests. Two issues survived verification against the actual code paths: a stale `?source=` query param in GameMap that can hijack a later, unrelated order once the wizard cycles back to needing a source (reachable in the normal desktop split-view flow), and a keyboard-only double-navigation bug in the new PlayerInfoContent row markup from nested interactive elements inside a keydown-handled parent.

**2** inline comment(s) (1 blocking, 1 suggestion).

packages/web/src/components/PlayerProfileContent.tsx
```diff
@@ -76,42 +77,85 @@ export const PlayerProfileContent: React.FC<PlayerProfileContentProps> = ({
...
+                profile.totalGames > 0
+                  ? `${formatPercent(profile.soloWins / profile.totalGames)} of games`
+                  : undefined
               }
             />
-            <StatRow
-              label="NMR Rate"
-              value={formatPercent(profile.nmrRate)}
-              info="The percentage of movement phases where this player submitted no orders, based on their last 10 games."
+            <StatTile label="Draws" value={profile.draws} />
```
JorenC, line 227: 🛑 **blocking**: The old profile stats (Total Games / Solo Wins / Draws / Losses) are replaced by this 2x2 grid, but `losses` has no slot in it — Reliability took the fourth tile instead. `profile.losses` is still returned by the API and set in the test fixture (`losses: 7`), but nothing renders it anymore, and no test asserts the grid's labels so this slipped through silently.

packages/web/src/components/GameInfoContent.tsx
```diff
@@ -119,168 +96,190 @@ export const GameInfoContent: React.FC<GameInfoContentProps> = ({
...
+      label: "Deadline extensions",
+      value:
+        game.nmrExtensionsAllowed > 0
+          ? `${game.nmrExtensionsAllowed} per player`
+          : "None",
+      info: "If a player does not submit orders on time, a deadline extension is used.",
+    },
+  ];
+
+  const settingsRows: SettingsRow[] = [
```
JorenC, line 123: 💡 **suggestion**: The old Game Info screen always showed a "Number of nations" row (`variant.nations.length`). None of `settingsRows`, `phaseDeadlineRows`, or `variantInfoRows.ts` carry it forward for active/completed games — it's now only used internally to build the pending-game "will start once N players have joined" copy in `GameStatusAlerts.tsx`. Lower severity since it's purely informational, but it's a value that used to always be visible and now only shows before the game starts.

JorenC (commented):
### PR Review

Focused pass on regressions in game logic, phases, and the redesigned game-detail screens. The bulk of the restructuring (pause/extend/clone moved to Game Info, PhaseGuidance folded into PhaseStepper/OrdersScreen, phase-resolved chat chip) checks out — functionality moved, but I could still trace it to a new home. Two things did drop without a replacement.

**2** inline comment(s) (1 blocking, 1 suggestion).

packages/web/src/components/PlayerProfileContent.tsx
JorenC: Acceptable — losses can stay backend-only for now. It's derivable by the player anyway (games − wins − draws), the only wrinkle being ongoing games muddy that arithmetic slightly since `totalGames` now includes them, but that's a minor discrepancy, not worth a UI slot.

packages/web/src/components/GameInfoContent.tsx
JorenC: Acceptable for now — pending games are moving into the Game Shell next, so this screen's variant-nation-count won't be needed here. Player count/open-slot visibility already lives in the Player Info tab.

JorenC:
Follow-up on the review above: both findings are acceptable as-is for now.

- **Losses stat**: fine to leave backend-only — derivable by the player as games − wins − draws, with a minor wrinkle that `totalGames` now includes ongoing games, but not worth a UI slot for that.
- **Number of nations row**: fine to drop — pending games are moving into the Game Shell next, so this screen won't need the variant-nation-count for pending games, and player count/open-slot visibility already lives in the Player Info tab.

One more observation from testing the pending-game state on this branch: the pending game's Game Info screen looks a bit off now — the variant card opens an info dialogue instead of a dedicated view, which isn't optimal for that screen. Acceptable for now since pending games move into the Game Shell next anyway, but flagging it so it isn't forgotten once that migration happens.

packages/web/src/components/PlayerInfoContent.tsx
JorenC: Resolved with Claude in later commit

packages/web/src/components/GameMap.tsx
JorenC: Resolved with Claude in later commit

johnpooch:
Marking this as closed. Used just as a reference for smaller PRs.

PR #1360, opened 2026-09-17
MegaPR-2: Redesign screen headers for Map, Orders, Chats, Game Info, Player Info

JorenC:
## What this PR does

Same change as #1354. That PR was opened against the `sidebar-panel-shell` branch while #1353 (MegaPR-1, the sidebar icon rail) was still open, and merged into the `sidebar-panel-shell` branch ref itself — but by then #1353 had already landed on `main` and that branch should have been retired. This PR re-targets the same commits at `main` so the header redesign actually ships.

Second slice pulled out of #1349. Ports the shared header redesign to all five GameDetail screens: Map, Orders, Chats, Game Info, Player Info.

- `GameDetailAppBar` (`screens/GameDetail/AppBar.tsx`) is ported wholesale from #1349: left-aligned truncating title with a tooltip on overflow, no separator, no explicit background (inherits the panel's `bg-sidebar`), only renders the left slot when there's actually a back button.
- New `PhaseStepper.tsx` component (also ported wholesale) replaces `PhaseSelect`/`PhaseGuidance` on Map and Orders: phase name + popover phase picker as the title, prev/next chevrons as an action. Both old components are now dead and removed.
- Chats: title "Chat" → "Chats", adds a header create-channel icon button.
- Game Info: title "Game Info" → the game's name, adds a header join-game button (shown when `game.canJoin`).
- Player Info: title "Player Info" → "Players", adds a header share button; back button now returns home instead of to the map (matches the other GameDetail screens now that these are top-level sidebar destinations from #1353).

## Not changed

Content bodies are untouched — Orders' order list, the chat channel list, and `GameInfoContent` all keep their current layout exactly as on `main`. The full content redesigns for these screens (labelled Game Info sections, Eliminated/Former grouping on Player Info, card-based chat rows, unfilled-order-row styling on Orders) stay out of this PR — they're the next slices in the stack (#1355, #1357, #1358, #1359).

## Known temporary artifacts

Scoping this to headers only (rather than pulling in the content redesign) leaves two rough edges, both discussed and accepted rather than worked around with new code:

- **Map and Orders keep `GameDropdownMenu`** (pause / extend deadline / leave / delete game) alongside the new `PhaseStepper`, instead of dropping it as #1349 does. #1349 can drop it because those actions move onto the Game Info screen via a new `GameAdminActions` component — which is part of the content redesign this PR defers. Removing the menu now with no replacement would be a real functional regression, so it stays until that later slice lands.
- **Game Info shows the game name twice** — once in the new header, once in `GameInfoContent`'s own `CardTitle`. #1349's fix (a `showTitle` prop) is wired into a fully rewritten `GameInfoContent` layout, so there's no copy-paste fix available without writing new, non-ported code. Left as-is; will resolve naturally once the Game Info content slice lands.
- **Chats has two create-channel entry points** — the new header icon button and the existing footer button. #1349 keeps a similar pair (header icon + full-width inline button) after its content redesign, so this isn't a new pattern, just today's footer button standing in for #1349's inline one.

## Review follow-ups

- **Fixed:** `GameDetailAppBarProps.hideBackButton` was added but never consumed anywhere in this PR or the rest of the stack. Removed the prop and its one internal usage in `showBackButton` — it was dead scaffolding per this repo's "absent beats dormant scaffolding" convention.
- **Accepted as-is:** the rewritten `showBackButton` logic (`variant === "secondary" || isMobile`) changes desktop header behavior for six screens outside this PR's stated scope (`ChannelCreateScreen`, `ChannelScreen`, `DrawProposalsScreen`, `PlayerProfileScreen`, `ProposeDrawScreen`, `ReplaceScreen`): they used to show no left button and a close X on desktop; now they show a back arrow instead. This is acceptable — those screens get their own pass later in this stack, and a back arrow vs. a close X are functionally equivalent here (both just navigate away).

## Screenshots

map
orders
chats
game-info
player-info

## Checklist

- [x] Tests cover the change — carried over from #1354, full suite passing at that time
- [x] Screenshots embedded in the PR description for the visual changes

https://claude.ai/code/session_01CriqQNexbZJwKuL1h1trLV


packages/web/src/screens/GameDetail/AppBar.tsx
```diff
@@ -12,6 +11,7 @@ interface GameDetailAppBarProps {
   leftButton?: React.ReactNode;
   rightButton?: React.ReactNode;
   variant?: "primary" | "secondary";
+  hideBackButton?: boolean;
```
JorenC: 💡 **suggestion**: `hideBackButton` is added to `GameDetailAppBarProps` but has zero call sites anywhere in this diff or the rest of the stack. Per CLAUDE.md's "absent beats dormant scaffolding" convention, either wire it up or drop it until a caller actually needs it.

packages/web/src/screens/GameDetail/AppBar.tsx
```diff
@@ -36,52 +46,58 @@ const GameDetailAppBar: React.FC<GameDetailAppBarProps> = ({
     }
   };
 
-  const showCloseButton = variant === "secondary" && !isMobile;
-  const showBackButton = isMobile;
+  const showBackButton =
```
JorenC: This changes desktop header behavior for six screens outside this PR's stated scope (`ChannelCreateScreen`, `ChannelScreen`, `DrawProposalsScreen`, `PlayerProfileScreen`, `ProposeDrawScreen`, `ReplaceScreen`) — they used to show no left button and a close X on desktop; now they show a back arrow instead. Confirmed acceptable: those screens get their own pass later in this stack, and back-button vs. close-X are functionally equivalent here (both just navigate away), so no action needed in this PR.

JorenC (commented):
### PR Review

Follow-up on the header redesign: one dormant-prop suggestion, plus a note on an out-of-scope behavior change that's acceptable given the rest of the stack.

**2** inline comment(s) (0 blocking, 2 suggestion).

PR #1362, opened 2026-09-18
MegaPR-3.1 - [MicroPR] - Less aggressive missing orders

JorenC:
## Summary
- Mutes the missing-order indicator on active phases: replaces the destructive-red hold icon/border with a border-toned hexagon and "!" badge, matching the styling already used for the historical unsubmitted-order variant.

## Test plan
- [ ] Visually verify the missing-order indicator on an active phase in both light and dark mode


JorenC:
BTW @johnpooch  you said you gave us merge power, but it still requires a review before I can merge. I'm not against that, but the 'merge power' then becomes an empty power :). 

I thought I'd just merge this to prevent some overhead, as it's such a microPR

PR #1380, opened 2026-09-22
MegaPR-7: Redesign the chat channel screens

JorenC:
## Summary

Frontend-only redesign of the chat/channel screens, split out of #1349. This is the half of the channel-naming slice that doesn't need a backend change — the rename feature itself (custom channel titles) will follow as a separate PR based on this branch, since it needs new backend fields and would otherwise duplicate this redesign.

- Card-based channel list: per-channel avatars (composite flag avatars for private channels, a megaphone badge for public ones), unread badges, latest-message preview.
- Redesigned create-channel screen: card-based member picker with nation flags and role labels (bot), replacing the old checkbox list.
- Channel header now shows a two-line title (channel name + the players behind it) instead of a flag-avatar-plus-name row, with tooltip-on-truncation for both lines.
- Message thread: group vs. direct channel detection now controls whether sender labels are shown on bubbles; game-master messages get a distinct square avatar treatment.
- Consistent `AppBar` height/title sizing across `GameDetail` screens (needed for the two-line channel header).

## Screenshots

Chat list:

chat list

Create channel:

create channel

Channel thread:

channel thread

## Test plan

- [x] `npx vitest run` — all 752 tests pass (docker)
- [x] `npx eslint` on changed files — clean (docker)
- [x] `npx tsc -b --noEmit` — clean (docker)
- [x] Manually verified chat list, create-channel, and channel thread screens against mock fixtures

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1381, opened 2026-09-22
MegaPR-8: Add custom channel names

JorenC:
## Summary

Backend + the coupled frontend half of the channel-naming slice split out of #1349, layered on top of #1380 (MegaPR-7). This is the part that actually needs a backend change — the channel `title` field, a rename endpoint, and channel-event system notices in the thread.

- New `title` field on `Channel` (`migrations/0007_channel_title.py`), optional at creation, settable afterwards via a new `ChannelUpdateView` (PATCH), restricted to private channels the user is a member of (`IsPrivateChannel` permission).
- New `ChannelRenameScreen`, reachable via a rename button in the channel header (private channels only).
- Renaming emits a `channel_renamed` event, rendered as a system notice in the thread at the point it happened (`migrations/0008_channel_event_payload.py` adds the `payload` field channel events need to carry the rename text). The same `ChannelEvent` display mechanism now also surfaces `phase_resolved`/`phase_resolved_early` notices in every channel.
- `getChannelDisplayName`/`getChannelSubtitle` prefer the custom title when set, falling back to the nation-derived name from #1380 otherwise.
- Create-channel screen gets its optional "Channel name" field back.

## Why split from #1380

`channel.title` and `ChannelEvent` are new fields sourced from the generated API client — code referencing them won't compile until the backend ships. Splitting let #1380 ship the visual redesign immediately without waiting on backend review, with this PR adding just the title/event-consuming pieces on top.

## Screenshots

Rename button in a private channel's header:

rename button

Rename screen:

rename screen

Create channel with the name field:

create channel with name

## Test plan

- [x] `python -m pytest channel -v` — 95 passed (docker)
- [x] `npx vitest run` — all 758 tests pass (docker)
- [x] `npx eslint` on changed files — clean (docker)
- [x] `npx tsc -b --noEmit` — clean (docker)
- [x] Manually verified rename button, rename screen, and title field on create against mock fixtures
- [x] Full backend suite (`pytest -n auto --reuse-db`) — passing (a batch of unrelated `variant`/`victory` test errors on the first run reproduced as xdist/parallel-worker flakiness, not a regression — isolated reruns of the same tests pass)


JorenC:
Hey John,

You closed this one without merging. Any issue?

JorenC:
What was the reason to close-not-merge this PR? I can update if there is an issue.

JorenC:
What was the reason to close-not-merge this PR? I can update if there is an issue.

JorenC (commented):
### PR Review

This PR adds a channel title field, a rename screen/endpoint, and channel-event system notices (rename + phase-resolved), all layered on the already-merged channel-screens-redesign. Backend permissions, the no-op-on-unchanged-title guard, event registry wiring, and OpenAPI/generated-client regeneration are all internally consistent and well covered by tests, and I traced the display-name/subtitle precedence logic, the message/event thread-merge sort, and the no-press-game branch (where the rename button turns out to be unreachable since private channels can't exist in an active no-press game) without finding a correctness gap. No comments — nothing here clears the bar for a confident finding.

